PR #39, opened 2011-09-06
benchmarking and performance debugging advanced features in scalable table stores

iampad:
A variety of scalable, semistructured, weak-semantic table stores have been developed and optimized for different priorities such as query speed, ingest speed, availability, and interactivity. As these systems mature, performance benchmarking will advance from measuring the rate of simple workloads to understanding and debugging the performance of advanced features such as ingest speed-up techniques and function shipping filters from client to servers.

We implement a set of extensions to the Yahoo! Cloud Serving Benchmark (YCSB) to improve performance understanding and debugging of these advanced features. The extensions include multi-tester coordination for increased load and eventual consistency measurement, multi-phase workloads to quantify the consequences of work deferment and the benefits of anticipatory configuration optimization such as B-tree pre-splitting or bulk loading, and abstract APIs for explicit incorporation of advanced features in benchmark tests. To enhance performance debugging, we customized an existing cluster monitoring tool to gather the internal statistics of YCSB, table stores, system services like HDFS, and operating systems, and to offer easy post-test correlation and reporting of performance behaviors.

### DIFF
--- a/db/accumulo/src/com/yahoo/ycsb/db/AccumuloClientFilter.java
+++ b/db/accumulo/src/com/yahoo/ycsb/db/AccumuloClientFilter.java
@@ -1,0 +1,313 @@
+package com.yahoo.ycsb.db;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.Vector;
+import java.util.Map.Entry;
+
+import org.apache.hadoop.io.Text;
+
+import org.apache.accumulo.core.Constants;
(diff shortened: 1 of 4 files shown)


busbey:
anyone interested in taking up the YCSB++ changes again? there are several things I like in here (and the Accumulo client was the basis for the one that ended up making it in), but I agree with earlier feedback that this is too big as one giant ball.

looks like some overlap with #255 and maybe #264 ?

PR #47, opened 2011-11-18
Fix an issue with JDBC scans

sudiptodas:
When issuing scans, a JDBC driver can either use a cursor to stream the results as they are consumed or can fetch the entire result set. The postgres JDBC driver does the latter. Therefore, YCSB runs out of heap space in workloads with scans on large tables. I added a new parameter to specify a limit on the number of rows fetched. This is a hint to the driver, not a binding contract, but seems to work with postgres. I felt this is a more portable way compared to using say LIMIT clause, whose syntax varies from one DBMS to another.

The added parameter is scan.fetch.limit whose default is 10 rows and can be overridden by the -p option of the YCSB command line arguments.

Hope this is helpful.

Sudipto


busbey:
this got fixed back in commit 7b564cc3ffdb340f3359c70d2206995e138a151e

PR #90, opened 2012-08-15
Added variance calculation of histogram latency reporting.

enixon:
Part of previous pull request 87:
https://github.com/brianfrankcooper/YCSB/pull/87


busbey:
thank you for taking the time to put this together. our apologies for the lag. I'd like to get this merged quickly, just a couple of comments.

core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementHistogram.java
```diff
@@ -36,11 +36,39 @@
...
+	 * The total number of reported operations.
+	 */
 	int operations;
-	long totallatency;
+	
+	/**
+	 * The sum of each latency measurement over all operations.
+	 * Calculated in ms.
+	 */
+	double totallatency;
```
busbey: why the move to floating point?

enixon: Totallatency was always cast to a double when consumed. I doubt that we'd ever overflow a long as this variable but I felt better with it as a double from the start instead of a late cast. There's no strict reason that it has to be a double and I could change it back easily enough.

busbey: let's leave it our intermediate calculations using fixed point then. I presume the double for display is an attempt to get fractional units. I'd rather not deal with floating point calculation error in our core measurements.

enixon: Done!

busbey:
can you take a look at the travis failures?

core/src/test/java/com/yahoo/ycsb/measurements/exporter/TestMeasurementsExporter.java
```diff
@@ -52,7 +52,7 @@ public void testJSONArrayMeasurementsExporter() throws IOException {
         JsonNode  json = mapper.readTree(out.toString("UTF-8"));
         assertTrue(json.isArray());
         assertEquals(json.get(0).get("measurement").asText(), "Operations");
-        assertEquals(json.get(3).get("measurement").asText(), "MaxLatency(us)");
-        assertEquals(json.get(11).get("measurement").asText(), "5");
+        assertEquals(json.get(4).get("measurement").asText(), "MaxLatency(us)");
+        assertEquals(json.get(11).get("measurement").asText(), "4");
```
enixon: My additional measure is moving the indices of the JSONArrayMeasurementsExporter. The build also works with the following line.
```
assertEquals(json.get(12).get("measurement").asText(), "5");
```

enixon:
Build fixed, merge conflicts fixed. 
Issues caused by move to JSONArrayMeasurementsExporter.

core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementHistogram.java
```diff
@@ -97,8 +128,11 @@ public synchronized void measure(int latency)
   @Override
   public void exportMeasurements(MeasurementsExporter exporter) throws IOException
   {
+    double mean = totallatency/((double)operations);
+    double variance = totalsquaredlatency/((double)operations) - (mean * mean);
     exporter.write(getName(), "Operations", operations);
-    exporter.write(getName(), "AverageLatency(us)", (((double)totallatency)/((double)operations)));
+    exporter.write(getName(), "AverageLatency(ms)", mean);
```
busbey: please keep this in microseconds.

the variance should be in microseconds as well

enixon: Done.

busbey:
Thanks!

PR #91, opened 2012-08-15
Data Correctness Checking

enixon:
Added additional parameter to turn on deterministic creation of values for
specified keys. When running a workload, read data is checked against the
deterministic value and inconsistencies exported to the measurements object.

Discovered flaws in multithreaded YCSB 'latest' workload distribution as
data may be read by one thread before it is written by another.

Part of previous pull request 87:
https://github.com/brianfrankcooper/YCSB/pull/87


busbey:
thanks for the contribution! sorry for the lag.

PR #92, opened 2012-08-20
Workload template

enixon:
Included a template collecting all workload properties into a single file. It shows the defaults and explains usage.


busbey:
thanks for the contribution!

PR #94, opened 2012-09-04
bump gemfire version to 7.0.2

sbawaska:
I have bumped gemfire version to 6.6.3 in order to resolve the problem in finding gemfire.jar from maven repo. I have also fixed an NPE in the client code.


busbey:
the mapkeeper error was fixed in #210 could someone rebase this to current master (and possibly update gemfire versions again if desired)

PR #108, opened 2012-12-18
TransportClient support for connecting to remote elastic search clusters

phaneesh:
The ability to connect to multiple elasticseach hosts in a cluster environment  using transport client is very useful where auto discovery is turned off. This will enable connecting to remote hots without starting a node on the client benchmarking machine.


busbey:
I applied this change locally and build succeeded. If we run into problems when verifying that the next release candidate works with elasticsearch we can start a new issue.

Thanks for doing this! sorry for the lag.

PR #112, opened 2013-01-17
Use the new MongoClient constructor

pquerna:
* Upgrade to MongoClient constructor used in 2.10+ of the MongoDB java client.
* Use the URL for most properties instead fo having multiple configuration options that aren't needed: http://docs.mongodb.org/manual/reference/connection-string/


busbey:

PR #115, opened 2013-01-23
MongoDB readPreference and writeConcern improvements

gelin:
Added ability to set readPreference for read operations.
Added ability to set more detailed writeConcern, which is supported by the latest Mongo driver.
Added ability to ignore insert operation errors, which can be useful to reload incomplete data into MongoDB.


busbey:

PR #119, opened 2013-01-28
Retry failed operations

gelin:
Added ability to retry failed operations. The max number of retries is limited.
The total number of retries is counted. The retries are done within the same
operation, so they doesn’t affect the number of operations, but increases
the latency of the operation. The retries can be done with delays between them.


busbey:
sorry for the lag. I'd like to get this merged in soon. Could we add a test that retries work?

PR #121, opened 2013-01-28
Warm-up

gelin:
Allow to do some number of read operations before gathering statistics, to warm-up the database. The length of warm-up can be limited by the number of operations or by the time period.


busbey:
please don't include unrelated formatting changes.

PR #122, opened 2013-01-29
Automation to run on multiple clients

gelin:
Here are Fabric scripts which allows to run YCSB on multiple client machines and gather the results (output and error logs) from them. And a merge.py script to produce from logs the common result in tabbed-separated form to be inserted into a spreadsheet.

### DIFF
--- a/bin/merge.py
+++ b/bin/merge.py
@@ -38,6 +38,7 @@ def merge():
     ops['INSERT']  = other_ops
     ops['READ']    = other_ops
     ops['UPDATE']  = other_ops
+    ops['CLEANUP'] = other_ops
     ops_keys = ops.keys()
     regexps = map(re.compile, metrics)
     cns = []


busbey:
Sorry for the lag on this issue. This is a great addition!

I'd like to have a generalized approach for folks using different deployment frameworks for launching their YCSB clients. If you're still interested I'd love to have this as one of a handful initial implementations.

We'd need to isolate things in a module, I think. But that might require a better reworking of our assemblies.

k1xme:
Try PerfKitBenchmarker

Nilkanthpatel04:
Hi @gelin , Can you please explain (more details about) how to use merge.py script, I mean script usage like inputs if any and where to generate common stats in tabular form. I tried but getting following

Total
[root@ycsbclient01 ycsb-benchmark]# python YCSB/bin/merge.py


Total
[root@ycsbclient01 ycsb-benchmark]#

gelin:
After running a workload you should get `.out` and `.err` files on each client. One file per each run. The filename contains a time when the workload was started.

`fab ycsb_get` is to download that `.out` and `.err` files into your local machine (into the current dir) to analyse further. You should get a set of files with names ended with `-c1.out`, `-c2.out`, `-c1.err`, `-c2.err`, etc... in the current directory. The command requires `tar` and `bzip2` to be available on remove and local machines.

The `bin/merge.py` is to run in the folder where the `.out` files from all the clients are collected. It scans the `.out` files to gather the final statistics (like throughput and latency) and produce (to stdout) a tab-separated report (to be inserted into a spreadsheet).

gelin:
If YCSB output format was changed in these 9 years, `merge.py` may return empty result. It uses regexes to find lines in output with final statistics. If the format changed it finds nothing.

PR #126, opened 2013-03-27
Enable larger tables

twilli3141:
Need ability to handle larger tables.  Convert to JAVA long for row counters and indices.


README
```diff
@@ -1,5 +1,5 @@
-Yahoo! Cloud System Benchmark (YCSB)
-====================================
+
```
busbey: Leave out the README update

core/src/main/java/com/yahoo/ycsb/generator/ExponentialLongGenerator.java
```diff
@@ -0,0 +1,98 @@
+/**                                                                                                                                                                                
+ * Copyright (c) 2011 Yahoo! Inc. All rights reserved.
```
busbey: Nit: this should be year of authorship

core/src/main/java/com/yahoo/ycsb/measurements/exporter/MeasurementsExporter.java
```diff
@@ -31,10 +31,10 @@
    * 
    * @param metric Metric name, for example "READ LATENCY".
    * @param measurement Measurement name, for example "Average latency".
-   * @param i Measurement to write.
+   * @param l Measurement to write.
    * @throws IOException if writing failed
    */
-  public void write(String metric, String measurement, int i) throws IOException;
```
busbey: this change will be breaking. can we do this by adding an additional method with the long parameter and keep the one that's int?

pom.xml
```diff
@@ -98,13 +98,13 @@
           <configLocation>checkstyle.xml</configLocation>
         </configuration>
         <executions>
-          <execution>
```
busbey: please don't include disabling checkstyle.

twilli3141:
Updated per feedback

core/src/main/java/com/yahoo/ycsb/generator/LongGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+	{
+        lastlong=last;
+	}
+
+    /**
+	 * Generate the next string in the distribution.
+	 */
+    public String nextString()
+    {
+        return ""+nextLong();
```
busbey: please use Long.toString(long)

core/src/main/java/com/yahoo/ycsb/generator/LongGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+	 * Return the previous string generated by the distribution; e.g., returned from the last nextString() call. 
+	 * Calling lastString() should not advance the distribution or have any side effects. If nextString() has not yet 
+	 * been called, lastString() should return something reasonable.
+	 */
+	@Override
+	public String lastString()
+	{
+		return ""+nextLong();
```
busbey: please use Long.toString(long)

core/src/main/java/com/yahoo/ycsb/generator/ExponentialLongGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+	public long nextLong()
+	{
+		return (long) (-Math.log(Utils.random().nextDouble()) / _gamma);
+	}
+
+	@Override
+	public double mean() {
+		return 1.0/_gamma;
+	}
+    public static void main(String args[]) {
```
busbey: do not add main() methods for testing, instead add as a junit test if you want to include.

core/src/main/java/com/yahoo/ycsb/generator/ScrambledZipfianGenerator.java
```diff
@@ -105,11 +104,12 @@ public int nextInt() {
 	/**
 	 * Return the next long in the sequence.
 	 */
+    @Override
 	public long nextLong()
 	{
 		long ret=gen.nextLong();
 		ret=_min+Utils.FNVhash64(ret)%_itemcount;
-		setLastInt((int)ret);
+		setLastLong((int)ret);
```
busbey: this shouldn't truncate ret to an int.

core/src/main/java/com/yahoo/ycsb/generator/SkewedLatestGenerator.java
```diff
@@ -20,32 +20,32 @@
...
 	 */
-	public int nextInt()
+	public long nextLong()
 	{
-		int max=_basis.lastInt();
-		int nextint=max-_zipfian.nextInt(max);
-		setLastInt(nextint);
-		return nextint;
+        long max=_basis.lastLong();
+        long nextlong=max-_zipfian.nextInt((int)max);
```
busbey: shouldn't this use zipfian.nextLong?

core/src/main/java/com/yahoo/ycsb/generator/ConstantIntegerGenerator.java
```diff
@@ -36,7 +36,12 @@ public int nextInt() {
 		return i;
 	}
 
-	@Override
+    public long nextLong()
```
busbey: IntegerGenerator doesn't have a nextLong method, so it's a bit odd for these other subclasses to include a nextLong.

core/src/main/java/com/yahoo/ycsb/generator/LongGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+
+package com.yahoo.ycsb.generator;
+
+/**
+ * A generator that is capable of generating ints as well as strings
+ * 
+ * @author cooperb
+ *
+ */
+public abstract class LongGenerator extends Generator
```
busbey: To maintain compatibility, you could have LongGenerator extend IntegerGenerator so that all the current implementers will stay have IntegerGenerator as an ancestor (and having getInt will make sense)

busbey:
sorry I didn't catch some of these issues on the first review pass.

twilli3141:
Updated per feedback

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -498,28 +500,35 @@ else if (op.compareTo("SCAN")==0)
...
-        int keynum;
-        if(keychooser instanceof ExponentialGenerator) {
+    long nextKeynum() {
+        long keynum=-1;
+
+        if(keychooser instanceof ExponentialLongGenerator) {
             do
                 {
-                    keynum=transactioninsertkeysequence.lastInt() - keychooser.nextInt();
+                    //keynum=transactioninsertkeysequence.lastLong() - keychooser.nextInt();
```
busbey: I started getting this merged in, but testing failed. looks like having this commented out results in an infinite loop.

PR #130, opened 2013-05-23
Addition of a statistics Measurement class

cspann:
Hello,

I have created another Measurement called `OneMeasurementStatistics` and would like to share it with others in need of some statistical data output. I have modified the `Measurements` class in the way that I added an enum that specifies the existing Measurement types as they can be named in the workload config file. Pleas let me know if the inclusion of my code would require further work. The new class requires the `apache.commons.math` library, which does the statistical stuff except the 95% confidence interval calculation.

_Commit Msg_
It creates statistically oriented output including the mean, standard
deviation and the 95% confidence interval if more than 30 values are
mesured. To select the output type, a third "measurementtype" option
named "statistics" was added to the possibilities.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Client.java
+++ b/core/src/main/java/com/yahoo/ycsb/Client.java
@@ -398,7 +398,7 @@ public class Client
 				out = System.out;
 			} else
 			{
-				out = new FileOutputStream(exportFile);
+				out = new FileOutputStream(exportFile, true);
 			}
 
 			// if no exporter is provided the default text one will be used


busbey:
Thanks for working on this, apologies for the project lag. I'll start reviewing now.

If you aren't interested in working on it still, that's totally understandable. Just leave the PR open and eventually I or someone else will try to take it over.

core/pom.xml
```diff
@@ -32,6 +32,11 @@
       <version>6.1.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-math</artifactId>
+      <version>2.2</version>
```
busbey: Please use commons-math 3.

core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -1,802 +1,802 @@
-/**
```
busbey: Please don't include formatting changes as it makes reviewing the substantive changes much harder.

core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementStatistics.java
```diff
@@ -0,0 +1,102 @@
+/**                                                                                                                                                                                
+ * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
```
busbey: Nit: the year should be the year(s) the file was authored during

busbey:
Is mean + stddev going to be a useful addition compared to e.g. median and 99th percentile?

PR #131, opened 2013-06-21
Added support of testing MongoDB replica set with automatic failover.

sergeytsivin:
1) MongoDB driver upgraded to the latest version 2.11.2
2) Use mongodb.host and mongodb.port properties instead of mongodb.url
3) Added mongodb.readPreference property which can be secondary or primary (default)


busbey:

PR #141, opened 2013-09-20
Use the clear buffer on fail option. Not using it is now deprecated.

nkeywal:
Reasons are detailed in https://issues.apache.org/jira/browse/HBASE-9521
This makes errors simpler to understand: without this setting, the failed insertions are retried when we close the connection to HBase.


busbey:
Thanks for the contribution! sorry for the lag.

nkeywal:
I'm pleased to see that you now have the commit bit, Sean :-)

PR #145, opened 2013-10-17
Added Cassandra CQL binding.

cmatser:
Cassandra CQL is the newer query language for accessing Cassandra DB.  It's been around for a while and seems to be the language of choice going forward with Cassandra.  It's very similar to SQL.  As such, the learning curve to using Cassandra is easier through CQL.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,8 @@
   <properties>
     <maven.assembly.version>2.2.1</maven.assembly.version>
     <hbase.version>0.92.1</hbase.version>
-    <cassandra.version>0.7.0</cassandra.version>
+    <cassandra.version>1.2.9</cassandra.version>
+    <cassandra.cql.version>1.0.3</cassandra.cql.version>
     <infinispan.version>7.1.0.CR1</infinispan.version>
     <openjpa.jdbc.version>2.1.1</openjpa.jdbc.version>
     <mapkeeper.version>1.0</mapkeeper.version>
@@ -67,7 +68,7 @@
     <module>dynamodb</module>
     <module>elasticsearch</module>
     <!--<module>gemfire</module>-->
-    <module>infinispan</module>
+    <!--<module>infinispan</module>-->
     <module>jdbc</module>
     <module>mapkeeper</module>
     <module>mongodb</module>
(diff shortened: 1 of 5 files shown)


busbey:
thanks for the contribution and sorry for the lag. I intend to merge this in by the end of the week. I just have a couple of feedback points.

If you're not still working on this, I can make the modifications myself.

pom.xml
```diff
@@ -67,7 +68,7 @@
     <module>dynamodb</module>
     <module>elasticsearch</module>
     <!--<module>gemfire</module>-->
-    <module>infinispan</module>
```
busbey: please don't include changes to other modules. if infinispan currently isn't building, we can disable it as a different PR.

pom.xml
```diff
@@ -45,7 +45,8 @@
   <properties>
     <maven.assembly.version>2.2.1</maven.assembly.version>
     <hbase.version>0.92.1</hbase.version>
-    <cassandra.version>0.7.0</cassandra.version>
+    <cassandra.version>1.2.9</cassandra.version>
```
busbey: does this change mean we can't work with pre-1.2 servers?

pom.xml
cmatser: Good point.  Please update this PR for me.

pom.xml
cmatser: I added the new binding as a separate class CassandraCQLClient and left the previous bindings untouched.  So, you can use those with pre-1.2 servers.  But this property is used in the maven build to pull in the right dependencies for the new CassandraCQLClient.  I don't know if the dependencies for the pre-1.2 version are still being pulled in with this change.

I suppose I could have created a "cassandracql.version" property?  Then both old and new jars would be pulled, but I'm skeptical about that working.  Some dependency jars could get clobbered with older or newer versions.  I might tend towards creating it as a separate binding completely, with its own directory if we need old and new to work side-by-side.

I'm not sure this answers your question.  The folks at Datastax told me that Cassandra CQL was the direction of the data store moving forward.  They did not recommend using the older binding.  I did not test with it.

Thanks for your work!  Unfortunately, I don't have access to the environment I had when I did this work, but I'll help as much as I can.

busbey: that helps a bunch, thanks for getting back to me.

I suspect separate bindings are the way to go, but I agree that still working with older versions can be something we worry about later.

PR #169, opened 2014-04-14
YCSB+T: Added transactions to DB and DBWrapper and validate() to Workloa...

akon-dey:
...d.

YCSB+T: Added an example ClosedEconomyWorkload with a simple implementation of validate and a Simple Anomaly Score.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/workloads/ClosedEconomyWorkload.java
+++ b/core/src/main/java/com/yahoo/ycsb/workloads/ClosedEconomyWorkload.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
+ * Copyright (c) 2014 YCSB contributors.. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You


busbey:
bump on this issue. would like to see this resolved for the Sep release.

busbey:
since this changes the core binding definition it'll probably take some work to get right. /cc @allanbank since the patch as is touches the Mongo code, though I've no reason to believe it will need substantially different handling than other datastores.

open quesiton: if we have commit / abort, what do we do with datastores that have no notion of transactions?

akon-dey:
The default start(), commit() and abort() are defined as empty methods in DB.java. This is similar to the cleanup() and init() default implementations.

busbey:
ah. and the default consistency check just says yes blindly. I see. could merge that with the new verify stuff in core workload.

thanks for the pointer!

akon-dey:
I have merged the changes but the verification code is not fully integrated yet. I'd be happy to help with it.

core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -309,57 +309,76 @@ public void run()
...
+		   //GH issue 4 - throws exception if _target>1 because random.nextInt argument must be >0
+		   //and the sleep() doesn't make sense for granularities < 1 ms anyway
+		   if ( (_targetOpsPerMs>0) && (_targetOpsPerMs<=1.0) ) 
+		   {
+		      sleep(Utils.random().nextInt((int)(1.0/_targetOpsTickNs)));
+		   }
+		}
+		catch (InterruptedException e)
+		{
+		  // do nothing.
```
busbey: this is probably wrong. we should at a minimum set the interrupted status of the thread.

busbey:
please don't include additional format fixes, it makes it harder to review the substance of the change.

core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -309,57 +309,76 @@ public void run()
...
-      System.exit(0);
-    }
+		//spread the thread operations out so they don't all hit the DB at the same time
+		try
+		{
+		   //GH issue 4 - throws exception if _target>1 because random.nextInt argument must be >0
+		   //and the sleep() doesn't make sense for granularities < 1 ms anyway
+		   if ( (_targetOpsPerMs>0) && (_targetOpsPerMs<=1.0) ) 
+		   {
+		      sleep(Utils.random().nextInt((int)(1.0/_targetOpsTickNs)));
```
busbey: this looks unrelated to the change mentioned in the PR description, could you leave it out?

akon-dey: Sorry about the format changes. I could merge it again if that helps.

core/src/main/java/com/yahoo/ycsb/Workload.java
```diff
@@ -80,8 +80,9 @@ public void cleanup() throws WorkloadException
        * other, and it will be difficult to reach the target throughput. Ideally, this function would have no side
        * effects other than DB operations and mutations on threadstate. Mutations to threadstate do not need to be
        * synchronized, since each thread has its own threadstate instance.
+       * @throws WorkloadException
```
busbey: what purpose does the exception serve?

core/src/main/java/com/yahoo/ycsb/workloads/ClosedEconomyWorkload.java
```diff
@@ -0,0 +1,736 @@
+/**
+ * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
```
busbey: Please use Copyright (c) YCSB contributors. with the year this file was authored.

core/src/main/java/com/yahoo/ycsb/workloads/ClosedEconomyWorkload.java
```diff
@@ -0,0 +1,736 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Hashtable;
+import java.util.Vector;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * The core benchmark scenario. Represents a set of clients doing simple CRUD
```
busbey: this description doesn't line up with what the workload is doing, or at least what distinguishes it from CoreWorkload.

also there's a fair bit of duplication. better to make use of hte CoreWorkload directly.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -290,6 +291,14 @@
    * Default value of the percentage operations accessing the hot set.
    */
   public static final String HOTSPOT_OPN_FRACTION_DEFAULT = "0.8";
+  
+  private Hashtable<String, String> _operations = new Hashtable<String, String>() {{
```
busbey: AFAICT, this map does not need to be synchronized. Replace it with final Map and instantiate it in an initializer block where you can wrap it in Collections.unmodifiableMap

uakfdotb:
It seems like the update operations reset records to the initial value (totalcash / recordcount), if there has been a readmodifywrite performed on the record then it seems like this breaks the closed economy invariant.

akon-dey:
Please review the recent changes and let me know if you want me to address anything.

core/src/main/java/com/yahoo/ycsb/Workload.java
akon-dey: It says whether the workload operation (mostly for validate()) was successful or not. This is necessary since we now have the ability to determine whether a database operation was successful or not within the context of the transaction.

benjaminbrent:
@akon-dey I wrote a very similar `start()`, `commit()`, `abort()` implementation for `JdbcDBClient` last year, except I added the ability to set the transaction isolation level (see https://github.com/benjaminbrent/YCSB/blob/e2aa4f4be142afc2acfd98e5c3e22bd659da8d01/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java#L255).

It would be nice to have this in your implementation too. I could submit a pull request if you think this is worthwhile?

akon-dey:
This would be useful. We should be able to set the isolation level using the workload configuration properties file. Thanks twill make it truly is able.

akon-dey:
Please submit a pull request with the ability to set the workload properties.

morcelicaio:
Hello @akon-dey .

Have you noticed the behavior of ycsb+t when running the load phase (in the ClosedEconomyWorkload layer) using the parameter w = 0 in the mongo db cluster?

It simply gives a NumberFormatException error on the last record (in the ClosedEconomyWorkload layer).

I'm running on Mongodb Atlas and I noticed this error doing tests.

Do you have any idea about this behavior?

akon-dey:
@morcelicaio I have not experienced it. If you can send me a stack or better, a pull request, I would be happy to incorporate it.

morcelicaio:
@akon-dey , I did some more tests on another computer and didn't get the same error. Maybe it could have been some point on the other machine. If I get the same error again, I'll let you know. Thanks!

morcelicaio:
Hi @akon-dey, I would like to ask another question.

Wasn't YCSB+T merged with the original YCSB?

As I understand it, in the latest available version of YCSB there is no transactional layer you added. I would like to know the status of this.

Thanks!

morcelicaio:
Hi @akon-dey and @busbey , I would like to ask another question.

Wasn't YCSB+T merged with the original YCSB?

As I understand it, in the latest available version of YCSB there is no transactional layer you added. I would like to know the status of this.

Thanks!

akon-dey:
@morcelicaio I don't believe it was merged. There were some changes to the PR that had questions and hence not merged. One was code format related and the other was regarding completeness of the ClosedEconomyWorkload. The first was trivial and the second was not intended to be addressed by the code. It would be best to recreate the PR to address the first issue and merge if it looks good.

busbey:
it has not yet been merged. There's also merge conflicts, so it will need to go through a rebase before review.

morcelicaio:
Thanks for the clarification, guys.

However, @akon-dey , do you think it's safe for me to use YCSB+T to check if the MongoDB database guarantees ACID properties? Does it serve this purpose?

As I understand YCSB without the extension you added doesn't work with database transaction checks.

I would like to do this kind of test in my work.
In this case I would do the tests from the lowest consistency configuration level of MongoDB to the strongest consistency level and then I would check the anomalies found during the transactions.

morcelicaio:
Hi, @akon-dey, about using YCSB+T from my previous question, what do you think? Is it reliable to do this test?

Another question regarding the extension.
To assess the transactional consistency of the database, can I use any workload? Both CoreWorkload and ClosedEconomyWorkload?

What is the difference between the two? I have verified that the ClosedEconomyWorkload workload you use to test with currency movement. I was left with this doubt to choose which of the workloads I can use to do this test that I'm intending to.

morcelicaio:
Hello guys, @busbey, @akon-dey .

Regarding YCSB+T, he talks about evaluating the transactional part of NoSQL databases.

I'm still testing with MongoDB.
I noticed that in the YCSB+T implementation we don't have the code implemented that creates the transaction as it is in the MongoDB documentation Transaction API (https://www.mongodb.com/docs/manual/core/transactions/).

Do I need to implement this code to use YCSB+T?
Or do the start() abort() and commit() methods somehow perform these "transactions" implicitly?

Thanks

busbey:
I would assume that if you don't see the mongo transaction API being called then probably the existing code is not making use of it.

morcelicaio:
Just that, @busbey . So how does YCSB+T assess transactional overhead and have a layer to assess transactional consistency when the transaction code implementation is not in the YCSB+T code?

It's a question that maybe other users have as well.

busbey:
as mentioned previously, this _proposed_ change hasn't been merged yet. It also started in 2014, so it's possible it predates the mongo transactional api.

PR #183, opened 2014-07-10
Upgraded OrientDB

lvca:
I've upgraded from the ancient version 1.0.1 (3 years ago) to 1.7.5-SNAPSHOT.


busbey:
Happy to update the orientdb version, but it needs to be to a non - snapshot version.has 1.7.5 come out? Can you isolate the PR to just updating the version?

lvca:
My comment was 1y old, we could update to latest. @laa do you have any more update release to merge?

busbey:
Apologies for the lag on the project. We're just spinning things up again. In the future, responses should be much faster.

It looks like 1.7.5 definitely came out, but your most recent version is 2.0.9? Is there any reason not to just start support on the 2.0 line? Are folks likely to want to compare 1.7 with 2.0?

PR #188, opened 2014-09-01
Fix hbase client's write bug in debug mode.

y0un5:
Hbase client has a bug to write all values as empty byte arrays in debug mode. It is caused by an illegal use of ByteIterator. Values returned from toString() or toArray() should be cached if they are needed later.

### DIFF
--- a/hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
+++ b/hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
@@ -316,11 +316,12 @@ public class HBaseClient extends com.yahoo.ycsb.DB
         Put p = new Put(Bytes.toBytes(key));
         for (Map.Entry<String, ByteIterator> entry : values.entrySet())
         {
+            byte[] value = entry.getValue().toArray();
             if (_debug) {
                 System.out.println("Adding field/value " + entry.getKey() + "/"+
-                  entry.getValue() + " to put request");
+                  Bytes.toStringBinary(value) + " to put request");
             }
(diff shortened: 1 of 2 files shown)


busbey:
restarted CI

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
```diff
@@ -316,11 +316,12 @@ public int update(String table, String key, HashMap<String,ByteIterator> values)
         Put p = new Put(Bytes.toBytes(key));
         for (Map.Entry<String, ByteIterator> entry : values.entrySet())
         {
+            byte[] value = entry.getValue().toArray();
             if (_debug) {
                 System.out.println("Adding field/value " + entry.getKey() + "/"+
-                  entry.getValue() + " to put request");
+                  Bytes.toString(value) + " to put request");
```
busbey: please use Bytes.toHexString, since the workload might not use a value that can be turned into a string.

y0un5: What do you think about using Bytes.toStringBinary?
Its Javadoc says "Write a printable representation of a byte array. Non-printable characters are hex escaped in the format \\x%02X, eg: \x00 \x05 etc"
http://hbase.apache.org/apidocs/org/apache/hadoop/hbase/util/Bytes.html#toStringBinary(byte[])

busbey: that would do nicely.

y0un5:
I've pushed new commit that uses Bytes.toStringBinary and fixes HBaseClient10, too.

busbey:
Thanks for the fix!

PR #192, opened 2014-10-26
Adding supports for variable-sized keys, additionally a Generalized Extreme Value Distribution generator

fomy:
As reported in a paper (Workload Analysis of a Large-Scale Key-Value Store) at SIGMETRIC'2012, the key size  is Generalized Extreme Value distribution in Facebook. 
I hence implemented a new IntegerGenerator for Generalized Extreme Value distribution (see com.yahoo.ycsb.generator.ExtremeValueGenerator).
The comparisons of the theoretical distribution and the generated distribution are shown in the attached files.
extreme_cdf
extreme_density

To support variable-sized keys, I added a persistent array of key lengths.
The array records the lengths of all keys, one byte for each key.
The length of the first key (namely "1") is in the first slot, second in the second slot, and so on.
The array would be initialized in Load phase, and persisted into a file named "keylengths.array".
In Run phase, the file is read back to re-construct the array.
It will grow up if there are insert transactions.
In each operation, buildVariableKeyName() is called to construct the key according to the key length and the key number.

Additionally, since the keylengths.array file records the size of the current key space, we can execute several Run phases after a single Load phase without the need of manually updating the 'recordcount' in the workload file.

Best regards
Min Fu


busbey:
Hi! Could you rebase to current master so we can try to get a clean CI build?

core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -487,11 +487,15 @@ else if (args[argindex].compareTo("-load")==0)
 			{
 				dotransactions=false;
 				argindex++;
+				/* added by Min Fu. Let _workload be aware of this. */
```
busbey: No need to note who added this.

core/src/main/java/com/yahoo/ycsb/generator/ExtremeValueGenerator.java
```diff
@@ -0,0 +1,84 @@
+/**                                                                                                                                                                                
+ * Added by Min Fu.
```
busbey: File needs a license header.

Git log will show who adeed the file. If yoy want an additional notice, use an author tag.

core/src/main/java/com/yahoo/ycsb/generator/ExtremeValueGenerator.java
```diff
@@ -0,0 +1,84 @@
...
+	 * @return The next item in the sequence.
+	 */
+	public long nextLong() {
+		double u = Utils.random().nextDouble();
+		double ret = (mu + sigma * (Math.pow(-Math.log(u), -xi) - 1) / xi);
+		return (long) (ret > maxlength ? maxlength : ret);
+	}
+
+	/**
+	 * @todo Implement ZipfianGenerator.mean()
```
busbey: looks like a stale class reference?

please open an issue to reference for the todo

core/src/main/java/com/yahoo/ycsb/generator/ExtremeValueGenerator.java
```diff
@@ -0,0 +1,84 @@
...
+	 */
+	@Override
+	public double mean() {
+		throw new UnsupportedOperationException(
+				"@todo implement ZipfianGenerator.mean()");
+	}
+
+	public static void main(String[] args) {
```
busbey: Please don't include test code in a main() section. If a utility for generating example output is needed, please add an additional class for it. (or maybe open an issue so we can generalize adding new cli tools)

core/src/main/java/com/yahoo/ycsb/generator/FacebookValueGenerator.java
```diff
@@ -0,0 +1,91 @@
+package com.yahoo.ycsb.generator;
+
```
busbey: Need a license header.

core/src/main/java/com/yahoo/ycsb/generator/FacebookValueGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	@Override
+	public double mean() {
+		// TODO Auto-generated method stub
+		return 0;
+	}
+	
+}
+
```
busbey: need javadocs. I don't see a description of this in the top level. is it also from the paper?

Please find a suitable name that does not use Facebook's trademarks.

workloads/workloadg
```diff
@@ -0,0 +1,39 @@
...
+# Unless required by applicable law or agreed to in writing, software                                                                                                             
+# distributed under the License is distributed on an "AS IS" BASIS,                                                                                                               
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or                                                                                                                 
+# implied. See the License for the specific language governing                                                                                                                    
+# permissions and limitations under the License. See accompanying                                                                                                                 
+# LICENSE file.                                                                                                                                                                   
+
+
+# Yahoo! Cloud System Benchmark
+# Workload A: Update heavy workload
```
busbey: update for new workload name

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -456,7 +651,8 @@ public String buildKeyName(long keynum) {
 	public boolean doInsert(DB db, Object threadstate)
 	{
 		int keynum=keysequence.nextInt();
-		String dbkey = buildKeyName(keynum);
+		//String dbkey = buildKeyName(keynum);
```
busbey: if you're going to change these lines, just change them rather than commenting them out. we can always use git to retrieve them.

busbey:
Could we break the new workload into a new class rather than trying to fit it into CoreWorkload? something like VariableKeyLengthWorkload?

workloads/workloada
```diff
@@ -34,4 +34,3 @@ scanproportion=0
 insertproportion=0
 
 requestdistribution=zipfian
-
```
busbey: please don't include unrelated formatting changes.

Issue #181, opened 2014-07-07
Custom repo for MapKeeper no long available


busbey:
As a part of #64, @m1ch1 set up a custom maven repo to hold mapkeeper releases at:

https://raw.github.com/m1ch1/m1ch1-mvn-repo/master/releases

AFAICT, that repo is no longer present under @m1ch1's github, so building now fails, unless you comment out the mapkeeper module.

(I suspect this is the cause of #150 and #152 )

Before I make a patch to remove the module from the normal top level build, would someone else like to set up a repo for the mapkeeper artifacts?

Issue #152, opened 2013-12-05
"mvn clean package" failed when binding MapKeeper.


amber-zhang:
Hi,

When I ran "mvn clean package" to create project, I got this error when binding MapKeeper. Although I manually downloaded MapKeeper.jar and put it ./YCSB-master/mapkeeper/target, the error still occur. How can I fix the problem? thank you!

[INFO] YCSB Root ......................................... SUCCESS [0.689s]
[INFO] Core YCSB ......................................... SUCCESS [4.593s]
[INFO] Cassandra DB Binding .............................. SUCCESS [5.774s]
[INFO] HBase DB Binding .................................. SUCCESS [15.647s]
[INFO] Hypertable DB Binding ............................. SUCCESS [4.104s]
[INFO] DynamoDB DB Binding ............................... SUCCESS [5.083s]
[INFO] ElasticSearch Binding ............................. SUCCESS [7.186s]
[INFO] Infinispan DB Binding ............................. SUCCESS [29.247s]
[INFO] JDBC DB Binding ................................... SUCCESS [4.041s]
[INFO] Mapkeeper DB Binding .............................. FAILURE [1.477s]
[INFO] Mongo DB Binding .................................. SKIPPED
[INFO] OrientDB Binding .................................. SKIPPED
[INFO] Redis DB Binding .................................. SKIPPED
[INFO] Voldemort DB Binding .............................. SKIPPED
[INFO] YCSB Release Distribution Builder ................. SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 1:18.023s
[INFO] Finished at: Thu Dec 05 05:17:52 EST 2013
[INFO] Final Memory: 125M/304M
[INFO] ------------------------------------------------------------------------
[...]

PR #207, opened 2015-01-21
Synchronize Measurements.getSummary() method to prevent ConcurrentModificationExceptions

benjaminbrent:
Previously the method `Measurements.getSummary()` wasn't set to synchronized, which was occasionally resulting in `java.util.ConcurrentModificationException` being raised when running YCSB with multiple threads. This commit fixes the issue by simply adding the `synchronized` keyword to `Measurements.getSummary()`.

Let me know if you can think of any potential issues with making this change. I'm unsure if `getSummary()` was deliberately left unsynchronized for a reason?

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
+++ b/core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
@@ -154,7 +154,7 @@ public class Measurements
       /**
        * Return a one line summary of the measurements.
        */
-	public String getSummary()
+	public synchronized String getSummary()
 	{
 		String ret="";
 		for (OneMeasurement m : data.values())


benjaminbrent:
Similarly to #209, the build failure here is caused by #152, #181.

busbey:
restarted CI build.

Won't this have a big perf impact?

benjaminbrent:
@busbey I haven't witnessed any major performance impact, but I could be missing something. As far as I can see, `Measurements.getSummary()` is only called [here](https://github.com/brianfrankcooper/YCSB/blob/5659fc582c8280e1431ebcfa0891979f806c70ed/core/src/main/java/com/yahoo/ycsb/Client.java#L106) ever 10 seconds.

busbey:
that's a good point. There's also enough going wrong with synchronization across the rest of the class that this probably won't matter. hopefully all of this will get fixed by #214.

Doesn't exportMeasurements have the same problem?

benjaminbrent:
@busbey `exportMeasurements` may also have a synchronisation issue, but I haven't been using the exporter so am unsure.

busbey:
Filed #249 for the other method.

busbey:
Thanks for the fix!

PR #209, opened 2015-01-28
Remove hard-coded property key strings in favour of string constants

benjaminbrent:
Previously the YCSB core code contained some hard-coded property keys, with others defined as constants. This pull request removes the remaining hard-coded strings in favour of string constants defined in `core/src/main/java/com/yahoo/ycsb/Client.java` and `core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java`.


benjaminbrent:
Note: The build failure is due to #152, #181.

busbey:
restarted CI build.

busbey:
Thanks for the improvement!

PR #213, opened 2015-02-09
Add thread synchronisation to OneMeasurementTimeSeries

hwegert:
affected methods:
 + public void reportReturnCode(int code)
 + public void measure(int latency)

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementTimeSeries.java
+++ b/core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementTimeSeries.java
@@ -103,7 +103,7 @@ public class OneMeasurementTimeSeries extends OneMeasurement
 	}
 	
 	@Override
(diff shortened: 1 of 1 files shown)


allanbank:
@hwegert -

There was another pull request (#357) that fixed the reportReturnCode() using ConcurrentHashMaps. For the measure(int) method I think it will be better to use AtomicIntegerFieldUpdater/AtomicLongFieldUpdater classes to update the fields without holding a lock. 

AtomicIntegerFieldUpdater/AtomicLongFieldUpdater are a little esoteric but reduce memory overhead for integer and long fields that need to be accessed in a concurrent way. I am happy to help walk you through porting what you have over to using the atomic field updaters if you are not familiar with them.

Rob.

busbey:
This effectively for done in #605 by synchronizing on an internal method shared by both calls here.

If we're still interested in the approach that avoids synchronized, we should do it in a new PR.

PR #215, opened 2015-03-08
Use more recent monogdb bindings

kaimast:
Versions >= 2.12 use the new bulk API. So there might be a big performance difference.


busbey:
combining with #221

PR #256, opened 2015-05-21
Add autoflush property to HBaseClient

takebayashi:
Add `autoflush` property to HBaseClient. It enables to turn on or off auto flush flag of HTable.


busbey:
We should try to make the configuration of this behavior common between the 1.0 and pre-1.0 clients.

The relationship between HTable.autoflush in pre-1.0 and BufferedMutator in 1.0 is explained [in the client section of the hbase reference guide](http://hbase.apache.org/book.html#client.writebuffer).

It looks like right now the 1.0 binding uses a property called "clientBuffering". Can we do the same in the older api? They should also probably both default to not doing client buffering.

takebayashi:
OK, I renamed the property to `clientbuffering`, the same name as that of the 1.0 client, and changed their default value to false.

busbey:
Thanks for the improvement!

PR #259, opened 2015-05-23
Remove unnecessary configurations in hbase-site.xml

y0un5:
hbase.rootdir and hbase.master are unnecessary for hbase client.

### DIFF
--- a/hbase/src/main/conf/hbase-site.xml
+++ b/hbase/src/main/conf/hbase-site.xml
@@ -23,19 +23,7 @@
 -->
 <configuration>
   <property>
-    <name>hbase.rootdir</name>
-    <value>hdfs://<HBASE_MASTER_SERVER>:20001/hbase</value>
-    <description>The directory shared by region servers.
-    </description>
-  </property>
-  <property>
-    <name>hbase.master</name>
-    <value><HBASE_MASTER_SERVER>:60000</value>
-    <description>The host and port that the HBase master runs at.
-    </description>
-  </property>
-  <property>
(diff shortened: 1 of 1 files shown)


busbey:
do we need this file at all? Shouldn't the client be getting the hbase-site.xml file from the actual cluster we're running against?

y0un5:
As you said, hbase-site.xml file can be removed entirely, and we can guide that users should provide their own hbase-site.xmls.

But the existence of this file is self-explanatory for a user's question such that "Where should I put my hbase-site.xml?". So, I think that it is better to include hbase-site.xml with minimum configurations.

What do you think about this?

busbey:
the issue with using this as a sample location is that we shouldn't require users to put their hbase-site file in our code base. Any solution for #261 should trivially enable passing the hbase client configs in where they live.

y0un5:
I understand. I close this PR.

busbey:
Just wanted to note my thanks for bringing this up in the first place. The
presence of a mis-configured client hbase-site file remains confusing for
new folks.

Would you consider reopening the PR with a commit that just deletes the
file entirely?

y0un5:
I think that to delete the file entirely, we should provide a guide document about how to configure hbase-binding.

If you can wait a few days, I'll make the doc for hbase-binding(such as https://github.com/brianfrankcooper/YCSB/tree/master/couchbase), and open a new pull request including the doc and the deletion of the file.

How about it?

busbey:
that sounds great!

PR #260, opened 2015-05-23
[version] update to 0.2.0-SNAPSHOT.

### LINKED ISSUE #236: versioning YCSB
Our last release was ~3 years ago. Before working towards the next one, I'd like to document what we convey to others with our version numbers.

Given version numbers that look like x.y.z I'd like to define compatibility with

* _Major_ version x: incremented when pre-existing custom Workloads or Database bindings would break.
* _Minor_ version y: incremented when new Workloads or Database bindings are added.
* _Patch_ version z: just bug fixes

One thing I'm not sure about is improvements to the Client or changes in statistics gathered. These things can change how comparable results are for YCSB runs, but I'm not sure tests that run on different hardware set ups are comparable anyways.

Once we have some consensus I'll put together a new wiki page.

busbey:
* Need a SNAPSHOT version because non-SNAPSHOT is reserved for actual releases.
* Increment minor version per pre-1.0 guidelines from the Version Numbers guide

Fixes #236.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
   <groupId>com.yahoo.ycsb</groupId>
   <artifactId>root</artifactId>
-  <version>0.1.4</version>
+  <version>0.2.0-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <name>YCSB Root</name>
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -13,6 +13,12 @@
   <description>
     This is the top level project that builds, packages the core and all the DB bindings for YCSB infrastructure.
   </description>
+
+  <scm>
+    <connection>scm:git:git://github.com/someuser/somelibrary.git</connection>
```
cmccoy: brianfrankcooper/YCSB?

busbey: Excellent catch. Should have been more careful when replacing my original git@github ssh connection.

busbey:
updated the scm tag to properly point at the repo instead of example text. :smile_cat:

cmccoy:
:+1:

busbey:
rebased to resolve conflict with previous merge. content is unchanged.

PR #268, opened 2015-05-29
Rebase and updated PR-94: "bump gemfire version to 7.0.2"

busbey:
* rebased the changes from #94 to current master, squashed them into one commit.
* looking at the references S3 based repo, the most recent version of the artifact the original binding is based on is 8.1.0.

I'd really like to update to a more recent gemfire binding, and one that isn't coming from an s3 bucket. I'll file a follow-on once this goes in. if we can find someone to do the work maybe spring-data-gemfire is the way to go?

### DIFF
--- a/gemfire/pom.xml
+++ b/gemfire/pom.xml
@@ -15,7 +15,7 @@
     <dependency>
       <groupId>com.gemstone.gemfire</groupId>
       <artifactId>gemfire</artifactId>
-      <version>6.6</version>
+      <version>7.0.2</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


cmccoy:
Looks good to me.

busbey:
while rebasing so I could merge I found a logic error in how the NPE for un-filtered reads was fixed. It was just adding the record id as the key over and over, when it should have been each field name.

busbey:
ah. that part was fixed correctly back in #123. so I'll just merge the rest of the approved changes then.

Issue #241, opened 2015-05-17
document logging guidance for new components


busbey:
document what components like db bindings should put at various log levels.

PR #270, opened 2015-05-29
rebased PR145: "Added Cassandra CQL binding"

busbey:
PR from #145 rebased for current master with change to unrelated module removed.


cmccoy:
This will need to be updated once there's consensus on #241. Looks good to me.

PR #274, opened 2015-06-02
Make write buffer size of HBase client configurable

takebayashi:
Make write buffer size of HBase client configurable by adding a new property `writebuffersize`.


busbey:
I was going to ask for an update to the HBase docs to include the new property, but I see that we have no README for the HBase bindings. :sob:

Issue #266, opened 2015-05-29
Release version 0.2.0


busbey:
I'd like to get 0.2.0 out the door by the end of June. Normally, I'd want to set some goals for things to include in a release, but this one is kind of simpler

* everything that's gone in the last ~3 years
* verified working guide for running on windows
* verification that at least 3 of the bindings work

I expect this first go will be bumpy. To meet the end-of-June goal, I plan to cut a release branch off of master starting June 15th (barring special circumstances).

busbey:
If folks could volunteer for testing particular data store bindings that would be a super big help. To make the release notes the testing will have to be on one of the release candidates, but seeing if things work at all right now (and filling in missing READMEs) would be grand.

busbey:
I think #250 and #261 will have to be blockers.

busbey:
we need a manual check that #91 behaves correctly.

busbey:
My kid was ill over the weekend, so I didn't get to finish #250. Presuming no one has a strong objection, I'm going to punt the release branch to Wednesday Jun 17th.

busbey:
All of the blocker and critical issues for 0.2.0 are solved now, so shortly I'll create the release candidate branch and post up what kind of testing I'd like to see here.

busbey:
release branch blocked on #306

busbey:
Here's my plan for driving the release:

1. Branch 0.2.0 (**0.2.0-staging**) and tag version 0.2.0-RC1 (source only)
1. Testers should build and use either the distribution artifact (**ycsb-0.2.0-RC1.tar.gz**) or a binding specific distribution (e.g. **ycsb-hbase-0.2.0-RC1.tar.gz**)
1. Testers should file issues/prs for problems found
1. Fixes for issues should go to master. I'll pull things back to the release staging branch. if you are feeling generous feel free to cherry-pick bug fixes back, but no new functionality please (adding a README.md so that a datastore gets its own distribution tarball is fine)
1. When issues come up, I'll start cutting new release candidates every other day off of the 0.2.0-staging branch.
1. When the final release is published, I'll delete the branch. The tag from the release will remain as a starting point if we need to make a patch release.


Successful testing of a datastore binding should be noted here. please include:
  * datastore binding (e.g. hbase)
  * version of datastore tested against (e.g. hbase 1.0.0 cluster)
  * if the datastore has different single node / clustered modes, which was used (e.g. hbase standalone)
  * OS for client
[...]

PR #275, opened 2015-06-02
Given the min and max parameter the ZipfianGenerator should not return *0*…

steffenfriedrich:
… `if (uz < 1)` nor **1** `if (uz<1.0+Math.pow(0.5,theta))`. The pseudo code in <cite>[1]</cite> (referenced by ycsb) assumes that only values between 1 and N are drawn from the distribution, but given a min parameter the min value should be returned `if (uz < 1)`, resp. min + 1 `if (uz<1.0+Math.pow(0.5,theta))`.

[1]  J. Gray et al. Quickly generating billion-record synthetic databases. In SIGMOD, 1994.

### DIFF
--- a/core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
+++ b/core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
@@ -1,0 +1,23 @@
+package com.yahoo.ycsb.generator;
+
+import org.testng.annotations.Test;
+import static org.testng.AssertJUnit.assertTrue;
+
+public class TestZipfianGenerator {
+    @Test
+    public void testMinParameter() {
+        long min = 5;
+        long max = 100;
+        Boolean gr_max = true;
+
+        ZipfianGenerator zipfian = new ZipfianGenerator(min, max);
+
+        for (int i = 0; i < 1000; i++) {
+            long rnd = zipfian.nextLong();
+            // System.out.println(rnd);
+            if(rnd < min) gr_max = false;
+        }
+
+        assertTrue(gr_max);
+    }
+}
--- a/core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
+++ b/core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
@@ -1,23 +1,39 @@
+/**
+ * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
```diff
@@ -0,0 +1,23 @@
+package com.yahoo.ycsb.generator;
```
busbey: this file needs a license header.

core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.testng.annotations.Test;
+import static org.testng.AssertJUnit.assertTrue;
+
+public class TestZipfianGenerator {
+    @Test
+    public void testMinParameter() {
```
busbey: should we similarly be testing the Max parameter?

core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+        long min = 5;
+        long max = 100;
+        Boolean gr_max = true;
+
+        ZipfianGenerator zipfian = new ZipfianGenerator(min, max);
+
+        for (int i = 0; i < 1000; i++) {
+            long rnd = zipfian.nextLong();
+            // System.out.println(rnd);
+            if(rnd < min) gr_max = false;
```
busbey: any reason not to just assert directly here?

core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+    public void testMinParameter() {
+        long min = 5;
+        long max = 100;
+        Boolean gr_max = true;
+
+        ZipfianGenerator zipfian = new ZipfianGenerator(min, max);
+
+        for (int i = 0; i < 1000; i++) {
+            long rnd = zipfian.nextLong();
+            // System.out.println(rnd);
```
busbey: just leave out the commented out bit.

core/src/test/java/com/yahoo/ycsb/generator/TestZipfianGenerator.java
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
```
busbey: nit: the year here should be the year of authorship. I'm fine fixing this while doing a sanity check on license headers as a part of #266. Let me know if you want to fix it now instead. If I don't hear anything I'll merge this tonight.

steffenfriedrich: Sorry for that. Please feel free to fix it as part of #266.

PR #276, opened 2015-06-03
Clean up bin/ycsb.

cmccoy:
* Use argparse for argument parsing.
* Add a '--jvm-args' flag, to pass arbirary options to java.
* Add a '-cp' flag, which allows prepending paths to the java classpath.
* Exit with the return code from java.

Requires Python 2.7.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -1,6 +1,9 @@
 #!/usr/bin/env python
 
+import argparse
+import io
 import os
+import shlex
 import sys
 import subprocess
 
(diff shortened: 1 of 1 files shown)


busbey:
nit: could you include a prefix in the commit message's first line for what got altered. We don't have a good list of these yet, but something like [deployment], [packaging], or [shellcode] would all be great.

(this is one of the things I want to make sure we cover in the contribution guide, but I recognize without that document for guidance this will be inconsistent for some time.)

cmccoy:
Sure - added `[scripts]`. Seem OK?

bin/ycsb
```diff
@@ -43,40 +46,46 @@ DATABASES = {
...
-    "-s"           : "Print status to stderr",
-    "-target n"    : "Target ops/sec (default: unthrottled)",
-    "-threads n"   : "Number of client threads (default: 1)",
+    "-P file"         : "Specify workload file",
+    "-p key=value"    : "Override workload property",
+    "-s"              : "Print status to stderr",
+    "-target n"       : "Target ops/sec (default: unthrottled)",
+    "-threads n"      : "Number of client threads (default: 1)",
+    "-cp path"        : "Additional Java classpath entries",
+    "--jvm-args args" : "Additional arguments to the JVM",
```
busbey: above we use single dash for long-form arguments, and here we're using gnu style '--'. I don't have a strong preference one way or the other, but we should be consistent.

busbey:
yeah [scripts] is fine by me.

bin/ycsb
cmccoy: I used `-cp` since it gets passed directly to `java`.  Seems like single dash long form is more consistent - I'll change `--jvm-args` to `-jvm-args`.

cmccoy:
I tested this against HBase:

    bin/ycsb -jvm-args='-Xmx1024m' -cp /tmp/hbase-1.0.1.1/conf load hbase -P workloads/workloadc -p columnfamily=cf 

worked for me.

busbey:
:+1: presuming Travis comes back clean.

PR #281, opened 2015-06-06
[scripts] remove defunct ycsb.sh

busbey:
We should have a single entry point for our client. Right now we have two and it's clear the python one is being better maintained.

While I'd prefer the portability of having things use shell scripts, we're not there right now and having a partially-done version is confusing.


cmccoy:
Looks good to me.

PR #285, opened 2015-06-06
Fix #284 - Switch to travis starting the test mongod instance.

### LINKED ISSUE #284: Use the Travis configuration to start the test MongoDB instance.
The MongoDB test cases need a MongoDB instance to connect to. Currently they are using the [Embedmongo.flapdoodle.de](http://flapdoodle-oss.github.io/de.flapdoodle.embed.mongo/) jar to start the MongoDB instance. 

This works fine but leaves copies the MongoDB distribution in the user's home directory which is at a minimum surprising if not obnoxious. 

Instead of starting a MongoDB instance manually the tests should instead test for the existence of a process running on the default MongoDB port and only run the tests if the connection is successful. The travis configuration can be changed to start the MongoDB service for the automated tests.

allanbank:


mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java
```diff
@@ -272,7 +275,8 @@ public int insert(String table, String key,
                 BulkWriteResult result = collection.withWriteConcern(
                         writeConcern).bulkWrite(bulkInserts,
                         new BulkWriteOptions().ordered(false));
-                if (result.getInsertedCount() == bulkInserts.size()) {
+                if (!result.wasAcknowledged()
```
cmccoy: Is this correct? It looks like you'll return 0 if the result was *not* acknowledged.

allanbank: The short answer is yes.

A few things to make sure we are on the same page.
* 0 is "success"
* the "wasAcknowledged()" returns true if the user requested an 
acknowledgment from the server via the WriteConcern.

So this is saying return success if the user did _not_ request an 
acknowledgment from the server.

The user of YCSB controls if an acknowledgment is requested via the 
MongoDB URL's WriteConcern.  w=0 in the URL (e.g., 
mongodb://locahost:27017/?w=0) mean no acknowledgment. Anything else 
requests an acknowledgment, including w=1 which is the default.

For completeness, the second half of the || handles the case where there 
is an requested acknowledgment by checking the inserted count.

Rob.


On 06/07/2015 01:45 PM, Connor McCoy wrote:
> In mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java
> <https://github.com/brianfrankcooper/YCSB/pull/285#discussion_r31878835>:
>
>> @@ -272,7 +275,8 @@ public int insert(String table, String key,
>>                  BulkWriteResult result = collection.withWriteConcern(
>>                          writeConcern).bulkWrite(bulkInserts,
>>                          new BulkWriteOptions().ordered(false));
>> -                if (result.getInsertedCount() == bulkInserts.size()) {
>> +                if (!result.wasAcknowledged()
>
> Is this correct? It looks like you'll return 0 if the result was /not/
> acknowledged.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/brianfrankcooper/YCSB/pull/285/files#r31878835>.
>


-- 
Robert J. Moore
President
Allanbank Consulting, Inc.
Robert.J.Moore@allanbank.com

cmccoy: Ah, I see. I was interpreting it as a result, rather than an indicator of the write configuration. Thanks for the explanation.

PR #287, opened 2015-06-08
[exporter] Fixed deprecated 'org.codehaus.jackson.impl.DefaultPrettyPrinter' ...

steffenfriedrich:
…  and changed  json export  to an array of measurement objects. 
Once again a small change. I wanted to parse an exported json file, but  with the `readTree()` method (see my test case) i got only the first measurement value ("Operations"), because the json export consists of many small objects which are not separated by comas. My suggestion is to export the measurements as an json array, so you are able to simply parse and plot the histogram, e.g. in javascript.


busbey:
Please make this a new exporter rather than changing the old one. There are probably lots of folks expecting the "JSON object per line" serialization, even if it's less easy to use.

Issue #279, opened 2015-06-06
BytesIterator creates Strings incorrectly


busbey:
Noticed this in BytesIterator while browsing for another issue

    /** Consumes remaining contents of this object, and returns them as a string. */
    public String toString() {
      StringBuilder sb = new StringBuilder();
      while(this.hasNext()) { sb.append((char)nextByte()); }
      return sb.toString();
    }

Each byte does not correspond to a character. We should change this to use a defined Charset decoder.

PR #288, opened 2015-06-08
[core] according to issue #279 RandomByteIterator now generates also …

steffenfriedrich:
…some characters. See #279.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/ByteIterator.java
+++ b/core/src/main/java/com/yahoo/ycsb/ByteIterator.java
@@ -16,8 +16,10 @@
  */
 package com.yahoo.ycsb;
 
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
 import java.util.Iterator;
-import java.util.ArrayList;
 /**
  * YCSB-specific buffer class.  ByteIterators are designed to support
  * efficient field generation, and to allow backend drivers that can stream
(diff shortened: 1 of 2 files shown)


busbey:
This doesn't help the general problem with BytesIterator in #279. Other workloads could put anything they like in a ByteIterator (as could other data stores) and we should do a better job of turning htings into a String.

Specifically, in ByteIterator we should be using a CharsetDecoder (either for UTF8 or a user configurable encoding that defaults to UTF8). (and similarly if we need to go from String -> byte we should use a CharsetEncoder)

steffenfriedrich:
I see, so I changed the the `toString()` method of the ByteIterator to decode the byte array by using the java.nio.charset. This way the iterator is still be consumed by using its `toArray()` method and additionally the `toString()` method is also faster (according to a microbenchmark I did).

busbey:
The new toString looks great. Why still change the way RandomBytesIterator creates its sequence of bytes?

steffenfriedrich:
Because the old one generated only characters between *space* and *?* (http://www.asciitable.com/). Now the buffer is also filled with some letters because of the `& 63` and `& 95`.

steffenfriedrich:
Should I revert the RandomByteIterator?

busbey:
Sorry, this last week got busy even for me. We just need to confirm the new ranges, I expect it'll be fine. If @gkamat or @cmccoy have time to check it it'll get in sooner.

Eventually I want to change RandomBytesIterator to be random binary (and have a "random printable" or some such for the role the class plays now). But I don't think that burden should fall on this ticket, especially since no one has gone through the ramifications of doing such a change.

Issue #250, opened 2015-05-19
Improve deployment of optional dependencies


busbey:
We have DB bindings that need different optional dependencies. Currently we rely on users placing jars in a particular location on the filesystem which is error prone (see #232 ).

Get a better story together around this. Needs to account for the fact that different bindings might need conflicting versions of dependencies.

busbey:
maybe the answer to this is classpath isolation of the drivers? then we could just build all of them as with-dependencies and only load the classpath of the module in use.

busbey:
Several of the modules rely on jar - with - dependencies. We should clean this up as a part of this ticket, so that we can make it more apparent to an end user what dependencies are on their system. We cut hopefully also down on our binary distribution size by deduplicating.

nitsanw:
Not sure you want to tackle the deployment model as part of this work, but I would think ideally we'd want a plugin model which allows:
- Separate development and versioning of core/plugins
- A compatibility API for the core(should be the DB API)
- Packaging of vendor specific deployable client (core + plugin)
This will result in faster turnaround for vendor bug fixes/contribution in their own plugin area and reduce pressure on core releases from vendor PRs.

busbey:
> * Separate development and versioning of core/plugins
> * Packaging of vendor specific deployable client (core + plugin)

I think having convenience binaries that are just the deployable bits for particular vendors would be great at easing getting started. We get tons of questions that are essentially "I just want to use datastore X, what am I doing wrong with these thousand configuration knobs."

I have some strong reservations about versioning things independently though. It complicates providing a reference point for folks to reproduce results. As opposed to just 'use YCSB version foo', now folks have to track down particular bindings that might come from lord-knows-where. It also makes our testing way more complicated (because we'd need to test X versions of core with Y versions of Z plugins)

> * A compatibility API for the core(should be the DB API)

We definitely need a defined API, for workloads and exporters as as well. Defining our API needs to be its own issue. I thought we had one. Please feel free to take a stab at defining one.

Personally, I've found the annotation-based approach of the Hadoop and HBase projects super easy to maintain. Essentially we make an annotation for "this is public" and "this is not public" and then label classes with them. There's even tooling to generate appropriately scoped javadocs that separate out the two.

busbey:
Status update from this weekend: I got started on updating the distribution but ran into some troubles with the maven assembly plugin. Should still be able to finish by end of week, so the 0.2.0 branch date isn't at risk.

busbey:
[...]

PR #292, opened 2015-06-11
Add Tarantool database

bigbes:
http://tarantool.org and https://github.com/tarantool/tarantool
In "mail.ru" we wrote and widely use Tarantool key-value database.
It's key properties include:

* Defferent index types with iterators:
    - HASH (the fastest)
    - TREE (range and ordered retreival)
    - BITSET (bit mask search)
    - RTREE (geo search)
* multipart keys for HASH and TREE indexes
* Data persistence with by Write Ahead Log (WAL) and snapshots.
* asynchronous master-master replication, hot standby.
* coroutines and async. IO are used to implement high-performance lock-free access to data.
    - socket-io/file-io with yeilds from lua
* stored procedures in Lua (Using LuaJIT)
* supports plugins written on C/C++ (Have two basic plugins for working with MySQL and PostgreSQL)
* Authentication and access control

Move 'distribution target' to the end (it's needed for .jar to be in the .tar.gz)


bigbes:
@busbey i've rebased and pushed latest version. Tell me about everything that i must fix. Thanks.

pom.xml
```diff
@@ -67,6 +67,7 @@
     <thrift.version>0.8.0</thrift.version>
     <hypertable.version>0.9.5.6</hypertable.version>
     <couchbase.version>1.1.8</couchbase.version>
+    <tarantool.version>1.6.1-SNAPSHOT</tarantool.version>
```
busbey: is there a non-SNAPSHOT version we could use here instead? we're starting the release process on Monday and expect it to take no more than 2 weeks. When we release, all versions will have to be non-SNAPSHOT.

pom.xml
```diff
@@ -85,10 +86,11 @@
     <module>orientdb</module>
     <module>redis</module>
     <module>voldemort</module>
-    <module>distribution</module>
     <!--<module>mapkeeper</module>-->
     <!--module>nosqldb</module-->
     <module>couchbase</module>
+    <module>tarantool</module>
+    <module>distribution</module>
```
busbey: nit: distribution needn't change location.

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -0,0 +1,168 @@
+package com.yahoo.ycsb.db;
```
busbey: file needs a license header

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -0,0 +1,168 @@
...
+		if (spaceString != null) {
+			spaceNo = Integer.parseInt(spaceString);
+		}
+
+		try {
+			this.connection = new TarantoolConnection16Impl(host, port);
+		}
+		catch (Exception exc) {
+			System.err.println("Can't init Tarantool connection:" + exc.toString());
+			exc.printStackTrace();
```
busbey: nit: this should use your Logger and log at FATAL rather than writing to System.err and printing a stacktrace.

(nits are suggestions that you needn't fix if you're pressed for time or you disagree)

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -0,0 +1,168 @@
...
+		tuple[0] = key;
+		for (Map.Entry<String, ByteIterator> i: values.entrySet()) {
+			tuple[j + 1] = i.getKey();
+			tuple[j + 2] = i.getValue().toString();
+			j += 2;
+		}
+		try {
+			this.connection.insert(this.spaceNo, tuple);
+		} catch (TarantoolException e) {
+			e.printStackTrace();
```
busbey: nit: as above, log the exception via your logger rather than printing the stacktrace yourself

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -0,0 +1,168 @@
...
+	@Override
+	public int read(String table, String key, Set<String> fields,
+			HashMap<String, ByteIterator> result) {
+		try {
+			List<String> response;
+			response = this.connection.select(this.spaceNo, 0, Arrays.asList(key), 0, 1, 0);
+			result = tuple_convert_filter(response, fields);
+			return 0;
+		} catch (TarantoolException e) {
+			e.printStackTrace();
```
busbey: nit: as above, log the exception via your logger rather than printing the stacktrace yourself

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -0,0 +1,168 @@
...
+
+	@Override
+	public int scan(String table, String startkey,
+			int recordcount, Set<String> fields,
+			Vector<HashMap<String, ByteIterator>> result) {
+		List<String> response;
+		try {
+			response = this.connection.select(this.spaceNo, 0, Arrays.asList(startkey), 0, recordcount, 6);
+		} catch (TarantoolException e) {
+			e.printStackTrace();
```
busbey: nit: as above, log the exception via your logger rather than printing the stacktrace yourself

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -0,0 +1,168 @@
...
+			result.add((HashMap<String, ByteIterator>) temp.clone());
+		return 0;
+	}
+
+	@Override
+	public int delete(String table, String key) {
+		try {
+			this.connection.delete(this.spaceNo, Arrays.asList(key));
+		} catch (TarantoolException e) {
+			e.printStackTrace();
```
busbey: nit: as above, log the exception via your logger rather than printing the stacktrace yourself

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -0,0 +1,168 @@
...
+		tuple[0] = key;
+		for (Map.Entry<String, ByteIterator> i: values.entrySet()) {
+			tuple[j + 1] = i.getKey();
+			tuple[j + 2] = i.getValue().toString();
+			j += 2;
+		}
+		try {
+			this.connection.replace(this.spaceNo, tuple);
+		} catch (TarantoolException e) {
+			e.printStackTrace();
```
busbey: nit: as above, log the exception via your logger rather than printing the stacktrace yourself

pom.xml
bigbes: I've tried to build without changing `<module>distribution</module>` position, but i can't, tarantool/couchbase jar weren't in .tar.gz file.

pom.xml
bigbes: I'll figure something out.

bigbes:
Okay, i've fixed all.

pom.xml
busbey: Yeah, that's a problem with how we're building our distribution. It'll be fixed properly in #250, so leaving this work-around here is fine.

busbey:
I just noticed the lua scripts are in `tarantool/config/`. Since users need to install them in their Tarantool cluster, they'll need to be in the binary artifact. To get included there, they need to be in `tarantool/src/main/conf`.

I'm fine with this being done either as a follow-on or as another change here. Just let me know which you prefer.

bigbes:
I've pushed commit, that moves configuration files.

tarantool/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Quick start
+
+This section descrives how to run YCSB against a local Tarantool instance
+
+### 1. Start Tarantool
+
+First, clone Tarantool from it's own git repo and build it (described in our [README.md][tarantool-readme]):
+
+    cp YCSB/tarantool/config/tarantool-tree.lua <vardir>/tarantool.lua
```
busbey: this refers to the wrong location now. in the YCSB binary artifact, the location will be %YCSB%/tarantool/conf/tarantool-tree.lua

bigbes:
@busbey, yes, i'm sorry. It's late night here and i'm sleepy:)

busbey:
Thanks for the contribution!

PR #296, opened 2015-06-13
Due to recent changes in MongoDB Client node, the mongodb.url connection string has been changed to mongodatabase.url. Updating the READme file

saggarsunil:
… in Mongo Client code


busbey:
Instead, let's change things to actually use 'mongodb.url'.

All the other properties start with 'mongodb' and the async client even still uses it.

saggarsunil:
Sure. I thought of the same. I will close it and open a new pull request.

Issue #278, opened 2015-06-05
YCSB does not throw error if there is no table created


ramkrish86:
If we run the ycsb tool without creating the table the tool does not throw any error. It just runs and infact displays some statistics also. But internally nothing actually works.

busbey:
which data store binding are you running against? which version of YCSB? could you include the command line invocation?

ramkrish86:
I have a YCSB version pulled out of this
https://github.com/apurtell/YCSB/tree/new_hbase_client and it is running against 0.98.120hadoop2.

Am running hbase store. 
the command line invocation is this
./ycsb load hbase -P ../workloads/workloada -p columnfamily=f1 -p recordcount=5000000 -p threadcount=20 -s | tee -a workloada.dat

busbey:
please try to recreate the problem with the current hbase client.

    bin/ycsb -jvm-args='-Xmx1024m' -cp "${HBASE_CONF_DIR}" load hbase -P workloads/workloada -p columnfamily=f1 -p recordcount=5000000 -p threadcount=20 -s

Or if you prefer the new API:

    bin/ycsb -jvm-args='-Xmx1024m' -cp "${HBASE_CONF_DIR}" load hbase-10 -P workloads/workloada -p columnfamily=f1 -p recordcount=5000000 -p threadcount=20 -s

gkamat:
The TableNotFound exception is indeed reported; however, the problem is that an error status reported by a transaction in the client stub is not propagated upstream to the core module.  Therefore, YCSB will continue to attempt the specified number of transactions rather than terminating.

Will create a category of errors on encountering which YCSB should terminate; that should fix this issue.

busbey:
How about instead we tell datastore bindings that they should check for things like this in their init call? At that point the binding can throw a DBException to indicate the problem and things ought to terminate.

gkamat:
Submitted pull request #298.

busbey:
closed by #304

PR #298, opened 2015-06-17
Issue #278: YCSB does not throw error if there is no table created

gkamat:
Check that the table exists during DB binding initialization.  If this is tested for here, the workload will continue running until it finishes.


busbey:
nit: please start the commit message with "[hbase]" to indicate what's impacted.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
```diff
@@ -71,6 +72,16 @@
     public static final Object tableLock = new Object();
 
     /**
+     * The name of the database table to run queries against.
+     */
+    public static final String TABLENAME_PROPERTY="table";
```
busbey: could we reference these from CoreWorkload instead?

gkamat:
Having an issue amending the comment.  Closing this pull request and creating a new one.

PR #300, opened 2015-06-17
refactor packaging to avoid fat-jars.

busbey:
This PR fixes the key issues for #250 and does some minor related cleanup:

* remove unused declared deps in a few places
* lex sort modules in top level pom
* update maven plugin versions.
* no more jar-with-dependencies
* set ycsb-core to provided in bindings
* introduce a bindings-parent pom that sets
** stage dependency copy on building binding
** iff binding has a README.md, build a binding-specific dist artifact
* update distribution assembly to
** properly build after modules
** use per-binding sets of dependencies

TODO:
* update ycsb launcher script to only load those jars needed for the requested binding.


binding-parent/datastore-specific-descriptor/src/main/resources/assemblies/datastore-specific-assembly.xml
```diff
@@ -0,0 +1,57 @@
...
+      <outputDirectory></outputDirectory>
+      <fileMode>0644</fileMode>
+      <includes>
+        <include>CHANGELOG</include>
+        <include>LICENSE.txt</include>
+        <include>NOTICE.txt</include>
+      </includes>
+    </fileSet>
+    <fileSet>
+      <directory>../bin</directory>
```
cmccoy: What do you think about adding an `<include>ycsb</include>` here?  I had some backups from merging in `bin/`, which got carried into the distribution.

distribution/src/main/assembly/distribution.xml
```diff
@@ -28,36 +28,60 @@
...
+        <exclude>com.yahoo.ycsb:binding-parent</exclude>
+        <exclude>com.yahoo.ycsb:datastore-specific-descriptor</exclude>
+        <exclude>com.yahoo.ycsb:ycsb</exclude>
+        <exclude>com.yahoo.ycsb:root</exclude>
+      </excludes>
       <sources>
-        <includeModuleDirectory>true</includeModuleDirectory>
-
+<!--
+        <outputDirectory>${module.artifactId}</outputDirectory>
```
cmccoy: `outputDirectory` is added below - maybe just drop this?

distribution/src/main/assembly/distribution.xml
```diff
@@ -28,36 +28,60 @@
...
           </fileSet>
         </fileSets>
       </sources>
+      <binaries>
+        <includeDependencies>false</includeDependencies>
+<!--
+        <excludes>
+          <exclude>com.yahoo.ycsb:core:jar:*</exclude>
+        </excludes>
+-->
```
cmccoy: Maybe just delete this if core is included?

cmccoy:
Nice. Two things:

1. This increases build time for me by a fair bit. Maybe we could add instructions for building the full distribution and a single binding to the top-level README?  I took a stab in cmccoy@dbda161e7bf0ca27b4a43ee08131467d5cc0cd60, but infinispan fails.
1. I think some of the database binding READMEs now give invalid instructions on building the project. Fixes in cmccoy@bfe068a25605a6984a54dd50dea8c7a65b6f71ea.

binding-parent/datastore-specific-descriptor/src/main/resources/assemblies/datastore-specific-assembly.xml
busbey: sure, I could do that. I'd guess this was a left over from when we had the two scripts. Can always switch back if we end up with more things in the directory.

I'd like it even more to remove the top level bin and keep the ycsb script in src/main/python or some such and only have bin made in the distro, but I think folks rely on being able to run directly from source check out too often right now. :/

busbey:
It should be possible to build just a single binding, but I think the command line would have to include telling it to build core and binding-parent as well. Okay if we figure that out as a follow on?

cmccoy:
> It should be possible to build just a single binding, but I think the command line would have to include telling it to build core and binding-parent as well. Okay if we figure that out as a follow on?

Sure, sounds good.  FWIW, commands like:

    mvn -pl com.yahoo.ycsb:mongodb-binding -am clean package

build single binding tarballs for me.

busbey:
interesting. even without doing a previous install? something like `mvn -Dmaven.repo.local=/tmp/some_path_you_make -pl com.yahoo.ycsb:mongodb-binding -am clean package` would make sure you aren't pulling in jars from previous builds (provided you clean out some_path_you_make before building).

Maybe our intra-project dependencies are actually set up correctly! :smile:

cmccoy:
> interesting. even without doing a previous install? something like mvn -Dmaven.repo.local=/tmp/some_path_you_make -pl com.yahoo.ycsb:mongodb-binding -am clean package would make sure you aren't pulling in jars from previous builds (provided you clean out some_path_you_make before building).

It does!

busbey:
How's the update look @cmccoy ? If it's good I'll push and you can make a PR with your README changes?

cmccoy:
Looks good.

cmccoy:
README update in #303

PR #304, opened 2015-06-17
Issue #278: YCSB does not throw error if there is no table created

gkamat:
[hbase] Verify that the table exists during DB binding initialization
and bail out if not found.  If this is not checked here, the workload
will continue running and generating errors until it finishes.


hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -133,6 +135,20 @@ public void init() throws DBException
...
         _columnFamilyBytes = Bytes.toBytes(_columnFamily);
+
+      // Terminate right now if table does not exist, since the client
+      // will not propagate this error upstream once the workload
+      // starts.
+      String table = com.yahoo.ycsb.workloads.CoreWorkload.table;
+      try
+	  {
+	      HTable ht = new HTable(config, table);
+	      HTableDescriptor dsc = ht.getTableDescriptor();
```
cmccoy: We might want to use [`Connection.getTable(...).getTableDescriptor()`](https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/client/Table.html#getTableDescriptor()) for the 1.0 client.

gkamat: Updated the change for the 1.0 client.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
```diff
@@ -99,6 +100,19 @@ public void init() throws DBException
...
+      // starts.
+      String table = com.yahoo.ycsb.workloads.CoreWorkload.table;
+      try
+	  {
+	      HTable ht = new HTable(config, table);
+	      HTableDescriptor dsc = ht.getTableDescriptor();
+	  }
+      catch (IOException e)
+	  {
+	      throw new DBException("Error accessing HBase table: " + table);
```
busbey: this should pass the source IOException in as the cause.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -133,6 +135,20 @@ public void init() throws DBException
...
+      // starts.
+      String table = com.yahoo.ycsb.workloads.CoreWorkload.table;
+      try
+	  {
+	      final TableName tableName = TableName.valueOf(table);
+	      HTableDescriptor dsc = _connection.getTable(tableName).getTableDescriptor();
+	  }
+      catch (IOException e)
+	  {
+	      throw new DBException("Error accessing HBase table: " + table);
```
busbey: this should pass the source IOException in as the cause.

busbey:
+1

Issue #302, opened 2015-06-17
ycsb launcher should only include binding specific jars in the classpath


busbey:
right now when you use the ycsb launcher, we build a classpath with all the jars from all the bindings.

instead, find_jars should build a classpath with just the core libraries (in lib/) and those for the selected binding (in _datastore_-binding/lib as of #300)

gkamat:
That's what seems to be happening with the current launcher.  For instance, when running a HBase workload, the following jars are included in the classpath:
.../ycsb-0.2.0-SNAPSHOT/core/lib/core-0.2.0-SNAPSHOT.jar:.../ycsb-0.2.0-SNAPSHOT/hbase-binding/conf:.../ycsb-0.2.0-SNAPSHOT/hbase-binding/lib/hbase-binding-0.2.0-SNAPSHOT.jar

Perhaps this was different in the launcher before a recent change?

busbey:
that's with the pre-#300 launcher? or with the current launcher in the hbase dist tarball? or with the current launcher in the everything-dist tarball?

gkamat:
This was the current launcher, but I had changed the pom.xml to build only HBase.  Reverting it to build everything still includes only the core and hbase-binding jars (but all conf directories -- that could be fixed, if needed):

.../ycsb-0.2.0-SNAPSHOT/gemfire-binding/conf:.../ycsb-0.2.0-SNAPSHOT/core/lib/core-0.2.0-SNAPSHOT.jar:.../ycsb-0.2.0-SNAPSHOT/voldemort-binding/conf:.../ycsb-0.2.0-SNAPSHOT/tarantool-binding/conf:.../ycsb-0.2.0-SNAPSHOT/jdbc-binding/conf:.../ycsb-0.2.0-SNAPSHOT/accumulo-binding/conf:.../ycsb-0.2.0-SNAPSHOT/hbase-binding/conf:.../ycsb-0.2.0-SNAPSHOT/hbase-binding/lib/hbase-binding-0.2.0-SNAPSHOT.jar

cmccoy:
This is biting me - `accumulo` and `hbase` include different hadoop versions.

busbey:
Thinking about this more, I think this is a regression . before the
packaging refactoring, all the bindings were jar - with - dependencies, so
excluding the other binding jars was enough.

@gkamat, will you have time to work through this soon? If not, would you
mind if I made a patch to fix it (but not the conf dir part)?

gkamat:
[...]

PR #306, opened 2015-06-18
[packaging] ensure datastore binding specific distributions include core transitive deps

busbey:
found an error in our packaging while reviewing #302 that keeps all of our datastore binding specific distribution artifacts from working.

this should block starting the release branch for #266, since we know it breaks so much.


busbey:
tested by building and inspecting the generated .tar.gz files.

cmccoy:
:+1:

PR #113, opened 2013-01-21
Aerospike support

gelin:
This change add support of Aerospike (formely Citrusleaf) database to YCSB.

PR #324, opened 2015-06-25
Support Aerospike

tlopatic:
Hey guys,

We've been using YCSB quite a bit at Aerospike lately. It would be nice, if we could get support for our database into the upstream YCSB code.

Thomas

### DIFF
--- a/aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
+++ b/aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
@@ -29,7 +29,7 @@ public class AerospikeClient extends com.yahoo.ycsb.DB {
   private static final String DEFAULT_NAMESPACE = "ycsb";
 
   private static final int RESULT_OK = 0;
-  private static final int RESULT_ERROR = 1;
+  private static final int RESULT_ERROR = -1;
 
   private static final int WRITE_OVERLOAD_DELAY = 5;
   private static final int WRITE_OVERLOAD_TRIES = 3;
--- a/aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
+++ b/aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
@@ -1,0 +1,205 @@
+package com.yahoo.ycsb.db;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.Vector;
+
+import com.aerospike.client.AerospikeException;
+import com.aerospike.client.Bin;
+import com.aerospike.client.Key;
+import com.aerospike.client.Record;
+import com.aerospike.client.ResultCode;
+import com.aerospike.client.policy.ClientPolicy;
+import com.aerospike.client.policy.Policy;
+import com.aerospike.client.policy.RecordExistsAction;
+import com.aerospike.client.policy.WritePolicy;
+
+import com.yahoo.ycsb.ByteArrayByteIterator;
+import com.yahoo.ycsb.ByteIterator;
+import com.yahoo.ycsb.DBException;
+
+public class AerospikeClient extends com.yahoo.ycsb.DB {
+  private static final boolean DEBUG = false;
+
+  private static final String DEFAULT_HOST = "localhost";
+  private static final String DEFAULT_PORT = "3000";
+  private static final String DEFAULT_TIMEOUT = "10000";
+  private static final String DEFAULT_NAMESPACE = "ycsb";
+
+  private static final int RESULT_OK = 0;
+  private static final int RESULT_ERROR = 1;
+
+  private static final int WRITE_OVERLOAD_DELAY = 5;
+  private static final int WRITE_OVERLOAD_TRIES = 3;
+
+  private String namespace = null;
+
+  private com.aerospike.client.AerospikeClient client = null;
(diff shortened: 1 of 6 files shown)


aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
```diff
@@ -0,0 +1,207 @@
...
+public class AerospikeClient extends com.yahoo.ycsb.DB {
+  private static final boolean DEBUG = false;
+
+  private static final String DEFAULT_HOST = "localhost";
+  private static final String DEFAULT_PORT = "3000";
+  private static final String DEFAULT_TIMEOUT = "10000";
+  private static final String DEFAULT_NAMESPACE = "ycsb";
+
+  private static final int RESULT_OK = 0;
+  private static final int RESULT_ERROR = 1;
```
cmccoy: These should be unified eventually, but negative values seem more prevalent for errors in most YCSB clients - consider using -1.

aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
```diff
@@ -0,0 +1,207 @@
...
+        if (DEBUG) {
+          System.err.println("Record key " + key + " not found (delete)");
+        }
+
+        return RESULT_ERROR;
+      }
+
+      return RESULT_OK;
+    } catch (AerospikeException e) {
+      System.err.println("Error while deleting key " + key + ": " + e);
```
cmccoy: Wrap this in an `if(DEBUG)`?

aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
```diff
@@ -0,0 +1,207 @@
...
+    Key keyObj = new Key(namespace, table, key);
+
+    while (true) {
+      try {
+        client.put(writePolicy, keyObj, bins);
+        writeOverloadTries = WRITE_OVERLOAD_TRIES;
+        return RESULT_OK;
+      } catch (AerospikeException e) {
+        if (e.getResultCode() != ResultCode.DEVICE_OVERLOAD) {
+          System.err.println("Error while writing key " + key + ": " + e);
```
cmccoy: `if(DEBUG)?`

cmccoy:
Nice! This will be great to have. Left a few comments.

tlopatic:
This is exciting!

Thanks for the constructive feedback.

A bit on the DEBUG flag: Yes, its use isn't entirely consistent right now. Generally, the idea is to use it to distinguish between expected/recoverable and unexpected errors.

Accordingly, the "Record key ... not found" error in read() should not be wrapped in DEBUG. We should never run into a situation where we read a record that doesn't exist. However, there seems to be a race condition in workload D that causes exactly that.

The client threads generate keys for the records to be inserted and there is a small time window between the thread generating the key and it actually inserting the record. However, as soon as the key has been generated, that key is considered to be available for other threads to read. (Because transactioninsertkeysequence.lastInt() now returns that key.) That results in the other threads potentially trying to read records that don't yet exist.

The right fix is, obviously, to take care of the race condition. I just didn't have time, yet, to think about this a little more in depth.

So, for now, I am mimicking the behavior of other drivers, which don't output an error message and just return an error code.

Or were you referring to something else regarding the flag?

Oh, and let me open a ticket regarding the race condition, so that it gets tracked.

tlopatic:
Oh, and I changed the error result to -1.

busbey:
whoever merges this, please not in the merge commit that it also closes #113

aerospike/src/main/java/com/yahoo/ycsb/db/AerospikeClient.java
```diff
@@ -0,0 +1,207 @@
...
+  @Override
+  public int scan(String table, String start, int count, Set<String> fields,
+      Vector<HashMap<String, ByteIterator>> result) {
+    System.err.println("Scan not implemented");
+    return RESULT_ERROR;
+  }
+
+  private int write(String table, String key, WritePolicy writePolicy,
+      HashMap<String, ByteIterator> values) {
+    if (writeOverloadTries == 0) {
```
cmccoy: A little description of the retry policy here would be very helpful :).

Is the intent here that once a write fails `WRITE_OVERLOAD_TRIES` consecutive times with error code `DEVICE_OVERLOAD`, no further writes will be issued by the client?

cmccoy:
@tlopatic  - looks great. Left one other comment, if you don't mind.

tlopatic:
Sorry for the delay. I was traveling.

Yes, your interpretation of the retry policy is correct. And, thinking about it again, it's a little... idiosyncratic. It did fit the specific way in which I was using YCSB, but it was probably a little counter-intuitive for everybody else.

So, instead of documenting its behavior, let's correct the behavior, so that it makes more sense.

Now each write is tried 3 times, followed by delays of 5, 10, and 20 ms, respectively. That's probably a little less surprising than the previous behavior. It'll allow people to see what fraction of their writes actually failed because of server overload. (Instead of failing every subsequent write, once an overload situation is encountered.)

Sound good?

cmccoy:
Great!

I do think it's a little counterintuitive that *any* retries happens within YCSB, but the Cassandra Thrift client also contains retry logic.  Opened #334 to discuss, merging this.

Thanks again, @tlopatic!

PR #329, opened 2015-06-26
[scripts] clean up runtime classpath.

### LINKED ISSUE #315: Cannot run YCSB from the source tree.
I am not finding the com.yahoo.ycsb.Client class when I run from the command line out of the source tree 

```
wget https://github.com/brianfrankcooper/YCSB/archive/0.2.0-RC2.tar.gz
tar xzvf 0.2.0-RC2.tar.gz
cd YCSB-0.2.0-RC2/
mvn clean package 
bin/ycsb load -db mongodb -s -P workloads/workloada -threads 10 -p "mongodb.url=mongodb://192.168.1.15:27017/ycsb?w=1&maxPoolSize=10&waitQueueMultiple=10"
java -cp /home/rjmoore/ycsb-testing/YCSB-0.2.0-RC2/mongodb-binding/conf com.yahoo.ycsb.Client -db com.yahoo.ycsb.db.MongoDbClient -db -s -P workloads/workloada -threads 10 -p mongodb.url=mongodb://192.168.1.15:27017/ycsb?w=1&maxPoolSize=10&waitQueueMultiple=10 -load
Error: Could not find or load main class com.yahoo.ycsb.Client
```

busbey:
closes #315 

In addition to not being able to run from source, the current classpath determination

* can't handle any case where the binding directory has a name different from the database specified (i.e. hbase-10, cassanda-cql, etc in the general purpose artifact)
* doesn't add the 'conf' dir for binding-specific artifacts to the classpath

Tested by running shell with tarantool, hbase, and accumulo from source and examining classpath, then doing the same from the general artifact, then doing the same from the tarantool and hbase specific artifacts.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import argparse
+import fnmatch
 import io
 import os
 import shlex
@@ -87,13 +88,20 @@ def usage():
 
     return output.getvalue()
 
+def debug(message):
+    print >> sys.stderr, "[DEBUG] ", message
 
-def find_jars(dir, database):
+def warn(message):
+    print >> sys.stderr, "[WARN] ", message
+
+def error(message):
+    print >> sys.stderr, "[ERROR] ", message
+
+def find_jars(dir, glob='*.jar'):
     jars = []
     for (dirpath, dirnames, filenames) in os.walk(dir):
-        for filename in filenames:
-            if filename.endswith(".jar"):
-                jars.append(os.path.join(dirpath, filename))
+        for filename in fnmatch.filter(filenames, glob):
+            jars.append(os.path.join(dirpath, filename))
     return jars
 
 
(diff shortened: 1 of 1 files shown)


bin/ycsb
```diff
@@ -103,6 +105,30 @@ def get_ycsb_home():
...
+    # If there's a top level pom, we're a source checkout. otherwise a dist artifact
+    return "pom.xml" not in os.listdir(get_ycsb_home())
+
+# Run the maven dependency plugin to get the local jar paths.
+# presumes maven can run, so should only be run on source checkouts
+# will invoke the 'package' goal for the given binding in order to resolve intra-project deps
+# presumes maven properly handles system-specific path separators
+def get_classpath_from_maven(database):
+    DEVNULL = open(os.devnull, 'w')
+    mvn = subprocess.Popen(["mvn", "-pl", "com.yahoo.ycsb:"+database+"-binding", "-am",
```
cmccoy: I'm worried this will be a bit brittle. Since stderr is redirected to `/dev/null`, and the `mvn` exit status isn't checked, failures will be a bit mysterious (and potentially unreported).

Is this script supposed to work on windows? I'm not sure about `cut/grep/tail` there.

bin/ycsb
```diff
@@ -129,10 +155,30 @@ def main():
...
+    if is_distribution():
+        db_dir = os.path.join(ycsb_home, binding + "-binding")
+        # include top-level conf for when we're a binding-specific artifact.
+        # If we add top-level conf to the general artifact, starting here
+        # will allow binding-specific conf to override (because it's prepended)
+        cp = [os.path.join(ycsb_home, "conf")]
+        cp.extend(find_jars(os.path.join(ycsb_home, "lib")))
+        cp.extend(find_jars(os.path.join(db_dir, "lib")))
+    else:
+        warn("Running against a source checkout. In order to get our runtime " \
```
cmccoy: The trailing \ here isn't needed.

bin/ycsb
busbey: It will work on cygwin. We have 1 open ticket of someone trying to run on bare windows without cygwin, but I'm not sure if we actually work in that environment as is.

I could switch to munging the output here, but it'll take a bit. Maybe we could make it a follow on?

busbey: it's definitely brittle. I'll update to check for errors.

bin/ycsb
busbey: really? well that's a nice surprise.

bin/ycsb
cmccoy: > I could switch to munging the output here, but it'll take a bit. Maybe we could make it a follow on?

Sure, sounds good.

busbey: turns out to get some decent error reporting from maven I had to take care of the output munging.

bin/ycsb
```diff
@@ -103,6 +108,33 @@ def get_ycsb_home():
...
+    # If there's a top level pom, we're a source checkout. otherwise a dist artifact
+    return "pom.xml" not in os.listdir(get_ycsb_home())
+
+# Run the maven dependency plugin to get the local jar paths.
+# presumes maven can run, so should only be run on source checkouts
+# will invoke the 'package' goal for the given binding in order to resolve intra-project deps
+# presumes maven properly handles system-specific path separators
+def get_classpath_from_maven(database):
+    try:
+        mvn_output = subprocess.check_output(["mvn", "-pl", "com.yahoo.ycsb:"+database+"-binding",
```
cmccoy: Nit: for debugging I think it would useful to print the command executed to stderr.

cmccoy:
Looks great - thanks!

busbey:
updated the "what does maven say the classpath should be" command to print a debug message with the command it's going to run.

cmccoy:
:+1:

PR #333, opened 2015-07-01
[hbase] Add kerberos authentication for HBaseClient.java

uijin:
For running YCSB on secure Hadoop.

### DIFF
--- a/hbase/README.md
+++ b/hbase/README.md
@@ -47,6 +47,20 @@ Please see the general instructions in the `doc` folder if you are not sure how 
 bin/ycsb run hbase -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family -p clientbuffering=true
 ```
 
+### 5. Kerberos
+For secure Hadoop, you could enable kerberos authentication option.
+
+```
+kinit -kt /path/to/keytab user@XXX.COM
+bin/ycsb run hbase -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family -p kerberos=true
+```
+
+Or specify kerberos principle and keytab as options:
+
+```
+bin/ycsb run hbase -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family -p kerberos=true -p kerberosuser=user@XXX.COM -p kerberoskeytab=/path/to/keytab
+```
+
 ## Configuration Options
 Following options can be configurable using `-p`.
 
@@ -54,3 +68,6 @@ Following options can be configurable using `-p`.
 * `clientbuffering` : If true, buffer mutations on the client. The default is false.
 * `writebuffersize` : Buffer size to be used when `clientbuffering` is activated. The default is 12MB.
 * `debug` : If true, debugging logs are activated. The default is false.
+* `kerberos` : If true, use Kerberos authentication.
+* `kerberosuser` : Kerberos principle [optional].
+* `kerberoskeytab`: Kerberos keytab path [optional].
--- a/hbase/README.md
+++ b/hbase/README.md
@@ -47,8 +47,8 @@ Please see the general instructions in the `doc` folder if you are not sure how 
 bin/ycsb run hbase -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family -p clientbuffering=true
 ```
 
-### 5. Kerberos
-For secure Hadoop, you could enable kerberos authentication option.
+### 5. Kerberos (optional)
+For secure HBase you can enable kerberos authentication option.
 
 ```
 kinit -kt /path/to/keytab user@XXX.COM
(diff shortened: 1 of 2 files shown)


hbase/README.md
```diff
@@ -47,6 +47,20 @@ Please see the general instructions in the `doc` folder if you are not sure how
 bin/ycsb run hbase -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family -p clientbuffering=true
 ```
 
+### 5. Kerberos
+For secure Hadoop, you could enable kerberos authentication option.
```
busbey: nit: "For secure HBase you can enable"

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
```diff
@@ -98,21 +109,30 @@ public void init() throws DBException
...
+        String table = com.yahoo.ycsb.workloads.CoreWorkload.table;
+        try
+        {
+            //if using kerberos for authentication, setup UserGroupInformation.
+            if (_useKerberos)
+            {
+                config.set("hadoop.security.authentication", "kerberos");
+                UserGroupInformation.setConfiguration(config);
+            }
+            if (_kerberosUser != null && _kerberosKeytab != null)
```
busbey: please include braces for the if block.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
```diff
@@ -98,21 +109,30 @@ public void init() throws DBException
...
-      try
-	  {
-	      HTable ht = new HTable(config, table);
-	      HTableDescriptor dsc = ht.getTableDescriptor();
-	  }
-      catch (IOException e)
-	  {
-	      throw new DBException(e);
-	  }
+        _columnFamilyBytes = Bytes.toBytes(_columnFamily);
```
busbey: please leave out unrelated formatting changes.

hbase/README.md
```diff
@@ -47,6 +47,20 @@ Please see the general instructions in the `doc` folder if you are not sure how
 bin/ycsb run hbase -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family -p clientbuffering=true
 ```
 
+### 5. Kerberos
```
busbey: nit: It might be clearer that this is only needed sometimes if the head is "Kerberos (optional)"

busbey:
this will be a great addition. thanks for working on it!

uijin:
Hi @busbey ,

Thanks for your comments. The modification has been committed.

busbey:
looks good to me. I'll grab a secure hbase cluster to test it tonight. If someone else beats me to hit, feel free to hit the merge PR button.

busbey:
Sorry that the final smoke test slipped my schedule. I'll get to this today.

busbey:
just noticed this is targeting the 0.2.0 RC branch. I've cherry picked it over to master. I'll test it there and then push.

busbey:
Testing against a kerberos-enabled cluster, it looks like the client works without this patch if I kinit ahead of time. Can we pare this down to the support for logging in via a provided keytab and the additional instructions?

Logging in via passed keytab isn't working for me, and there's nothing obvious in the logs. Could we add some logging, both to indicate that we're going to authenticate via kerberos (like what principal and the path to the keytab) and if we think it succeeded?

I can confirm that the keytab I'm using works, because I can kinit with it ahead of time and then rely on the extant kerberos support.

busbey:
Figured out why the keytab logins aren't working. Please make sure you also update the hbase-10 client. I'll need to find an hbase 0.94 cluster to test the older client.

uijin:
Hi @busbey,

I execute YCSB on HBase 0.98 with kerberos, which provided by HDP 2.2 distribution. Even I `kinit` ahead of time, YCSB is still failed on kerberos authentication, so that I patch the hbase client to make it works.

This patch is not added to hbase-10 client because I have such cluster for verification, and I am glad if someone could do the verificatino.

busbey:
A 0.98 cluster to test the hbase client is much easier for me to find, thanks for the heads up.

If you include the changes for the hbase-10 client I can test them on the cluster I'm already testing against.

busbey:
Okay, so against a 0.98 cluster I can use both the hbase and hbase-10 clients when the cluster is insecure.

Once the 0.98 cluster has kerberos, I can't run the client either before or after your changes. It appears to hang and then after a substantial amount of time shows a bunch of errors. Can you add some more logging so it's easier to figure out what's gone wrong?

uijin:
Today I install the latest HDP 2.2.6 distribution, running HBase 0.98.4.2.2 with kerberos-enabled.

In this environment, YCSB execute successfully if `kinit` ahead of time, or with kerberos parameter of my patch. If any one of these methods not applied, YCSB will fail.

busbey:
This got auto-closed because the 0.2.0-staging branch was deleted with the release. Sorry about that. Could you reopen with your changes rebased onto master?

PR #338, opened 2015-07-03
Execute correct number of operations

tlopatic:
As it is right now, the number of executed operations is rounded down to the nearest multiple of the thread count.

This is a little confusing, if you, say, insert 3,000,000 records into a database using 7 threads and the database then contains only 2,999,997 records after that.


busbey:
Thanks for the fix!

Issue #334, opened 2015-07-01
Retry logic in clients


cmccoy:
Right now the Cassandra Thrift clients (0.7, 0.8, and 1.0) and the Aerospike client (#324) implement client-side operation retries within YCSB code. I find that a little counterintuitive: I'd expect the operation failure rates reported by YCSB to be based on pretty simplistic client usage. Anyone have any thoughts?

busbey:
I can see the merit in both arguments, but having differences across stores is yet more for our users to grok while comparing. So I think this is something we need to set a guideline for and document (sort of like the issue of making sure clients flush to remote after a transaction).

If we take the position of no-retries, what do we do with a datastore binding that does the retry below us? I guess at that point it can be seen as a feature of using that datastore?

If we want retries in general, it seems better to pull that functionality up into the wrapper. then we can report on the error rate still.

cmccoy:
> If we take the position of no-retries, what do we do with a datastore binding that does the retry below us? I guess at that point it can be seen as a feature of using that datastore?

I think so. If the client itself implements retries, fine - you'd see the same behavior in the wild.

busbey:
yeah, I agree. We should start labeling these issues that have datastore binding guidelines. Eventually once we make a doc we can then do a sweep of bringing bindings in line. Not much use enforcing it until then.

tlopatic:
Agreed, valid point.

I added the retry logic to take care of the occasional Amazon EBS hiccup. Thinking about this again, though, that logic should actually go into the Aerospike client and not into the YCSB binding, so that everyone benefits from it.

I'll change that.

Issue #245, opened 2015-05-18
make a contribution guide


busbey:
Make a short guide for new contributors
- include pointers to get access to code, issue tracker, mailing list
- include style guide
- include pr guidance
- include licensing guidance
- include expectations around releases

busbey:
Are we review then commit, @cmccoy? I'd be in favor of adopting it as our policy.

For folks who aren't familiar, that would mean maintainers are on the same footing as other contributors: someone else has to agree that a change is ready for merge before things are pushed into the main repo.

Given our current low maintainer count, I'd be fine with accepting community reviews as authoritative (regardless they'd be lovely to have).

cmccoy:
On May 23, 2015 8:17 AM, "Sean Busbey" <notifications@github.com> wrote:
>
> Are we review then commit, @cmccoy? I'd be in favor of adopting it as our
policy.

Me too, especially for code changes.
I'm heading out if town for the weekend, but will catch up Monday or so.

> For folks who aren't familiar, that would mean maintainers are on the
same footing as other contributors: someone else has to agree that a change
is ready for merge before things are pushed into the main repo.
>
> Given our current low maintainer count, I'd be fine with accepting
community reviews as authoritative (regardless they'd be lovely to have).

busbey:
include expectations for adding new bindings / workloads

* _require_ a README that walks through running a load and a workload (for both bindings and workloads)
* javadocs
* example properties

busbey:
make sure to remind folks that we expect to work on Windows

kruthar:
Additional note as per @busbey comments on #500. 

>YCSB clients should default to the strictest consistency available and document consistency tuning and tradeoffs in their readme.

busbey:
other guidance that I don't see above that we've discussed:

* no client side buffering by default (but if you allow it please document how to use it and the tradeoffs)
* document which specific version(s) of the datastore you expect to work with
* document any set up of the datastore needed (like how some systems need to create a table with a particular schema)

risdenk:
As noted in #596:
> in general bindings should be in their own package

risdenk:
We should just make a CONTRIBUTING.md and have Github handle displaying it.

PR #351, opened 2015-07-13
MongoDBClient cleanup

jyemin:
I reviewed the changes to the MongoDBClient class that update it to use the 3.0 Java driver, and found a number of issues, all of which are addressed in the commits in this pull request.  Most are changes that simplify the MongoDBClient code in order to improve readability and reduce repetition.  The only functional change should be the fix to a bug in the way that journaling is enabled in the MongoClientURI.

### DIFF
--- a/mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java
+++ b/mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java
@@ -19,6 +19,7 @@ import java.util.Set;
 import java.util.Vector;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import com.mongodb.client.model.InsertManyOptions;
 import org.bson.Document;
 import org.bson.types.Binary;
 
@@ -53,14 +54,16 @@ import com.yahoo.ycsb.DBException;
  */
 public class MongoDbClient extends DB {
 
-    /** Update options to do an upsert. */
-    private static final UpdateOptions UPSERT = new UpdateOptions()
-            .upsert(true);
-
     /** Used to include a field in a response. */
     protected static final Integer INCLUDE = Integer.valueOf(1);
 
-    /** The database name to access. */
+    /** The options to use for inserting many documents */
+    protected static final InsertManyOptions INSERT_MANY_OPTIONS =
+    new InsertManyOptions().ordered(false);
+
+    /**
+     * The database name to access.
+     */
     private static String databaseName;
 
     /** The database name to access. */
@@ -85,7 +88,7 @@ public class MongoDbClient extends DB {
     private static int batchSize;
 
     /** The bulk inserts pending for the thread. */
-    private final List<InsertOneModel<Document>> bulkInserts = new ArrayList<InsertOneModel<Document>>();
+    private final List<Document> bulkInserts = new ArrayList<Document>();
 
     /**
      * Cleanup any state for this DB. Called once per DB instance; there is one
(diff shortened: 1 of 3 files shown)


jyemin:
I see that the Travis tests fail.  This is because I removed the special case for insert with batchSize of 1 to do an insertOne, instead of a replaceOne with upsert.  I believe this is proper (since batchSize > 1 uses insert rather than upsert, and the tests pass so long as you drop the collection prior to each run. 

Is there any way to configure these tests so that the collection is dropped between each run?

allanbank:
Jeff - Sorry for the delay. I'll start looking this over now.

mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java
```diff
@@ -53,14 +54,16 @@
...
-    private static final UpdateOptions UPSERT = new UpdateOptions()
-            .upsert(true);
-
     /** Used to include a field in a response. */
     protected static final Integer INCLUDE = Integer.valueOf(1);
 
-    /** The database name to access. */
+    /** The options to use for inserting many documents */
+    protected static final InsertManyOptions INSERT_MANY_OPTIONS =
+    new InsertManyOptions().ordered(false);
```
allanbank: Can this be renamed to show what it stands for instead of the type of object it is. Maybe INSERT_UNORDERED?

mongodb/src/main/java/com/yahoo/ycsb/db/OptionsSupport.java
```diff
@@ -76,7 +76,7 @@ else if ("acknowledged".equals(writeConcernType)) {
                 result = addUrlOption(result, "w", "1");
             }
             else if ("journaled".equals(writeConcernType)) {
-                result = addUrlOption(result, "journal", "true");
+                result = addUrlOption(result, "j", "true");
```
allanbank: The documentation for the standard URI says to use "journal":

  http://docs.mongodb.org/manual/reference/connection-string/#uri.journal

Does that not work with the Java Driver?

allanbank:
Jeff -

On the Insert vs. Upsert: That was a community contribution that I think was done for usability (e.g., reloads on the same workload without a drop of the collection just work). 

@busbey - Do you know if we have a standard for if the insert operation should be a pure insert or if it can be an upsert (if the DB supports it)?

Rob.

mongodb/src/main/java/com/yahoo/ycsb/db/OptionsSupport.java
jyemin: It never has, actually.  I reported it as https://jira.mongodb.org/browse/JAVA-1890, but for now let's switch to "j".

mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java
jyemin: Good idea.  Done.

jyemin:
On the Insert vs Upsert: This usability change only applies when batch size == 1, since it's using a bulk insert when batch size > 1, not a bulk upsert.

busbey:
We tend to give the binding implementors a lot of leeway. We've been discussing setting expectations over on #245 and #334. No one has mentioned choice of operations yet.

Personally, I'd prefer if we included some means to say "truncate the table" for each binding. That would make it easier to implement a quickstart for our recommended workload ordering.

jyemin:
I can go back to an upsert if that's the only way forward to get the tests passing again, but I'd prefer a better solution.

allanbank:
Jeff - 

Will the insert be faster than the upsert?

If so then lets leave the insert in place. I think the async and non-async test cases are colliding with each other since they use static ids. I think the easiest think to do is to add a truncate() method to the two clients and then call that to the setUp() for MongoDbClientTest and AsyncMongoDbClientTest. Thoughts?

 Rob

mongodb/src/main/java/com/yahoo/ycsb/db/OptionsSupport.java
allanbank: Oh.. Hummm. The problem is that the async driver only supports the "journal" parameter. Since "journal" is the documented name I think we should leave that. I wouldn't be opposed to adding both a "journal" and "j" parameter to the URI with a reference to remove it once the Jira ticket is released.

busbey:
Could they use different ids? The truncate will cause them to interfere with each other if we turn on parallel tests.

jyemin:
Updated.

I added journal=true back into the URL in addition to j=true, so that both drivers continue to work.

I also reverted to using upsert instead of insert when the batch size is 1 so that the tests are passing once again.

allanbank:
@busbey - Any problem with me merging this?

busbey:
Fine by me!

allanbank:
Rebased and Merged.

PR #354, opened 2015-07-15
change the port of CassandraClient10 from 9160 to get value form prop…

jixuan1989:
Because different clusters may monitor different ports. I think make the port as a parameter instead of a constant value is better. Just like what CassandraCQLClient does.


jixuan1989:
I feel so sorry that I make a mistake.

PR #355, opened 2015-07-15
set the port of Cassandra client variable (in CassandraClient10)

jixuan1989:
Because different clusters may monitor different ports. I think make the port as a parameter instead of a constant value is better. Just like what CassandraCQLClient does.


jixuan1989:
I think something is wrong. I will check the modification again.

PR #356, opened 2015-07-15
set the port of Cassandra client variable (in CassandraClient10)

jixuan1989:
Because different clusters may use different ports, get the value of the port from property file is better than a constant value. Just as what CassandraCQLClient does.


busbey:
Change looks good. One minor thing, could you squash the three commits into one and start the commit message with "[cassandra]" please?

jixuan1989:
OK, I will.

This a minor modification but useful for some cases. 

I also have writen several version for Cassandra 2.* version (by using Datastax Java Driver or Cassandra JDBC Driver) before I find YCSB project keeps be updated again.

YCSB is a good tool and I hope I can provide more contribution in future.

busbey:
happy to see more users.

If you prefer to make new PRs that's fine, but you can also just do a force push to the branch this PR tracks and it will update.

Issue #353, opened 2015-07-14
Exception while running redis


bigbes:
```
java -cp /tmp/ycsb-0.2.0-RC3/redis-binding/conf:/tmp/ycsb-0.2.0-RC3/conf:/tmp/ycsb-0.2.0-RC3/lib/core-0.2.0-RC3.jar:/tmp/ycsb-0.2.0-RC3/lib/jackson-core-asl-1.9.4.jar:/tmp/ycsb-0.2.0-RC3/lib/HdrHistogram-2.1.4.jar:/tmp/ycsb-0.2.0-RC3/lib/jackson-mapper-asl-1.9.4.jar:/tmp/ycsb-0.2.0-RC3/redis-binding/lib/commons-pool-1.5.5.jar:/tmp/ycsb-0.2.0-RC3/redis-binding/lib/redis-binding-0.2.0-RC3.jar:/tmp/ycsb-0.2.0-RC3/redis-binding/lib/jedis-2.0.0.jar com.yahoo.ycsb.Client -db com.yahoo.ycsb.db.RedisClient -P workloads/workloada -s -threads 64 -p measurementtype=timeseries -p timeseries.granularity=2000 -p operationcount=5000000 -p exporter=com.yahoo.ycsb.measurements.exporter.JSONArrayMeasurementsExporter -p recordcount=2000000 -p redis.host=sh4.tarantool.org -p redis.port=6400 -p exportfile=A-1.output -t
Loading workload...
Starting test.
2015-07-14 06:50:47:183 0 sec: 0 operations;
java.lang.NullPointerException
    at com.yahoo.ycsb.measurements.OneMeasurementTimeSeries.reportReturnCode(OneMeasurementTimeSeries.java:162)
    at com.yahoo.ycsb.measurements.Measurements.reportReturnCode(Measurements.java:240)
    at com.yahoo.ycsb.DBWrapper.update(DBWrapper.java:141)
    at com.yahoo.ycsb.workloads.CoreWorkload.doTransactionUpdate(CoreWorkload.java:754)
    at com.yahoo.ycsb.workloads.CoreWorkload.doTransaction(CoreWorkload.java:570)
    at com.yahoo.ycsb.ClientThread.run(Client.java:235)
```

It's hard to reproduce, so i can't create test case..

bigbes:
It's where it happens
``` java
        @Override
        public void reportReturnCode(int code) {
                Integer Icode=code;
                if (!returncodes.containsKey(Icode))
[...]

PR #357, opened 2015-07-15
Fix HashMap (gh-353) bug

bigbes:
and refactor *measurement* files

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
+++ b/core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
@@ -1,18 +1,18 @@
-/**                                                                                                                                                                                
- * Copyright (c) 2010 Yahoo! Inc. All rights reserved.                                                                                                                             
- *                                                                                                                                                                                 
- * Licensed under the Apache License, Version 2.0 (the "License"); you                                                                                                             
- * may not use this file except in compliance with the License. You                                                                                                                
- * may obtain a copy of the License at                                                                                                                                             
- *                                                                                                                                                                                 
- * http://www.apache.org/licenses/LICENSE-2.0                                                                                                                                      
- *                                                                                                                                                                                 
- * Unless required by applicable law or agreed to in writing, software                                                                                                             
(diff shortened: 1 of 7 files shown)


cmatser:
I noticed the following code in Measurements.java starting on line 133:
```java
  static class StartTimeHolder {
    long time;

    long startTime(){
      if(time == 0) {
        return System.nanoTime();
      }
      else {
        return time;
      }
    }
  }
```

While I don't doubt the code is working, I normally see methods like this set starting values in the initialize case.  I just want to point this out in case this was intended.

So, the change I would recommend would be:
```java
  static class StartTimeHolder {
    long time;

    long startTime(){
      if(time == 0) {
        time = System.nanoTime();
      }
      return time;
    }
  }
```

Chrisjan

bigbes:
@cmatser it wasn't intended, but i didn't modify this kind of constructions, before this project has strict format guidelines..

cmatser:
@bigbes Ah, sorry.  I saw it in the green section and thought it was new code.  I see now that it was just carried forward.  Thanks for clarifying.

busbey:
Looks good to me. What do you think @cmatser? Ready to merge?

cmatser:
@busbey Ready!

Do you know if there is a way for the web view diff to ignore whitespace?  So, when I click on the "Files changed" tab for this PR, I can reduce some of the noise caused by whitespace changes.

busbey:
I believe you can add "?w=1" to the URL

cmatser:
Yes!  That made my day!!!  :)

bigbes:
Any chance to merge this?

nitsanw:
@cmatser Your change suggestion to StartTimeHolder changes the behaviour of the method. In the original, if startTime is never set repeated calls to getStartTime will return __different__ values based on System.nanotime(). In you version the first call will set the startTime and it will stay at that value until someone resets it.
This is not to say that this is not confusing code which can do with some love, but the refactoring you propose needs further refining

cmatser:
@nitsanw  Agreed.  I wanted to point it out in case that was an unintentional side-effect.

Issue #316, opened 2015-06-22
StatusThread should wake up when the test has completed and exit.


allanbank:
Pre 0.2.0-rc2 the YCSB benchmark would exit as soon as all of the operations had completed. This allowed users to use the wall clock time as a coarse grain benchmark across runs. For instance I have several scripts to run YCSB across changes in the MongoDB driver. I use the wallclock time for each run to know if there are any regressions. Now all runs finish in 20 seconds regardless of the actual time to complete operations.

With 0.2.0-rc2 the StatusThread is keeping the ycsb client alive until the StatusThread does a poll after the all of the operations had completed. Thread dump below that show the only two remaining threads are the main which is in a Thread.join and the StatusThread which is in a sleepUnitl().

```
$ jstack 10828
2015-06-21 21:03:53
Full thread dump Java HotSpot(TM) 64-Bit Server VM (25.5-b02 mixed mode):

"Attach Listener" #30 daemon prio=9 os_prio=0 tid=0x00007f1300001000 nid=0x2aa3 waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"Thread-16" #26 prio=5 os_prio=0 tid=0x00007f1354163000 nid=0x2a61 runnable [0x00007f131adc9000]
   java.lang.Thread.State: RUNNABLE
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:338)
	at com.yahoo.ycsb.ClientThread.sleepUntil(Client.java:287)
	at com.yahoo.ycsb.StatusThread.run(Client.java:125)

"Service Thread" #9 daemon prio=9 os_prio=0 tid=0x00007f13540ca800 nid=0x2a5e runnable [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"C1 CompilerThread3" #8 daemon prio=9 os_prio=0 tid=0x00007f13540bd800 nid=0x2a5d waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"C2 CompilerThread2" #7 daemon prio=9 os_prio=0 tid=0x00007f13540bb000 nid=0x2a5c waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"C2 CompilerThread1" #6 daemon prio=9 os_prio=0 tid=0x00007f13540b9000 nid=0x2a5b waiting on condition [0x0000000000000000]
[...]

PR #286, opened 2015-06-07
[client] Add estimated time remaining to status messages.

busbey:
closes #239.

PR #359, opened 2015-07-16
Shutdown the status thread as soon as clients complete (Issue #316)

allanbank:
There are two commits. 

The first has the actual fix which uses a CountDownLatch to notify the StatusThread when all of the clients have completed. 

The second removes whitespace at the ends of lines and replaces tabs with 4 spaces.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Client.java
+++ b/core/src/main/java/com/yahoo/ycsb/Client.java
@@ -1,4 +1,4 @@
-/**                                                                                                                                                                                
+/**                                                                                                                                                                            
  * Copyright (c) 2010 Yahoo! Inc. All rights reserved.                                                                                                                             
  *                                                                                                                                                                                 
  * Licensed under the Apache License, Version 2.0 (the "License"); you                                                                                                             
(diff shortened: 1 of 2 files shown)


bigbes:
@allanbank as i suggest tabs must be 2 spaces: https://github.com/brianfrankcooper/YCSB/blob/master/checkstyle.xml#L161

core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -35,250 +37,302 @@
...
+    private boolean waitForClientsUntil(long deadline) {
+        boolean alldone=false;
+        long now=System.nanoTime();
+
+        while( !alldone && now < deadline ) {
+            try {
+                alldone = _completeLatch.await(deadline-now, TimeUnit.NANOSECONDS);
+            }
+            catch( InterruptedException ie) {
+                // Handled by while loop.
```
busbey: If we get interrupted, we ought to return false instead of continuing to wait. (Presuming that will cause us to exit soon)

Should also set the thread's interrupted status.

busbey:
Good catch @bigbes!

One other small not, please update the commit messages to start with what they impact, in this case "[client]".

allanbank:
I amended the commits.

allanbank:
I merged in pr-#286's patch since the two pr's conflict.

busbey:
can you amend the last to keep my authorship?

So long as it's the HEAD commit the command loolks like
```
 git commit --amend --author "Sean Busbey <busbey@apache.org>"
```

It'll open the commit message in your editor, just save and quit. You'll still show up as the committer when thing are in the repo. You'll have to force push the branch to update.

allanbank:
done

busbey:
:+1:

PR #360, opened 2015-07-16
[Cassandra] change the port of Cassandra (previous version 1.2) as a parameter instead of a constant value such as 9160

jixuan1989:
Because different clusters may use different ports, get the value of the port from property file is better than a constant value. Just as what CassandraCQLClient does.


busbey:
Thanks for the fix!

Issue #313, opened 2015-06-20
clean up file headers


busbey:
ensure all of our files have a license header

* For those that got added without a "(c) Yahoo" line, I don't think we can safely assume copyright assignment. Instead list it as "(c) The Contributors to YCSB" since that is at least accurate
* Make sure the copyright has a span of years that matches up with when contributions to that actual file happened. For example, Client.java should be (c) 2012-2015

PR #366, opened 2015-07-20
[mongodb] Update the pom.xml to make checkstyle fail the build. Fixes #365

### LINKED ISSUE #365: Make checkstyle enforcing for the MongoDB client.
I have been getting annoyed with the multiple formats rolling around so I have started with my own yard and made the checkstyle enforcing for the MongoDB client.

To get the build to pass most of the changes were done via automated formatters/cleanup. The rest where variable name changes.

The bad part of doing this kind of thing is it creates a patch horizon that it is extremely difficult to move patches across (in either direction). I think the consistency is worth the one time pain.

allanbank:
Update the pom.xml to make checkstyle fail the build. 
Updated code to satisfy the checkstyle validation.

Fixes #365


mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java
```diff
@@ -1,12 +1,19 @@
-/**
- * MongoDB client binding for YCSB.
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
```
busbey: We don't have any kind of CLA that assigns copyrights to Yahoo!, so let's leave out adding these for now. I've been trying to think of the best way to handle #313 and all I've got so far is that adding Yahoo copyright statements isn't the way to go. :)

If they're already present, please leave them. That's a whole different mess.

mongodb/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,26 @@
...
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+/**
+ * The YCSB driver for <a href="https://www.mongodb.org/">MongoDB</a>.
+ * Supports both the
+ * <a href="http://www.allanbank.com/mongodb-async-driver/">Asynchronous
+ * Java Driver</a> and the MongoDB Inc.
+ * <a href="http://docs.mongodb.org/ecosystem/drivers/java/">driver</a>.
```
busbey: nit: maybe a reference to check the README?

busbey:
with the copyright statement thing above addressed, I'm +1.

busbey:
(for other reviewiers, [the fill diff without whitespace changes](https://github.com/brianfrankcooper/YCSB/pull/366/files?w=1) helped speed things up.)

cmatser:
Should there be a "checkstyle.xml" file included in the PR?
 <configLocation>checkstyle.xml</configLocation>

allanbank:
@cmaster  - Interesting. Checkstyle was checking based on a config somewhere...  I'll be honest I and say I just copied the blob from the root pom and then flipped the enforcing flags on. There is the checkstyle.xml file in the root directory and I just assumed it was using that one.  I will see if setting the path to ```../checkstyle.xml``` makes a difference.

Maybe a better approach to this is to make the root pom enforcing and then update all of the poms except the MongoDB pom to override the configuration to not be enforcing. Then as we knock out sub-projects we just remove checkstyle item from the poms.

allanbank:
FYI - I did a force push of the amended commit to remove the bad copyright header.

I also changed the path to the checkstyle.xml to be relative and it had not effect. I left it that way in the commit as I think it is clearer that it is referencing the file one directory up.

cmatser:
Nice work!

Issue #222, opened 2015-03-30
Delete workload


vreniers:
Is there a way to execute a workload which benchmarks delete operations?
In the default workload files I can only see readproportions, updateproportions etc. but
no field for an amount of delete operations?

cmatser:
No, but you can add one fairly easily.  I did this, not because I wanted to measure deletes, but because I wanted to delete the records that I inserted during my insert runs so that I could start with the same data set for each run.  Each DB implementation is suppose to implement the "delete()" call.  So, I added a "deleteproportion" property in my local CoreWorkload.java to call the delete() method.  I also had to update my delete() method to make sure I deleted the specific records for my implementation.

I don't believe deletes are measured often because with these types of data stores, you typically never delete.... or you don't really care about the delete performance.  These stores are optimized for high scale reads.  However, everyone has there own needs.  So, I don't mean to suggest yours is invalid.  Good luck with your benchmarking.

Chrisjan

vreniers:
Okay, thanks for the info. I'll have to take a look at this approach.
I completely agree with you though that deletes aren't really of importance with most NoSQL data stores.
But it is something I'd thought I could benchmark.

busbey:
@vreniers if you're still interested in adding a delete workload let us know.

vreniers:
Hi, I'm not going to add the delete workload. 
I've already completed my benchmark without it.
And it's like you said, the delete performance isn't that important.

busbey:
sounds reasonable. thanks for getting back to me.

PR #368, opened 2015-07-23
Add a -delete command line option?

tlopatic:
Maybe this is useful to others, too? It's the complement to -load. It deletes the records that -load inserted.

Also, you can use it to remove the records inserted by workload D. It was a little tedious to fully reset and reload the database after each run of workload D.

It works completely analogously to -load and thus honors the insertstart and insertcount properties.

To remove the records inserted by workload D, just set insertstart to the recordcount and insertcount to something pretty high. If you work with 10,000,000 records, for example, you'd use

  ... -p insertstart=10000000 -p insertcount=1000000

to delete the records newly inserted by workload D. The -delete command doesn't abort on error, so it's safe to try to delete non-existent records. Hence we don't need to know the exact insertcount, just a safe upper bound.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -20,6 +20,11 @@ COMMANDS = {
         "description" : "Execute the load phase",
         "main"        : "com.yahoo.ycsb.Client",
     },
+    "delete" : {
+        "command"     : "-delete",
+        "description" : "Execute the clean-up phase",
+        "main"        : "com.yahoo.ycsb.Client",
+    },
     "run" : {
         "command"     : "-t",
         "description" : "Execute the transaction phase",
--- a/core/src/main/java/com/yahoo/ycsb/Client.java
+++ b/core/src/main/java/com/yahoo/ycsb/Client.java
@@ -236,6 +236,7 @@ class ClientThread extends Thread
   private static boolean _spinSleep;
   DB _db;
   boolean _dotransactions;
+  boolean _dodelete;
   Workload _workload;
   int _opcount;
   double _targetOpsPerMs;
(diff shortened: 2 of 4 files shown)


bin/ycsb
```diff
@@ -20,6 +20,11 @@ COMMANDS = {
         "description" : "Execute the load phase",
         "main"        : "com.yahoo.ycsb.Client",
     },
+    "delete" : {
+        "command"     : "-delete",
+        "description" : "Execute the clean-up phase",
```
allanbank: I would not call this the "clean-up" phase since there is actually a "CLEANUP" phase that shows up in the latency metrics. Maybe "Delete all data added during the load phase".

allanbank:
It is not obvious to me that records that are inserted as part of the workload are also deleted.  Have you tested that?

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -552,6 +552,19 @@ public boolean doInsert(DB db, Object threadstate)
...
 	/**
+	 * Do one delete operation. Works pretty much like doInsert(), but deletes records.
+	 * Accordingly, it uses the same parameters, "insertcount" and "insertstart", to
+	 * determine the record keys.
+	 */
+	public boolean doDelete(DB db, Object threadstate)
+	{
+		int keynum = keysequence.nextInt();
+		String dbkey = buildKeyName(keynum);
+		db.delete(table, dbkey);
```
allanbank: Need to track success failure like doInsert.

core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -343,7 +346,12 @@ public void run()
         while (((_opcount == 0) || (_opsdone < _opcount)) && !_workload.isStopRequested())
         {
 
-          if (!_workload.doInsert(_db,_workloadstate))
+          if (_dodelete)
+          {
+            _workload.doDelete(_db,_workloadstate);
```
allanbank: Should track the success failure of operations, somehow. Not sure breaking like the failure of doInsert() is the right thing here since we are likely trying to cleanup after a workload that may delete records.

allanbank:
I think what you are are after is really a drop command. Would that be a more descriptive name of the intended use?

Would it be better to add an optional ```drop(String table)``` method to the DB interface. Many data stores have an efficient mechanism to delete all of the data that will perform faster than deleting record by record. We can have the databases that don't implement the method then fall back to a record by record approach using ```scan()``` and ```delete()```.

@brianfrankcooper @busbey @cmatser @cmccoy @gkamat  - Thoughts?

busbey:
hurm. I remember another issue having a discussion about adding the equivalent of drop(trable). I'm of two minds. I agree that it would be super useful when automating going through multiple workloads and it's probably a better idea than row by row deletes (though in general a workload component of "delete" would be nice, see #222).

On the other hand, we don't do any of the other table maintenance work right now. We presume that will be handled out of band as a matter of simplifying the binding implementations.

On the gripping hand, it's _also_ inconvenient that we force all of that stuff out of band. If we're going to do table meta operations, I'd rather see an go at an API before we look at implementations. I don't think that general solution is what this request is after.

If we're going to add a "delete some" to what workloads can do, I'd rather see it done as "delete proportion" within the CoreWorkload that can run along side inserts and updates without impacting the Client code.

tlopatic:
Closing this. I agree that it doesn't make sense to have an isolated solution just for this very specific table maintenance use case.

PR #378, opened 2015-08-04
support download cloudera HBase

979969786:
Add cloudera repository to download cloudera HBase


busbey:
Normally you'd do this by adding the repo to your local maven settings.xml file. What's the advantage of doing it in our poms? AFAIK none of the vendor provided versions of HBase require e.g. a different module.

PR #381, opened 2015-08-05
[core] Fix #380 by working around recorder/logger initialization timing.

### LINKED ISSUE #380: [core] HdrHistogram logs are cannot be partially processed by HistogramLogProcessor
See https://github.com/HdrHistogram/HdrHistogram/issues/69
This issue with the HistogramLogReader/Writer interaction means that the histogram logs produced by YCSB cannot be processed correctly when the '-end' and '-start' options are used.

nitsanw:
The HdrHistogramReader makes some problematic assumptions on the roles played by start time, base time and histogram start/end times. This is to be revisited in a future release of HdrHistogram (see HdrHistogram/HdrHistogram#69). In the meantime this change will make YCSB record as expected by the current reader:
a. Set histogram log base time (same as start time)
b. Only create the recorder after the log has been setup

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementHdrHistogram.java
+++ b/core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementHdrHistogram.java
@@ -46,8 +46,7 @@ public class OneMeasurementHdrHistogram extends OneMeasurement {
   final PrintStream log;
   final HistogramLogWriter histogramLogWriter;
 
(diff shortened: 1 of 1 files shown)


busbey:
could you redo the commit message to start the first line with "[core]"?

busbey:
could you amend to leave out the formatting changes?

nitsanw:
Learned a new git trick in the process ;)
Thanks

busbey:
Thanks!

PR #388, opened 2015-08-14
Renamed GemFire driver to Geode

sbawaska:
Since GemFire is now incubating as Apache Geode, renaming the driver to Geode.

Also,
Removed an existing cache.xml which was meant to run on servers.
Added instructions on running the benchmark on Geode.


busbey:
is there any value in having the geode binding sit next to the gemfire one? say so that folks can compare the perf to older releases?

we don't include bindings that don't have a non-SNAPSHOT maven artifacts, because it prevents us from doing a release build. Please have the geode bindings disabled for now (you can look in the top level pom for an example in mapkeeper, nosqldb, and voldemort).

is there an ETA on a released version? I'd strongly prefer instructions that didn't rely on checking things out of source control (though I'm pretty sure there are other offenders in this category right now).

nit: it would be nice if the commits included a prefix of \[geode] (or \[gemfire] where changes were to that module) so that it's easier to scan the commit list for what's changing.

geode/pom.xml
```diff
@@ -45,8 +45,8 @@ LICENSE file.
   
   <repositories>
     <repository>
-      <id>gemstone</id>
-      <url>http://dist.gemstone.com.s3.amazonaws.com/maven/release</url>
+      <id>apache.snapshots</id>
+      <url>https://repository.apache.org/content/repositories/snapshots</url>
```
busbey: please no snapshot repositories. if building currently requires the SNAPSHOT artifact, won't the current instructions for building from source put the needed bits in the user's local maven repo?

busbey:
You folks have an eta on a released artifact?

What's the lilelyhood of having something in time for our 0.4.0 release (timeline on #409)

sbawaska:
Would you accept an alpha release? I think we may have an alpha build in
that timeframe.

busbey:
so long as the version number doesn't say SNAPSHOT and it's published in a release repo, it's fine by us.

busbey:
How's the Apache Geode release coming along?

sbawaska:
Vote on the first release is going on currently, should have a release out
soon.

sbawaska:
Merging upstream with my fork somehow generated a diff of 247 files. So, I am closing this pull request and will create a new one.

Issue #272, opened 2015-05-29
Handling incompatible API versions


cmccoy:
We should decide how to structure code handling incompatible API changes for different database versions. I've seen:
- Backwards-incompatible API changes (e.g., HBase 0.9x to 1.x, see #225).
- New APIs (e.g., Cassandra CQL, see #270).

Currently we're including multiple clients within a single module, but I think it would be cleaner to have a separate module for incompatible versions. The HBase client, for example, now needs to be compiled against HBase 1.0.x. I think having an `hbase-10` module and `hbase-09` module might be better.

busbey:
yes, definitely. in the case of HBase it should probably be hbase-94 and hbase-1.0 (and if the hbase-1.0 client doesn't work properly with a 0.98 cluster, we'll also need an hbase-98).

In addition to making the transition across versions clearer to downstream folks, this also makes it easier to deal with conflicting runtime needs.

cmccoy:
This is done for HBase.

risdenk:
Closing since this has been implemented with HBase and other bindings.

Issue #379, opened 2015-08-05
YCSB HBaseclient does not compile with HBase-2.0


ramkrish86:
Based on a discussion today raising this issue so that it can act as a placeholder. I can submit a patch too. 
The HBase-2.0 removes some of the deprcated API like KV.getValue(), kv.getRow() and the Result.raw() APIs.  We can modify the client code accordingly so that it compiles with HBase-2.0 and anyway it would work with HBase-1.0 also.
{code}
-                            Bytes.toString(kv.getQualifier()),
-                            new ByteArrayByteIterator(kv.getValue()));
+                            Bytes.toString(kv.getQualifierArray(), kv.getQualifierOffset(), kv.getQualifierLength()),
+                            new ByteArrayByteIterator(kv.getValueArray(), kv.getValueOffset(), kv.getValueLength()));
{code}

busbey:
sounds like we should break the 0.94 binding out into a different module so this can go into the hbase-10 binding.

cmccoy:
Updated #392 to remove the use of `KeyValue` in favor of Cell; it compiles without any deprecation warnings against HBase 1.1.1 as of 95e98d9604117685c2a41f09449d6d6e68228e3d.

cmccoy:
@ramkrish86: Could you try again with latest master?

ramkrish86:
Thanks for fixing this.  Will see if I can give a try on the latest checkin.

PR #392, opened 2015-08-19
[hbase] Separate HBase v0.9x and v1.x bindings

cmccoy:
See #272.
This change breaks `HBaseClient10` into a separate module, and downgrades the HBase version used by the `hbase` binding to v0.98.13.

Note that the HBase 1.0 binding is now accessed via `hbase1`, instead of `hbase-10` so that the runner script maps to the correct binding.

I tested against HBase v0.98.13 (with `hbase`) and v1.0.1.1 (with `hbase1`).  The naming is a bit confusing...does anyone have an opinion on this naming shceme versus an `hbase09` and `hbase1` binding?


busbey:
I'd like us to support HBase 0.94, and that won't work with the HBase 0.98 jars.

Do we want to have per-major-client-version bindings?

that would mean:

* HBase 0.94
* HBase 0.98
* HBase 1.y

That would mean we couldn't, e.g. take advantage of client-side optimizations in minor versions (atleast not without gymnastics). I certainly like this better than the alternative of per-minor-client-version bindings.

Unrelated, does keeping the 1.y API stuff in a module allow that module to build against the HBase 2.0.0-SNAPSHOT? Just wondering if we can close out #379 via this.

busbey:
wrt naming, what if we introduce a means for bindings to have submodules and a way to tell us which class corresponds to a version request?

Essentially, something that would let us do:
```
$ ycsb load hbase --binding-version 1.0
```
or
```
$ ycsb load hbase --binding-version 0.94
```

instead of having it tied to the binding name directly.

busbey:
the travis failure is because the distribution assembly pom needs to list the hbase1 binding as a dependency

cmccoy:
Fixed the travis failure, thanks.  I'll try to see if this addresses #379 tomorrow.

Re: the `--binding-version` flag...how would we handle new (and backwards compatible) versions? Would there be some mapping from DB version to binding version?

cmccoy:
OK, I separated the HBase bindings into three versions:

* HBase 1.x (`hbase10`)
* HBase 0.98.x (`hbase098`)
* HBase 0.94.x (`hbase094`)

The `hbase094` binding has no actual code, it just pulls in `hbase098` and replaces dependencies. I don't have a lot of experience with Maven; LMK if there's a tidier way of doing this.

I'd like to get this in and put off adding a `--binding-version` flag.

busbey:
I'm fine waiting on the binding-version flag.

@gkamat do you have time to review this?

@cmccoy are you wanting to get this into the release this weekend?

busbey:
as the patch currently stands, we'd have to make the next release version 0.4.0, since the "hbase" binding target goes away.

cmccoy:
I'm happy targeting this for the next release (leaving it out of 0.3.1).

hbase10/README.md
```diff
@@ -0,0 +1,74 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+# HBase (1.x) Driver for YCSB
+This driver is a binding for the YCSB facilities to operate against a HBase 1.x Server cluster.
+To run against an HBase 0.9x cluster, use the `hbase` binding.
```
busbey: I think with this patch there is no "hbase" binding, right?

hbase10/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Quickstart
+
+### 1. Start a HBase Server
+You need to start a single node or a cluster to point the client at. Please see [Apache HBase Reference Guide](http://hbase.apache.org/book.html) for more details and instructions.
+
+### 2. Set up YCSB
+You need to clone the repository and compile everything.
```
busbey: since we have a new file here, could we stick with downloading instructions?

pom.xml
```diff
@@ -69,7 +69,9 @@ LICENSE file.
   <properties>
     <maven.assembly.version>2.5.5</maven.assembly.version>
     <maven.dependency.version>2.10</maven.dependency.version>
-    <hbase.version>1.0.0</hbase.version>
+    <hbase094.version>0.94.27</hbase094.version>
+    <hbase098.version>0.98.13-hadoop2</hbase098.version>
+    <hbase10.version>1.0.1.1</hbase10.version>
```
busbey: 1.0.2 is out now, might as well update.

pom.xml
```diff
@@ -69,7 +69,9 @@ LICENSE file.
   <properties>
     <maven.assembly.version>2.5.5</maven.assembly.version>
     <maven.dependency.version>2.10</maven.dependency.version>
-    <hbase.version>1.0.0</hbase.version>
+    <hbase094.version>0.94.27</hbase094.version>
+    <hbase098.version>0.98.13-hadoop2</hbase098.version>
```
busbey: 0.98.14 is out now, might as well update.

hbase10/README.md
busbey: this doc error is the only fix-before-push thing I see.

cmccoy:
Sorry for the delay. All comments addressed - thanks!

busbey:
+1

PR #394, opened 2015-08-21
[docs] Fixes #393 - Update the curl command to use the --location option

### LINKED ISSUE #393: URL in the readme is a redirect
The curl command provided in the Getting Started section of the readme actually downloads a HTML redirect page instead of the .tar.gz file and it seems that the link in the redirect file cannot be used as is with curl/wget. I am not sure if it is possible to have a direct link to the file, but if someone knows how, that would be really useful. Thanks !

allanbank:


busbey:
I presume we're not concerned about github providing [malicious redirects](http://curl.haxx.se/docs/adv_20090303.html) and the use of https means we trust that the response comes from github.

+1

PR #403, opened 2015-09-02
YCSB-401 Introduce multiGet workload in HBaseclient like the PE tool in

ramkrish86:
Allows multiGet options to be specified while doing read only workloads specifically gets.

### DIFF
--- a/hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
+++ b/hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
@@ -46,9 +46,11 @@ import org.apache.hadoop.hbase.client.Table;
 import org.apache.hadoop.hbase.util.Bytes;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Random;
@@ -92,6 +94,9 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB
      */
     public boolean _clientSideBuffering = false;
     public long _writeBufferSize = 1024 * 1024 * 12;
+    public int multiGetsCount = 0;
+    // Collect the gets to form the multiGets
+    private List<Get> multiGets = null;
 
     public static final int Ok=0;
     public static final int ServerError=-1;
@@ -110,6 +115,10 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB
         }
         if (getProperties().containsKey("writebuffersize")) {
             _writeBufferSize = Long.parseLong(getProperties().getProperty("writebuffersize"));
+        }
+
+        if (getProperties().containsKey("multiGets")) {
+          multiGetsCount = Integer.parseInt(getProperties().getProperty("multiGets"));
         }
 
         if (getProperties().getProperty("durability") != null) {
@@ -149,6 +158,8 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB
 	  {
 	      throw new DBException(e);
 	  }
+    // Initialize the multiGets list
+    multiGets = new ArrayList<Get>(this.multiGetsCount);
     }
 
     /**
@@ -217,7 +228,7 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB
             }
         }
 
-        Result r = null;
+        Result[] r = new Result[1];
         try
         {
             if (_debug) {
@@ -232,7 +243,16 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB
                     g.addColumn(_columnFamilyBytes, Bytes.toBytes(field));
                 }
             }
-            r = _table.get(g);
+            if (multiGetsCount > 0) {
+              multiGets.add(g);
+              if (multiGetsCount == multiGets.size()) {
+                r = _table.get(multiGets);
+                // clear it every time after execution
+                this.multiGets.clear();
+              }
+            } else {
+              r[0] = _table.get(g);
+            }
         }
         catch (IOException e)
         {
@@ -246,17 +266,20 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB
             //do nothing for now...need to understand HBase concurrency model better
             return ServerError;
         }
-
-        if (r.isEmpty()) {
+        if (multiGetsCount == 0) {
+          if (r[0].isEmpty()) {
             return NoMatchingRecord;
+          }
         }
-        for (Cell c : r.listCells()) {
+        for (Result res : r) {
+          for (Cell c : res.listCells()) {
             result.put(Bytes.toString(CellUtil.cloneQualifier(c)),
-                    new ByteArrayByteIterator(CellUtil.cloneValue(c)));
+                new ByteArrayByteIterator(CellUtil.cloneValue(c)));
             if (_debug) {
-                System.out.println("Result for field: "+Bytes.toString(CellUtil.cloneQualifier(c))+
-                        " is: "+Bytes.toString(CellUtil.cloneValue(c)));
+              System.out.println("Result for field: " + Bytes.toString(CellUtil.cloneQualifier(c))
+                  + " is: " + Bytes.toString(CellUtil.cloneValue(c)));
             }
+          }
         }
         return Ok;
     }
@@ -323,10 +346,10 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB
 
                 HashMap<String,ByteIterator> rowResult = new HashMap<String, ByteIterator>();
 
-                for (KeyValue kv : rr.raw()) {
+                for (Cell kv : rr.rawCells()) {
                     rowResult.put(
-                            Bytes.toString(kv.getQualifier()),
-                            new ByteArrayByteIterator(kv.getValue()));
+                            Bytes.toString(kv.getQualifierArray(), kv.getQualifierOffset(), kv.getQualifierLength()),
+                            new ByteArrayByteIterator(kv.getValueArray(), kv.getValueOffset(), kv.getValueLength()));
                 }
                 //add rowResult to result vector
                 result.add(rowResult);


busbey:
Needs a README change explanung how and when to use the option.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -92,6 +94,9 @@
      */
     public boolean _clientSideBuffering = false;
     public long _writeBufferSize = 1024 * 1024 * 12;
+    public int multiGetsCount = 0;
```
busbey: Why keep this variable rather than check the size of the list?

allanbank: @busbey - I think that is the number of gets to put in each request. maybe a better name for it is multiGetBatchSize or maxGetBatchSize?

ramkrish86: Thanks for the reviews. I can change the name. 
Yes Sean, it is the number that indicates the count of the total number of gets that will be issued in one RPC call.

busbey:
nit: please update commit messages to start with "[hbase]" to indicate that you're only making changes to the hbase datastore binding.

README.md
```diff
@@ -47,7 +47,14 @@ Getting Started
     bin/ycsb run basic -P workloads/workloada
     ```
 
-  Running the `ycsb` command without any argument will print the usage. 
+   Running the `ycsb` command without any argument will print the usage. 
+   Inorder to use the multiGet feature where a set of 'gets' can be batched and sent to the server in one
```
busbey: please call this out as an optimization. Also note that it is effectively buffering the get calls client side, since that impact comparison with other systems.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -246,17 +266,20 @@ public int read(String table, String key, Set<String> fields, HashMap<String,Byt
...
         }
-
-        if (r.isEmpty()) {
+        if (multiGetsCount == 0) {
+          if (r[0].isEmpty()) {
             return NoMatchingRecord;
+          }
         }
-        for (Cell c : r.listCells()) {
+        for (Result res : r) {
```
busbey: isn't r[0] null here if we're using multiGet and the count is < multiGetsCount?

ramkrish86: If multiGetBatchSize > 0, only then we will use this whole multiGet based option. Else we will be going with the usual way of issuing a get and the result is stored in r[0]. We have this code anyway
          if (multiGetBatchSize == 0) {
          if (r[0].isEmpty()) {
            return NoMatchingRecord;
          }
        }
As you said if one of the get does return an empty result we need to have this
        for (Result res : r) {
          if (!res.isEmpty()) {

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -232,7 +243,16 @@ public int read(String table, String key, Set<String> fields, HashMap<String,Byt
                     g.addColumn(_columnFamilyBytes, Bytes.toBytes(field));
                 }
             }
-            r = _table.get(g);
+            if (multiGetsCount > 0) {
+              multiGets.add(g);
+              if (multiGetsCount == multiGets.size()) {
+                r = _table.get(multiGets);
```
busbey: what happens if one of the multigets fail? do we just fail all of them?

for those gets that get batched, we're just returning 'Ok' preemptively, right?

won't this batching up distort the latency distribution?

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -246,17 +266,20 @@ public int read(String table, String key, Set<String> fields, HashMap<String,Byt
...
-        if (r.isEmpty()) {
+        if (multiGetsCount == 0) {
+          if (r[0].isEmpty()) {
             return NoMatchingRecord;
+          }
         }
-        for (Cell c : r.listCells()) {
+        for (Result res : r) {
+          for (Cell c : res.listCells()) {
             result.put(Bytes.toString(CellUtil.cloneQualifier(c)),
```
busbey: in the case of multi-gets, won't this merge together the fields from multiple rows, effectively using last-result-wins?

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -232,7 +243,16 @@ public int read(String table, String key, Set<String> fields, HashMap<String,Byt
                     g.addColumn(_columnFamilyBytes, Bytes.toBytes(field));
                 }
             }
-            r = _table.get(g);
+            if (multiGetBatchSize > 0) {
+              multiGets.add(g);
+              if (multiGetBatchSize == multiGets.size()) {
+                r = _table.get(multiGets);
```
busbey: what happens if one of the multigets fail? do we just fail all of them?

for those gets that get batched, we're just returning 'Ok' preemptively, right?

won't this batching up distort the latency distribution?

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -246,17 +266,22 @@ public int read(String table, String key, Set<String> fields, HashMap<String,Byt
...
-        for (Cell c : r.listCells()) {
-            result.put(Bytes.toString(CellUtil.cloneQualifier(c)),
-                    new ByteArrayByteIterator(CellUtil.cloneValue(c)));
-            if (_debug) {
-                System.out.println("Result for field: "+Bytes.toString(CellUtil.cloneQualifier(c))+
-                        " is: "+Bytes.toString(CellUtil.cloneValue(c)));
+        for (Result res : r) {
+          if(!res.isEmpty()) {
+            for (Cell c : res.listCells()) {
+              result.put(Bytes.toString(CellUtil.cloneQualifier(c)),
```
busbey: in the case of multi-gets, won't this merge together the fields from multiple rows, effectively using last-result-wins?

and won't that mean the verification modes won't work?

README.md
```diff
@@ -47,7 +47,14 @@ Getting Started
     bin/ycsb run basic -P workloads/workloada
     ```
 
-  Running the `ycsb` command without any argument will print the usage. 
+   Running the `ycsb` command without any argument will print the usage. 
+   The 'multiGets' option is an optimization where a set of 'gets' can be batched (on the client side) and sent to the server in one
```
busbey: sorry, I just noticed this is in the top-level README. it should be in the hbase-specific README, since it doesn't appear to impact any of the other datastore bindings.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -246,17 +266,22 @@ public int read(String table, String key, Set<String> fields, HashMap<String,Byt
...
+          }
         }
-        for (Cell c : r.listCells()) {
-            result.put(Bytes.toString(CellUtil.cloneQualifier(c)),
-                    new ByteArrayByteIterator(CellUtil.cloneValue(c)));
-            if (_debug) {
-                System.out.println("Result for field: "+Bytes.toString(CellUtil.cloneQualifier(c))+
-                        " is: "+Bytes.toString(CellUtil.cloneValue(c)));
+        for (Result res : r) {
+          if(!res.isEmpty()) {
```
busbey: sorry, I'm still not seeing how this doesn't fail in the batch case.

take, for example, multiGetBatchSize <= 50, multiGets.size <= 0 (for the first time read is called)

1. we init r to a length 1 array of Result objects (each of those objects are null initially
2. the multiGets block activates (because multiGetBatchSize > 0)
3. we add the Get to the multiGets list
4. multiGets.size is now 1, which is less than 50, so we don't do the multiget call
5. multiGetBatchSize is still >0, so we skip the degenerate case check
6. we get to the for loop, which will iterator over the Result list (still size 1)
7. res will be the first reference, which is null
8. res.isEmpty explodes

I'd really like a test for this change, but our total lack of tests for hte HBase binding makes that a really steep request.

I haven't looked at #392 in some time; @cmccoy is it basically ready to go? I could review it quickly today and then submit a basic set of tests for the hbase-1.0+ binding.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
busbey: it looks like this also means that we have no way of getting NoMatchingRecord when using multigets mode?

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
ramkrish86: You mean with this change
{code}
if (multiGetBatchSize == multiGets.size()) {
{code}
I get it now. I thought the number of gets will be good enough to match it. I will update the code as I had it in my internal patch. I initialized the Result[] to null and for the non multiGet case I initialized the array locally.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
ramkrish86: The verification mode what  you say is right. I did not check that part as I was never using that verify mode.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
ramkrish86: Will check the latency distribution part also.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
ramkrish86: For the verification mode to work I think we need to modify the result HashMap<String,ByteIterator>() to a Vector of hashmap as we do in the case of scans. Is that acceptable?

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
ramkrish86: I can fix other comments. But the latency distribution part am not sure how to handle with multiGets.  So till the multiGet batch size is reached the latency distribution is going to be high since we don do the actual operation but once the batch size is reached then the multiGets latency gets included.

busbey:
We now have some minicluster based tests courtesy of @cmccoy.

@ramkrish86 could you add some tests that flex this proposed functionality. I think it'll give a better sense of the impact in practice.

hbase/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
ramkrish86: For the verification part to be successful we need to make the result HashMap to a Vector of hashmap. But doing that will have an impact on the signature of the read() API in the base class and it will have an impact on the other DB impl also.  So can we add an abstract class for HBase and make HBaseclient and HBaseClient10 to implement that new abstract class?

busbey: If read is defined as retrieveing a single record, why do we need map?

cmatser: This seems to be common across other bindings as well.  I attributed this to the need to pull the data out of the database specific object into the java object that your app will be using.  I think every app will want to do this rather than transport the DB's proprietary object.  In that respect, it makes sense to do it and log the time in a benchmark as long as we're consistent across the bindings for each DB.

PR #410, opened 2015-09-18
[hbase10] Add integration tests for HBase 1.0 client

cmccoy:
Uses a mini-cluster to test the HBase binding.


busbey:
can we add a src/test/resources/log4j.properties to squelch output that's below WARN?

cmccoy:
Good idea - done.

busbey:
<3 <3 <3 +1

PR #411, opened 2015-09-19
Use a PageFilter in HBase scans

cmccoy:
Other bindings limit the number of results retrieved from the server. The HBase bindings just close the scanner once they have received the desired number of records. Adding a [PageFilter](https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/filter/PageFilter.html) matches the behavior of other bindings, and should improve performance.

### DIFF
--- a/hbase098/README.md
+++ b/hbase098/README.md
@@ -71,3 +71,6 @@ Following options can be configurable using `-p`.
 
 * `columnfamily`: The HBase column family to target.
 * `debug` : If true, debugging logs are activated. The default is false.
+* `hbase.usepagefilter` : If true, HBase
+  [PageFilter](https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/filter/PageFilter.html)s
+  are used to limit the number of records consumed in a scan operation. The default is true.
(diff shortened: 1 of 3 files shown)


busbey:
Can we make a flag for this (default on) sonthat folks can see the perf impact?

busbey:
Note to release manager: this change should be flagged/called out because hbase evaluators shouldn't compare results across hbase configurations or versions with those from earlier ycsb releases if this feature is enabled.

cmccoy:
@busbey - yes. I'd prefer to avoid introducing another bare parameter (e.g., `pagefilter`), but that matches other HBase parameters used. What do you think about adding a prefix to all parameters (e.g., `hbase.clientbuffering`, `hbase.columnfamily`) in the next release? Too much churn/pain?

busbey:
That sounds fine. I think we can probably even offer some kind of deprecation to give folks a couple of versions to make the change. sounds like a different ticket though.

cmccoy:
Added a parameter: `hbase.usepagefilter`, defaulting to `true`.

busbey:
+1

PR #414, opened 2015-09-22
Bring Aerospike adapter in line with adapter policy

tlopatic:
As per @cmccoy's comment in #324: Database adapters shouldn't introduce any new functionality. They should just pass things back and forth between YCSB and the database client.

This change removes back-off/retry logic for cluster overload conditions from the Aerospike adapter.


busbey:
please update the commit message to start with "[aerospike]"

busbey:
otherwise, LGTM.

tlopatic:
Oh, yes, makes sense. Done.

PR #419, opened 2015-09-28
[cassandra-cql] Make CassandraCQLClient consistent.

cmccoy:
* Changes the default port to 9042 (this is the default CQL port, the current
  default is the Thrift API port).
* Changes property `host` from a space-separated list to property
  `hosts`, a comma-separated list (to match the Thrift clients).
* Updates a few messages.

PR #420, opened 2015-09-28
change the default port and host property parsing for cassandra CQL binding

Thomas-Yang:
Close the issue #419. The following changes are made to Cassandra CQL binding:
1.  change the default port from 9160 to 9042
2.  unify the host property name ("host" --> "hosts")
3.  separating multiple hosts with comma


cmccoy:
I merged #419, which has these changes. Thanks!

PR #423, opened 2015-10-02
Add warmup phase before starting the measurement

### LINKED ISSUE #422: Add warmup stage before starting the measurements
I think it would be more rigorous to add a configurable warmup stage before starting the actual latency/throughput measurements, which
*  avoids the cold start effect at the backend
*  allows to customize how long (or what kinds of) the warmup should run

I found this feature was partially implemented in [(source)](https://github.com/jbellis/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/Client.java#L201), which could be a good starting point. 
Any thoughts?

Thomas-Yang:
Allow to warmup the backend datastores before collecting the latency/throughput measurements. The length of the warmup phase is configurable by the number of operations and expected execution time. Reference PR #121 and close #422.


core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -835,6 +916,48 @@ public void run()
...
+        warmupThreads.add(t);
+      }
+      for (Thread t : warmupThreads)
+      {
+        t.start();
+      }
+      try {
+        warmupLatch.await();
+      } catch (InterruptedException e) {
+        e.printStackTrace();
```
cmccoy: Is swallowing this exception intentional?

core/src/main/java/com/yahoo/ycsb/DB.java
```diff
@@ -74,6 +74,14 @@ public void init() throws DBException
 	}
 
 	/**
+	 * Initialize the DB for the warmup phase.
+	 * The operations should not be accounted into the measurements.
+	 */
+	public void init(boolean warmup) throws DBException
```
cmccoy: Maybe make this abstract? or move the implementation from DBWrapper to DB?

core/src/main/java/com/yahoo/ycsb/Client.java
Thomas-Yang: @cmccoy We need to handle the exception when the waiting thread is interrupted. I followed the same way as we catch the exception from the join operation.

core/src/main/java/com/yahoo/ycsb/DB.java
Thomas-Yang: @cmccoy I will make it abstract.

core/src/main/java/com/yahoo/ycsb/Client.java
busbey: that sounds like the join operation is wrong. at a minimum when catching an interruption exception we need to set the thread's interrupted status.

additionally, if we've been interrupted we're supposed to "clean up and get out." that probably means the remaining work we do after is a bad idea.

Thomas-Yang: @busbey restore the interrupted status and exit.

cmccoy:
Looks fine to me. @busbey: any other comments?

busbey:
Sorry this has sat idle so long. Would you mind rebasing to the current master and squashing to a single commit?

Also, please update the commit message to start with the impacted module(s), i.e. "[core]"

busbey:
Sorry, one thing that wasn't clear to me. Is this work derived from the changes in #121, or implementing the same desired functionality from scratch?

danielpoltx:
This a great feature. Would be nice to get it in

nitsanw:
This is easily doable by logging the hdr histogram logs and creating a summary skipping the first X seconds.

Issue #409, opened 2015-09-17
Release version 0.4.0


allanbank:
Currently for the 0.4.0 release the following items have already been merged:
* #392 - Separate HBase v0.9x and v1.x bindings.
* #371 - Make hdrhistogram default measurement type.

Of particular note is that #371 introduces two *incompatibilities* with previous releases:
* Default measurement changed from histogram to hdrhistogram
  Users who want previous behavior can set the 'measurementtype' property to 'histogram'
* Reported 95th and 99th percentile latencies now in microseconds (previously in milliseconds)

After the RC0 release is created we need to verify that both of those changes are working as intended. Once both items are tested I will create a RC1. 

Once the RC1 release has been created I will ask for everyone to test as many of the binding as possible. I think the minimum set should be (as they were tested in 0.3.0 or saw a recent update):
* Apache Accumulo 1.6.0 via the accumulo binding.
* Apache Cassandra 2.0.15 via the cassandra-cql and cassandra-10 bindings.
* Apache HBase 1.0.0 via the hbase10 binding.
* Apache HBase 0.98.x via the hbase098 binding.
* Apache HBase 0.94.x via the hbase094 binding.
* MongoDB via the mongodb and mongodb-async bindings. 
* Tarantool 1.6.1 via the tarantool binding.
* Aerospike 3.5.15 via the aerospike binding.

While I will not pester anyone about testing the bindings until after the RC1 I want to encourage everyone to test as soon as RC0 is released.

In the interest of getting the release out quickly I am going to try and avoid patches unless absolutely necessary. The flip side of that is I want to give folks a chance to get more changes into the baseline so I won't do the initial RC0 build until September 21st. I will plan to do the RC1 build about a week later on September 28th and hope for a final release before October 8th. To summarize:
* RC0 - September 21st
* RC1 - September 28th
* 0.4.0 - before October 8th.

busbey:
Sounds like a good plan.

As things currently stand, #392 also introduces an incompatible change because it replaces the behavior of the 'hbase' binding with 'hbase098'.

busbey:
Just realized #392 also broke compat by renaming 'hbase-10' to 'hbase10'.

As an aside, back in 0.3.0 only hbase against version 1.0.0 was tested. AFAIK after we determined 0.98 and 0.94 didn't work back in 0.2.0 we didn't retest due to time.

cmccoy:
@allanbank - thanks for doing this!

[...]

PR #427, opened 2015-10-08
[core] Update the AcknowledgedCounterGenerator to use a power of two and mask.

allanbank:
Update the AcknowledgedCounterGenerator to use a power of two size for the window so a mask can be used instead of a modulo operation. This avoid issues when there is overflow of the id integer values to negative values.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/generator/AcknowledgedCounterGenerator.java
+++ b/core/src/main/java/com/yahoo/ycsb/generator/AcknowledgedCounterGenerator.java
@@ -1,3 +1,19 @@
+/**
+ * Copyright (c) 2015 YCSB contributors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
 package com.yahoo.ycsb.generator;
 
 import java.util.concurrent.locks.ReentrantLock;
@@ -8,7 +24,11 @@ import java.util.concurrent.locks.ReentrantLock;
  */
 public class AcknowledgedCounterGenerator extends CounterGenerator
 {
-	private static final int WINDOW_SIZE = 1000000;
+	/** The size of the window of pending id ack's. 2^20 = {@value} */
+	static final int WINDOW_SIZE = Integer.rotateLeft(1, 20);
+
+	/** The mask to use to turn an id into a slot in {@link #window}. */
+	private static final int WINDOW_MASK = WINDOW_SIZE - 1;
 
 	private final ReentrantLock lock;
 	private final boolean[] window;
@@ -40,22 +60,22 @@ public class AcknowledgedCounterGenerator extends CounterGenerator
 	 */
 	public void acknowledge(int value)
 	{
-		if (value > limit + WINDOW_SIZE) {
+		final int currentSlot = (value & WINDOW_MASK);
+		if (window[currentSlot] == true) {
 			throw new RuntimeException("Too many unacknowledged insertion keys.");
 		}
 
-		window[value % WINDOW_SIZE] = true;
+		window[currentSlot] = true;
 
 		if (lock.tryLock()) {
 			// move a contiguous sequence from the window
 			// over to the "limit" variable
-
 			try {
+			  // Only loop through the entire window at most once.
+			  int beforeFirstSlot = (limit & WINDOW_MASK);
 				int index;
-
-				for (index = limit + 1; index <= value; ++index) {
-					int slot = index % WINDOW_SIZE;
-
+				for (index = limit + 1; index != beforeFirstSlot; ++index) {
+					int slot = (index & WINDOW_MASK);
 					if (!window[slot]) {
 						break;
 					}
--- a/core/src/test/java/com/yahoo/ycsb/generator/AcknowledgedCounterGeneratorTest.java
+++ b/core/src/test/java/com/yahoo/ycsb/generator/AcknowledgedCounterGeneratorTest.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2015 YCSB contributors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+package com.yahoo.ycsb.generator;
+
+import java.util.Random;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+
+import org.testng.annotations.Test;
+
+/**
+ * Tests for the AcknowledgedCounterGenerator class.
+ */
+public class AcknowledgedCounterGeneratorTest {
+
+  /**
+   * Test that advancing past {@link Integer#MAX_VALUE} works.
+   */
+  @Test
+  public void testIncrementPastIntegerMaxValue() {
+    final long toTry = AcknowledgedCounterGenerator.WINDOW_SIZE * 3;
+
+    AcknowledgedCounterGenerator generator =
+        new AcknowledgedCounterGenerator(Integer.MAX_VALUE - 1000);
+
+    Random rand = new Random(System.currentTimeMillis());
+    BlockingQueue<Integer> pending = new ArrayBlockingQueue<Integer>(1000);
+    for (long i = 0; i < toTry; ++i) {
+      int value = generator.nextInt();
+
+      while (!pending.offer(value)) {
+
+        Integer first = pending.poll();
+
+        // Don't always advance by one.
+        if (rand.nextBoolean()) {
+          generator.acknowledge(first);
+        } else {
+          Integer second = pending.poll();
+          pending.add(first);
+          generator.acknowledge(second);
+        }
+      }
+    }
+
+  }
+}


core/src/main/java/com/yahoo/ycsb/generator/AcknowledgedCounterGenerator.java
```diff
@@ -40,11 +44,12 @@ public int lastInt()
 	 */
 	public void acknowledge(int value)
 	{
-		if (value > limit + WINDOW_SIZE) {
+                final int currentSlot = (value & WINDOW_MASK);
```
busbey: nit: indent is off, I think?

busbey:
looks reasonable. I'll retry on my test rig.

core/src/main/java/com/yahoo/ycsb/generator/AcknowledgedCounterGenerator.java
allanbank: fixed. darn tabs.

tlopatic:
I am not sure that the "<=" loop still works.

Imagine the following window situation. The limit is at 2^31 - 3, so the beginning of the window is at 2^31 - 2. I.e., the first two spots in the window belong to the two highest positive keys 2^31 - 2 and 2^31-1. After those two spots, we wrap around, i.e., the remaining spots are negative keys. All spots in the window are false.

Now consider the following interleaving:

  * Acknowledge 2^31 - 1. No window shift, as 2^31 - 2 is still outstanding.
  * Acknowledge 2^31 - 2. Limit now becomes 2^31 - 2. Window now stands at 2^31 - 1.

From now on, all acknowledgements will be negative. This means that the loop will not run for them, because limit + 1 == 2^31 - 1 (i.e., it's positive) and value < 0. The window is stuck.

Or am I missing something?

Admittedly, this specific scenario will probably never happen, but still. I think that the easiest and lowest risk fix would be:

  * On startup, test recordcount and print an error, if it is above 2^31-1.

  * Throw a "number of records exceeds 2^31-1" exception in doTransactionInsert(), if the generated keynum is negative.

The exception would have the advantage of alerting people to the fact that their measurements have been inaccurate in the past.

allanbank:
@tlopatic -

Good catch. I wrote a test and what you described is exactly what happened. I changed the code to not stop until it found an unclaimed slot instead of stopping at its value. That fixed the test.

This look good to everyone?

core/src/test/java/com/yahoo/ycsb/generator/AcknowledgedCounterGeneratorTest.java
```diff
@@ -0,0 +1,60 @@
+/**
+ * Copyright (c) 2012 YCSB contributors. All rights reserved.
```
busbey: should be 2015 since this file is new.

core/src/test/java/com/yahoo/ycsb/generator/AcknowledgedCounterGeneratorTest.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Random;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+
+import junit.framework.TestCase;
+
+/**
+ * Tests for the AcknowledgedCounterGenerator class.
+ */
+public class AcknowledgedCounterGeneratorTest extends TestCase {
```
busbey: please use the Test annotation version of running junit instead of TestCase.

core/src/main/java/com/yahoo/ycsb/generator/AcknowledgedCounterGenerator.java
```diff
@@ -1,3 +1,19 @@
+/**
+ * Copyright (c) 2012 YCSB contributors. All rights reserved.
```
busbey: also 2015

core/src/test/java/com/yahoo/ycsb/generator/AcknowledgedCounterGeneratorTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Test that advancing past {@link Integer#MAX_VALUE} works.
+   */
+  public void testIncrementPastIntegerMaxValue() {
+    final long toTry = AcknowledgedCounterGenerator.WINDOW_SIZE * 3;
+
```
busbey: there's extra whitespace on this line

tlopatic:
@allanbank Good idea. Then we don't have to deal with signs at all.

Now, I think, the only remaining edge case is that the loop doesn't terminate, if "window[currentSlot] = true" sets the last remaining "false" slot in the window to "true." Then all slots in the window are "true."

Maybe move the "window[currentSlot] = true" after the  "tryLock" block? In that way, there'll always be at least one false slot. Unless I am missing something, the only disadvantage is, that "acknowledge()" won't ever move the current "value" over to "limit". But that would then simply be done in one of the subsequent invocations of "acknowledge()."

busbey:
Using this patch on the 0.4.0-staging branch, I ran through two tests of load + run workload d and workload e, first for 5 minutes and then for an hour.

workload d worked fine. workload e appears to hang forever on the run phase. in each case I get:

```
Starting test.
Maximum execution time specified as: 3600 secs
Maximum time elapsed. Requesting stop for the workload.
Stop requested for workload. Now Joining!
```

and then on stdout
```
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
Still waiting for thread Thread-1 to complete. Workload status: true
```

it did this for 2 hours, so I'm pretty sure it'll go one forever. I grabbed some jstacks:

[17220.7.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10765/17220.7.jstack.txt)
[17220.6.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10767/17220.6.jstack.txt)
[17220.5.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10768/17220.5.jstack.txt)
[17220.4.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10769/17220.4.jstack.txt)
[17220.3.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10766/17220.3.jstack.txt)
[17220.2.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10770/17220.2.jstack.txt)
[17220.1.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10774/17220.1.jstack.txt)
[17220.0.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10771/17220.0.jstack.txt)
[6379.7.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10772/6379.7.jstack.txt)
[6379.6.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10775/6379.6.jstack.txt)
[6379.5.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10773/6379.5.jstack.txt)
[6379.4.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10776/6379.4.jstack.txt)
[6379.3.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10777/6379.3.jstack.txt)
[6379.2.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10778/6379.2.jstack.txt)
[6379.1.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10779/6379.1.jstack.txt)
[6379.0.jstack.txt](https://github.com/brianfrankcooper/YCSB/files/10780/6379.0.jstack.txt)

 and it looks like we're stuck waiting for a key?

tlopatic:
Oh, wow. It seems that nextKeynum() gets into an endless loop. My guess would be that transactioninsertkeysequence.lastInt() has wrapped around and is negative, whereas keychooser.nextInt() keeps producing non-negative numbers.

By default, the Zipfian generator is initialized to produce values up to:

recordcount + (operationcount * insertproportion * 2)

My guess off the top of my head would be that this doesn't overflow and keeps keychooser.nextInt() non-negative, while transactioninsertkeysequence.nextInt() ultimately overflows.

If this hypothesis is correct, though, this wouldn't have anything directly to do with the changes in AcknowledgedCounterGenerator(). It wouldn't explain why it is now that this problem shows up.

@busbey Could you post the current version of the script that you're using to run YCSB? The script posted in #409 - https://gist.github.com/busbey/fcb6f63128f3a1a85541 - seems to do "load" and "load" for workload E instead of "load" and "run". So I am not sure that this script is still representative of how exactly you're running workload E. I'm mostly interested in filling in the variables above: recordcount and operationcount.

Also, how many TPS do you get on workload E? It would be interesting to estimate how long a benchmark run it takes until insertions make the keys wrap around in your setup.

busbey:
I updated [the gist with a redacted version of the script as it was run last night](https://gist.github.com/busbey/fcb6f63128f3a1a85541/8b408d3fd9a6f60888c4b6650af78b2962c70821).

Hurm. it looks like I never get a measurement export from workload e due to the hang.

I'll launch another workloade run with ```-s``` so that there's some output as it goes.

busbey:
looks like it happens right away:

```
Loading workload...
Starting test.
2015-10-09 11:08:09:396 0 sec: 0 operations; est completion in 0 seconds 
Maximum execution time specified as: 3600 secs
2015-10-09 11:08:19:383 10 sec: 94 operations; 9.4 current ops/sec; est completion in 528 days 19 hours [SCAN: Count=88, Max=604159, Min=15448, Avg=229874.23, 90=477439, 99=586239, 99.9=604159, 99.99=6041
59] [INSERT: Count=6, Max=9391, Min=750, Avg=3158.83, 90=5567, 99=9391, 99.9=9391, 99.99=9391] 
2015-10-09 11:08:29:382 20 sec: 94 operations; 0 current ops/sec; est completion in 1057 days 15 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0,
 Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:08:39:382 30 sec: 94 operations; 0 current ops/sec; est completion in 1586 days 11 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0,
 Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:08:49:382 40 sec: 94 operations; 0 current ops/sec; est completion in 2115 days 7 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0, 
Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:08:59:382 50 sec: 94 operations; 0 current ops/sec; est completion in 2644 days 3 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0, 
Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:09:09:382 60 sec: 94 operations; 0 current ops/sec; est completion in 3172 days 23 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0,
 Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
...SNIP...
2015-10-09 11:20:39:382 750 sec: 94 operations; 0 current ops/sec; est completion in 39662 days 10 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:20:49:382 760 sec: 94 operations; 0 current ops/sec; est completion in 40191 days 6 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:20:59:382 770 sec: 94 operations; 0 current ops/sec; est completion in 40720 days 2 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:21:09:382 780 sec: 94 operations; 0 current ops/sec; est completion in 41248 days 22 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
2015-10-09 11:21:19:382 790 sec: 94 operations; 0 current ops/sec; est completion in 41777 days 18 hours [SCAN: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] [INSERT: Count=0, Max=0, Min=9223372036854775807, Avg=�, 90=0, 99=0, 99.9=0, 99.99=0] 
```

tlopatic:
Thanks for the additional details, @busbey!

Ah, I didn't realize that a recordcount of 0 makes YCSB set recordcount to 2^31 - 1. So, the wrap-around happens essentially instantaneously in your configuration, after the very first insertion.

So, here's what I think is going on. Replacing the variables from above with the values from the script, the ScrambledZipfianGenerator is instantiated with the following value:

(2^31 - 1) + (429,496,729 * 0.05 * 2) = -2,104,533,977

So the ScrambledZipfianGenerator now gives us these values:

(something non-negative) % -2,104,533,977

I.e., we get values in [-2,104,533,976, 0] as keys at which to start a scan. Now let's look at the insertions. There have been 6 insertions. So we're now at 2^31 - 1 + 6, which is -2,147,483,643.

We loop in nextKeynum(), until the ScrambledZipfianGenerator gives us a key that's less than or equal to the last inserted key. However, the lowest key the ScrambledZipfianGenerator produces is -2,104,533,996 and the last inserted key is -2,147,483,643. So each thread that tries to scan and enters nextKeynum() starts looping forever.

The likely reason why workload D (which also does inserts and which uses the same parameters as workload E) works is that it doesn't use the Zipfian distribution. Its SkewedLatestGenerator seems to work OK in the presence of wrap-arounds. Well, it least it doesn't cause endless loops. I can't tell right now off the top of my head whether we don't mess up the distribution.

So, all makes perfect sense. However: I am still not sure why this error is showing up now and didn't show up before the AcknowledgedCounterGenerator fix.

@allanbank Am I missing something obvious? I am a little confused.

core/src/test/java/com/yahoo/ycsb/generator/AcknowledgedCounterGeneratorTest.java
allanbank: Switched it but it is the TestNG annotation not the JUnit annotation. The JUnit version is 3.8.1. Something else we should consider cleaning up.

allanbank:
I just pushed the few minor changes pointed out by @busbey. Thanks for that. Live by the copy... die by the paste.

@tlopatic - I don't think you are missing anything. I have spent the last several hours convincing myself that the workloade behavior is not a regression. The bug has been there since at least 0.2.0. I ran the same command that deadlock on 0.3.0 and 0.2.0 and got the same behaviour. 

@busbey - Would you mind re-running the workloade run with 0.3.1 and reporting back?

Rob.

busbey:
Sure, I can do that today.

busbey:
I reproduced the workload e failure in 0.3.0, so I think we're fine to go with this solution, file a follow on for workload e, and list that follow on as a known issue for the release.

busbey:
(also verified in 0.3.1, btw)

PR #428, opened 2015-10-08
[mongodb] Add mongodb.upsert configuration option.

### LINKED ISSUE #400: YCSB MongoDB execute update instead of insert
i tried to use YCSB-0.3.0 to test the write ops of mongodb, and i used mongostat to show the status of mongodb, i can't find the insert operation not matter during load phase or run phase with insert operations, it turns out YCSB use update operation to insert all the records. So, how can i use YCSB to test insert ops of mongodb?

allanbank:
Added a configuration option (mongodb.upsert) to control if the driver
performs inserts or upserts operations for the DB.insert() method.
Defaults to doing inserts.

Note: Change in behavior from previous versions that performed upserts
for non-batched DB.insert() operations.

Fixes #400.

### DIFF
--- a/mongodb/README.md
+++ b/mongodb/README.md
@@ -108,8 +108,14 @@ See the next section for the list of configuration parameters for MongoDB.
   - Useful for the insert workload as it will submit the inserts in batches inproving throughput.
   - Default value is `1`.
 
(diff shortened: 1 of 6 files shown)


busbey:
+1

PR #432, opened 2015-10-11
[core] Scale YCSB to 2^63 operations and/or records.

### LINKED ISSUE #384: avoid integer overflow when setting up zipf distribution
I have workloade running in an infinite loop because I passed in recordcount=0 and operationcount > 0.

recordcount 0 is treated as a special case meaning "MAX_INT"

```
recordcount=Integer.parseInt(p.getProperty(Client.RECORD_COUNT_PROPERTY, Client.DEFAULT_RECORD_COUNT));
if(recordcount == 0)
  recordcount = Integer.MAX_VALUE;
```

and later workload e calculates the generator

```
else if (requestdistrib.compareTo("zipfian")==0)
{
  //it does this by generating a random "next key" in part by taking the modulus over the number of keys
  //if the number of keys changes, this would shift the modulus, and we don't want that to change which keys are popular
  //so we'll actually construct the scrambled zipfian generator with a keyspace that is larger than exists at the beginning
  //of the test. that is, we'll predict the number of inserts, and tell the scrambled zipfian generator the number of existing keys
  //plus the number of predicted keys as the total keyspace. then, if the generator picks a key that hasn't been inserted yet, will
  //just ignore it and pick another key. this way, the size of the keyspace doesn't change from the perspective of the scrambled zipfian generator                      
  int opcount=Integer.parseInt(p.getProperty(Client.OPERATION_COUNT_PROPERTY));
  int expectednewkeys=(int)(((double)opcount)*insertproportion*2.0); //2 is fudge factor
                        
  keychooser=new ScrambledZipfianGenerator(recordcount+expectednewkeys);
}

```

ScambledZG takes a long for number of items, but since recordcount and expectednewkeys are both integers, the result is overflow and then poor behavior while trying to get the next keynum.

workload e is the only one of the standard workloads to have a zipf distribution and insert operations in run mode. In all of my load statements I don't use recordcount=0 because I'm breaking up the inserts over multiple clients.

allanbank:
Updates the generators and measurements classes to use longs for ids and
counts. Also enforced where appropriate that the generated values will
not become negative. Instead the value will stop at Long.MAX_VALUE
avoiding issues for distributions that are looking for values in a
defined range.

Fixes #384


core/src/main/java/com/yahoo/ycsb/generator/LongGenerator.java
```diff
@@ -18,35 +18,35 @@
...
- *
  */
-public abstract class IntegerGenerator extends Generator 
+public abstract class LongGenerator extends Generator 
 {
-	int lastint;
+	long last;
 	
 	/**
 	 * Set the last value generated. IntegerGenerator subclasses must use this call
```
cmccoy: Nit: this comment is out of date (`LongGenerator`, `lastLong`)

core/src/main/java/com/yahoo/ycsb/generator/ScrambledZipfianGenerator.java
```diff
@@ -93,23 +79,16 @@ public ScrambledZipfianGenerator(long min, long max, double _zipfianconstant)
...
-	 */
-	@Override
-	public int nextInt() {
-		return (int)nextLong();
-	}
 
 	/**
 	 * Return the next long in the sequence.
 	 */
+  @Override
```
cmccoy: Nit: indentation doesn't match here.

core/src/main/java/com/yahoo/ycsb/generator/UniformGenerator.java
```diff
@@ -46,7 +46,8 @@ public UniformGenerator(Vector<String> values)
 	 */
 	public String nextString()
 	{
-		_laststring=_values.elementAt(_gen.nextInt());
+	  int offset = (int) Math.abs(_gen.nextLong() % _values.size());
```
cmccoy: Nit: indentation here

core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementHistogram.java
```diff
@@ -68,15 +68,16 @@ public OneMeasurementHistogram(String name, Properties props)
   /* (non-Javadoc)
    * @see com.yahoo.ycsb.OneMeasurement#measure(int)
```
cmccoy: This comment is out of date.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -521,7 +528,7 @@ public String buildKeyName(long keynum) {
    * Build a deterministic value given the key information.
    */
   private String buildDeterministicValue(String key, String fieldkey) {
-    int size = fieldlengthgenerator.nextInt();
+    int size = (int) fieldlengthgenerator.nextLong();
```
cmccoy: Worth checking for overflow here?

allanbank:
@cmcoy (and anyone else) keep the nits coming.

I should have also mentioned earlier that I am really looking for feedback on the changes at this point. I am working on a commit now to add checks and exceptions for values going out of bounds (with tests!). 

There will probably be another commit after that to get the core passing with checkstyle turned on that I will include all of the white space changes into.

Rob.

busbey:
Please make sure the merge commit mentions closing out both #384 and #126

core/src/main/java/com/yahoo/ycsb/generator/AcknowledgedCounterGenerator.java
```diff
@@ -28,16 +28,16 @@
...
 
 	private final ReentrantLock lock;
 	private final boolean[] window;
-	private volatile int limit;
+	private volatile long limit;
 
 	/**
 	 * Create a counter that starts at countstart.
 	 */
-	public AcknowledgedCounterGenerator(int countstart)
```
busbey: for the generators, could we maintain binary compatibility by leaving in int versions of the methods and marking them Deprecated?

core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
```diff
@@ -165,7 +165,7 @@ public long getIntendedtartTimeNs() {
    * Report a single value of a single metric. E.g. for read latency, operation="READ" and latency is the measured
    * value.
    */
-  public void measure(String operation, int latency)
+  public void measure(String operation, long latency)
```
busbey: why are we moving from int -> long on latencies in this patch? do we expect individual latencies to exceed ~35 minutes?

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -292,33 +292,33 @@
...
 
-	int recordcount;
+	long recordcount;
 
-    private Measurements _measurements = Measurements.getMeasurements();
+  private Measurements _measurements = Measurements.getMeasurements();
 	
-	protected static IntegerGenerator getFieldLengthGenerator(Properties p) throws WorkloadException{
-		IntegerGenerator fieldlengthgenerator;
+	protected static LongGenerator getFieldLengthGenerator(Properties p) throws WorkloadException{
```
busbey: We're already using this patch to go from tables with MAX_LONG instead of MAX_INT. Let's save moving from individual fields being up to MAX_INT to MAX_LONG in a different patch.

cmccoy:
Any ideas on why the travis tests are failing? `Too many unacknowledged insertion keys`

allanbank:
Yes - A change from int->long broke the code in a "some times we get
lucky" way.  I have the fix already but have not updated the patch yet.

allanbank:
I am going to close this pull and start a new one as I just noticed that I pushed the branch directly to the main git repo.  Sorry about that.

PR #444, opened 2015-10-22
Add Kudu bindings

jdcryans:
Currently doesn't include unit tests since we don't have mock objects for that.

### DIFF
--- a/kudu/README.md
+++ b/kudu/README.md
@@ -29,16 +29,17 @@ bin/ycsb load kudu -P workloads/workloada
 ```
 
 Additional configurations:
-* `masterQuorum`: The master's address. The default configuration expects a master on localhost.
-* `pre_split_num_tablets`: The number of tablets (or partitions) to create for the table. The default
+* `kudu_master_addresses`: The master's address. The default configuration expects a master on localhost.
+* `kudu_pre_split_num_tablets`: The number of tablets (or partitions) to create for the table. The default
 uses 4 tablets. A good rule of thumb is to use 5 per tablet server.
-* `table_num_replicas`: The number of replicas that each tablet will have. The default is 3. Should
+* `kudu_table_num_replicas`: The number of replicas that each tablet will have. The default is 3. Should
 only be configured to use 1 instead, for single node tests.
-* `sync_ops`: If the client should buffer data before sending it. The default is false. Should
+* `kudu_sync_ops`: If the client should buffer data before sending it. The default is false. Should
 always be set to true for the run phase.
+* `kudu_block_size`: The data block size used to configure columns. The default is 4096 bytes.
 
 Then, you can run the workload:
 
 ```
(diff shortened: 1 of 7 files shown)


kudu/src/main/java/com/yahoo/ycsb/db/KuduYCSBClient.java
```diff
@@ -0,0 +1,322 @@
...
+
+    client = new KuduClient.KuduClientBuilder(masterQuorum)
+        .defaultSocketReadTimeoutMs(DEFAULT_SLEEP)
+        .defaultOperationTimeoutMs(DEFAULT_SLEEP)
+        .build();
+    if (debug) {
+      System.out.println("Connecting to the masters at " + masterQuorum);
+    }
+
+    List<ColumnSchema> columns = new ArrayList<ColumnSchema>(11);
```
cmccoy: There's no guarantee that there will be 10 non-key columns. Can you make this a variable based on the value of `fieldcount`?

kudu/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Use the following command line to load the initial data into an existing Kudu cluster with default
+configurations.
+
+```
+bin/ycsb load kudu -P workloads/workloada
+```
+
+Additional configurations:
+* `masterQuorum`: The master's address. The default configuration expects a master on localhost.
```
cmccoy: Consider prefixing these with `kudu.`

kudu/README.md
```diff
@@ -0,0 +1,45 @@
...
+bin/ycsb load kudu -P workloads/workloada
+```
+
+Additional configurations:
+* `kudu_master_addresses`: The master's address. The default configuration expects a master on localhost.
+* `kudu_pre_split_num_tablets`: The number of tablets (or partitions) to create for the table. The default
+uses 4 tablets. A good rule of thumb is to use 5 per tablet server.
+* `kudu_table_num_replicas`: The number of replicas that each tablet will have. The default is 3. Should
+only be configured to use 1 instead, for single node tests.
+* `kudu_sync_ops`: If the client should buffer data before sending it. The default is false. Should
```
cmccoy: If this should always be true for the run phase, maybe that should be the default?

cmccoy:
Thanks! One more question, otherwise looks good.

kudu/README.md
busbey: we ask datastore bindings to default to no client-side buffering. I'd rather have this rephrased to not suggest it always be true for the run phase. we do the same to e.g. HBase, Cassandra, etc.

cmccoy: IIUC, `true` indicates that there *shouldn't* be client side buffering, so having it be true for the run phase is correct. I think the default should be true as well.

jdcryans: Ah yeah my description here is off. True means it won't buffer. I'll make it default to that and change that line.

cmccoy:
Thanks, @jdcryans!

PR #449, opened 2015-10-26
[core] Checkstyle enforcement for the core module.

allanbank:
Most common changes:
- tabs to spaces and other whitespace reformating.
- shorten lines over 80 chars.
- make class members private
- remove leading '_' in variable names.
- add package-info.java


allanbank:
Lots of churn in this code review but I think doing this first will make the move from int->long update easier to review.

cmccoy:
This looks a lot cleaner! It's hard to separate whitespace / name changes from structural changes (e.g., `CommandLine#setProp`) though. Could you elaborate on what changed besides names/spacing/access modifiers?

allanbank:
The setProps was due to the method length violation. I did a pretty straightforward refactoring to extract segments into methods.  There were two cases:

    mvn checkstyle:check | grep "Method length."
    /home/rjmoore/workspace/YCSB/core/src/main/java/com/yahoo/ycsb/CommandLine.java:66:7: Method length is 343 lines (max allowed is 150).
    /home/rjmoore/workspace/YCSB/core/src/main/java/com/yahoo/ycsb/Client.java:562:3: Method length is 385 lines (max allowed is 150).

You can run the following commands to find specific instances on the pre-commit baseline.

JavadocStyle: First sentence should end with a period.

    mvn checkstyle:check | grep "JavadocStyle: First sentence should end with a period."

Doing assignments in conditions: e.g., ```while( i = next() ) >= 0) {}```

    mvn checkstyle:check | grep "InnerAssignment: Inner assignments should be avoided."

Simplify expression like ```if ( expression ) { return true } else { return false }``` to just ```return expression;```:

    mvn checkstyle:check | grep "Expression can be simplified."
    
The rest is really little tedious stuff like the order of modifiers (ModifierOrder) making utility classes final and hidden constructor (Utility classes), and TODO comments (TODO).

There were 3953 error so I probably missed something but that is the major stuff where I needed to stop and think.

cmccoy:
Thanks. This is fine with me, and it would be nice to clean things up. @busbey: what do you think? This will make #98 even harder to merge.

allanbank:
Wow - #98 is a huge collection of changes. 

@busbey How close are you realistically to getting that merged. Maybe we can break it into pieces and work off each part as a group effort. I am more than willing to help but to be honest all of the different indenting and brace placement even within a single class have made making targeted changes difficult.

allanbank:
Rebased on top of the Status and raw metrics updates.

stfeng:
Thanks so much for doing this @allanbank! 

On a related note, is there a coding style guideline for YCSB? If not, perhaps we can just enforce maven's checkstyle as is, and establish a rule that all Pull Request from this/some point on must have 0 checkstyle error for both new code contributed and existing files touched (which means more contributors will help cleanup existing code as time goes by.)

busbey:
Sorry I missed this conversation. #98 remains an albatross around my neck. :/

I will have some free time to pick it up again this week, I think. I'm just on the last set of reverts, then I have to handle the merge.

Frankly, so much has changed since that PR started that I don't think we should hold up improvements while waiting for it.

kruthar:
I would really love to see this get merged, scrolling through checkstyle errors in Travis and local tests is getting old. @allanbank - do you have time for one more rebase? If not I can try and take it from where you left off. Thanks.

risdenk:
I share the same sentiment as @kruthar. I tried to modify this PR and rebase it. It had a lot of conflicts. Instead, I'm planning to open a few PRs to address the checkstyle. I'll probably do a PR per file since that would result in way less effort trying to stay up to date.

risdenk:
I created the following issue #895 and some separate PRs to address this in a piecemeal fashion.

PR #450, opened 2015-10-27
Rename README file to README.md file for the following DB bindings so…

stfeng:
… the build process can generate binding specific tar balls for them. (currently they are broken because the build process looks for README.md)

dynamodb
hypertable
mapkeeper
nosqldb


stfeng:
I spent quite some time investigating why I cannot create tar ball for the dynamoDB binding, it turned out to be the naming of the README file in that directory, must be "README.md". It works now and I also decided to fix it for all bindings that have the same problem.

@busbey or @cmccoy, please take a look. Thanks.

Issue #421, opened 2015-09-29
Document commands to create tables in Cassandra bindings


busbey:
over on #418, @Thomas-Yang has got the commands needed to properly initialize tables in both the old Cassandra thrift binding and the CQL one. We should make a README file and get them in there.

Here're his findings copied for convenience:

 ```cassandra-cli``` solves the problem. It seems the keyspace and table generated by ```cassandra-cli``` and ```cqlsh``` is not totally compatible. The commands I tried out are listed as follows:

 | cassandra-10 | cqlsh
------------ | ------------- | -------------
create keyspace | create keyspace **usertable** with placement_strategy = 'org.apache.cassandra.locator.SimpleStrategy' and strategy_options = {replication_factor:1}; | create keyspace **ycsb** with replication = {'class' : 'SimpleStrategy', 'replication_factor': 1 };
create table | create column family **data** with column_type = 'Standard' and comparator = 'UTF8Type'; | create table **usertable** (y_id text primary key,field0 varchar,field1 varchar,field2 varchar,field3 varchar,field4 varchar,field5 varchar,field6 varchar,field7 varchar,field8 varchar,field9 varchar);

Since the ```cassandra-cli``` (the Thrift API) will be removed soon (from v3.0 I guess), I am not sure it worths the effort to improve the consistency. Anyway, I do hope the above commands are useful for people still want to try out both bindings.

cmccoy:
These are now documented in `cassandra/README.md` (Thrift) and `cassandra2/README.md` (CQL).

Issue #293, opened 2015-06-12
Test CQL binding with Cassandra 2.1


busbey:
I believe Cassandra 2.1 is the current stable line. We know the CQL binding works with a Cassandra 2.0 cluster, need to try for 2.1.

busbey:
Note that we have confirmed accounts of not working on 2.1+

cmccoy:
I have a patch (cmccoy@3e8d7bb1aa314e42e65161ede4d8a260e51979e6), but it's a bit awkward - the incompatibility in the current implementation is that `Cluster#shutdown()` has been renamed `Cluster#close()`, so I added an abstract CQL client with an overridable cleanup method - for < 2.0 it's `Cluster#shutdown()`, for >= 2.0 it's `Cluster#close()`.

Anyone have a cleaner idea?

cmatser:
Does the old method compile with the new code... with some sort of deprecated warning?  or does it fail to build with the >= 2.0 cassandra libraries?

I'm wondering if we can create a property which defaults to the newer >= 2.0 method, but can be set to use the older <2.0 if desired.  Some sort of "cassandra_pre2=true|false" property.  Would that be cleaner?

Otherwise, you're talking about two different clients in the concrete classes, yes?  Something like CassandraCQLClientGreaterThanOrEqual2_0 and CassandraCQLClientLessThan2_0 would have to be specified on the run command line.   (Obviously we'd pick better names.  :)  )

What is the issue that you feel what you have is not quite a clean solution?

Thanks for tracking this down, btw!

cmccoy:
> Does the old method compile with the new code... with some sort of deprecated warning? or does it fail to build with the >= 2.0 cassandra libraries?

No, it fails to compile. And it turns out my solution doesn't work - the signature for `Session#execute(...)` also changed, so the CQL common code can't be compiled to work with both 2.0 and 1.0.

I don't like to do this, but maybe we should just duplicate the CQL binding targeting a new database `cassandra2`, as has been done for HBase.

cmccoy:
> Thanks for tracking this down, btw!

Credit to #204 :)

cmccoy:
Closed by #454.

PR #454, opened 2015-10-29
Adds a Cassandra 2 CQL binding.

cmccoy:
I couldn't find a way to make the Cassandra CQL driver work with both versions <2.0 and >2.1, so I added a separate binding, which is almost identical to the `CassandraCQLBinding` in `cassandra/`.

Also adds integration tests for the new binding, and some brief READMEs.

Addresses #421 and #293, possibly #369.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -51,6 +51,7 @@ DATABASES = {
     "cassandra-8"  : "com.yahoo.ycsb.db.CassandraClient8",
     "cassandra-10" : "com.yahoo.ycsb.db.CassandraClient10",
     "cassandra-cql": "com.yahoo.ycsb.db.CassandraCQLClient",
+    "cassandra2-cql": "com.yahoo.ycsb.db.CassandraCQLClient",
     "couchbase"    : "com.yahoo.ycsb.db.CouchbaseClient",
     "dynamodb"     : "com.yahoo.ycsb.db.DynamoDBClient",
     "elasticsearch": "com.yahoo.ycsb.db.ElasticSearchClient",
--- a/cassandra2/pom.xml
+++ b/cassandra2/pom.xml
@@ -44,6 +44,18 @@ LICENSE file.
       <version>${project.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.cassandraunit</groupId>
+      <artifactId>cassandra-unit-shaded</artifactId>
+      <version>2.1.9.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
 </project>
--- a/cassandra2/pom.xml
+++ b/cassandra2/pom.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Copyright (c) 2012-2015 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


bin/ycsb
```diff
@@ -44,33 +44,34 @@ COMMANDS = {
 }
 
 DATABASES = {
-    "accumulo"     : "com.yahoo.ycsb.db.AccumuloClient",
```
busbey: can we defer the whitespace cleanup?

cassandra2/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.yahoo.ycsb</groupId>
+    <artifactId>binding-parent</artifactId>
+    <version>0.5.0-SNAPSHOT</version>
+    <relativePath>../binding-parent</relativePath>
+  </parent>
+
+  <artifactId>cassandra2-binding</artifactId>
+  <name>Cassandra 2.1 DB Binding</name>
```
busbey: 2.1+ ? do we know if it works with the 2.2 series?

cassandra2/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -0,0 +1,447 @@
...
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.Vector;
+import java.util.concurrent.atomic.AtomicInteger;
+
+
+/**
+ * Tested with Cassandra 2.0, CQL client for YCSB framework
```
busbey: comment is out of date, since this is 2.1+?

best to replace CQLSH in both javadocs with a pointer to hte READMEs so that we don't have to maintain multiple places.

cassandra2/src/test/java/com/yahoo/ycsb/db/CassandraCQLClientTest.java
```diff
@@ -0,0 +1,174 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
```
busbey: missing copyright statement

cassandra2/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -0,0 +1,447 @@
+/**
+ * Copyright (c) 2013 Yahoo! Inc. All rights reserved.
```
busbey: Is none of this different from the other cassandra binding?

busbey: I ask because of keeping the same copyright year here.

cassandra2/pom.xml
cmccoy: Done - I tested against 2.2.3.

bin/ycsb
cmccoy: Yes - I'll add a separate commit.

cassandra2/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
cmccoy: Mostly identical, other than the method call to shut down a `Cluster`/`Session`. Updated copyright.

cassandra2/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
cmccoy: Done.

cmccoy:
@busbey - thanks much for looking. I believe I've addressed all of your comments.

cassandra2/pom.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2012 - 2015 YCSB contributors. All rights reserved. Licensed under
```
busbey: looks like missing line breaks between copyright and license as well as within license.

busbey:
:+1: I'm on board for fixing the formatting for the copyright/license header on cassandra2/pom.xml before merging rather than waiting for me to look again.

cassandra2/pom.xml
cmccoy: Thanks - done.

PR #461, opened 2015-10-31
Add a new and optional measurement type called "RAW" which outputs every dat apoint of a run.

stfeng:
Raw datapoint output is quite useful and important for two reasons:

     1) Allows the user to do more thorough statistical analysis on the complete set of the data.

     2) Using timestamps that are logged with raw data points, it becomes much easier to investigate and correlate latency outliers at the backend service.


stfeng:
@busbey Can you please take a look at this? (or feel free to reassign if necessary). Thanks.

core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementRaw.java
```diff
@@ -0,0 +1,192 @@
...
+    measurements = new ArrayList<RawDataPoint>(1000);
+  }
+
+  @Override
+  public synchronized void measure(int latency) {
+    totalLatency += latency;
+    windowTotalLatency += latency;
+    windowOperations++;
+
+    measurements.add(new RawDataPoint(latency));
```
allanbank: Since (I am pretty sure) that measure(int) can be called from multiple threads we need to protect measurements from concurrrent access.

allanbank:
I am not sure that we want to hold all of the measurements in memory for the entire run.  For small runs it won't be an issue but for large runs (where having the details is probably more interesting IMHO) we are going to exhaust memory pretty quick.

Can we change the writing of the raw metrics to the file to be delegated to a background thread and the measure(int) method just puts the RawDataPoint onto a queue that thread is reading from.  The exportMeasurements() will then interrupt the background thread so it flushes and closes the raw file.

This may also be a good use case for using the TwoInOneMeasurement where one of the other measurements does the statistical measurement and this implementation focuses on just doing the output of the raw data points.

core/src/main/java/com/yahoo/ycsb/measurements/OneMeasurementRaw.java
stfeng: This is protected because the method is marked as "synchronized". (this is the same approach taken by other measurement type classes)

allanbank: Sorry I missed that.

stfeng:
@allanbank, thanks so much for taking a look at this. Regarding your last comment of holding measurements in memory vs flushing in the background, I've thought about that too, and I did some calculation on this before i decided to go with the current way of implementation. 

The calculation is: each RawDataPoint is about 20 bytes (including Java object overhead), and for a 10-million to 20-million operation run, it's going to take about 200MB to 400MB of RAM, which is typically less than most test machines would have nowadays. In the last few days, I was able to consistently perform 20-million operation runs on a small 1 core Virtual Machine with 2.5 GB of memory without any problem. Btw, each of my 20-million op run would take about a day to finish with the backend service I am testing against. I think, if I want to try, I can probably do a 100-million operation run on my small VM just fine too. 

Yes, for very super-long runs, probably with Billions of operations per run, my current implementation would not work for them if they are run on a small VM like the one I had. When that happens, the user can either provision a larger client machine (e.g., a GCP VM n1-standard-8 with 30GB of RAM would work fine here), or split the super-long run into multiple not-so-super long runs.

Therefore at this point I am leaning towards simplicity. The asynchronous background flushing design would add a level of complexity to the code and make it a bit harder to maintain and debug down the road. I don't see a strong need to do that right now, however, if in the future we see a need from users or issues reported by users when they use this Raw measurement type, I'd be happy to implement the background flush as you described here.

By the way, I've already documented all the memory calculation and the resource requirements in the workload_template file as part of this pull request.

-Stanley

stfeng:
Hey @allanbank, I am wondering what your thoughts are regarding this PR? As commented above, I am leaning towards simplicity at this point, and I'd prefer to add complexity only when necessary. However, if you feel strongly otherwise, please let me know. Thanks!

allanbank:
I am OK leaving it the way it is. Two things.

1) We might want to take a guess at sizing the measurements list based on the expected operations count.  Doubling the list is OK when it is small but as operation counts grow I would be surprised if there are not any pauses in the client to resize the array and copy the objects over. The other option is to change the list to a LinkedList and recalculate the memory usage.

2) I had mentioned this before but to make it more concrete: How do you feel about switching from computing the percentile statistics in ```OneMeasurementRaw``` and instead let the ```OneMeasurementHdrHistogram``` do that work? There has been a lot of work that went in to HdrHistogram and I would prefer this not be an either/or situation. e.g., in ```OneMeasurement constructOneMeasurement(String name)```.

    case RAW:
       return new TwoInOneMeasurement(name,
               new OneMeasurementHdrHistogram("Hdr"+name, _props),
               new OneMeasurementRaw("Raw"+name, _props)); 

Then in ```OneMeasurementRaw.exportMeasurements(MeasurementsExporter)``` remove everything after the 

    if (outputStream != System.out) {
      outputStream.close();
    }

Eventually, it will be nice to generalize this to allow users to select any or all types of measurements to be applied but that can be another pull request.

stfeng:
1) Good point regarding the issue with ArrayList. I am leaning towards using a LinkedList based solution ( which is your suggestion as well). It's more robust and simpler than trying to guess an initial capacity here, because doing the guess work would need to make this class parse various operations properties such as total ops count, proportions for each ops type, matching the op name to measurement name, etc, etc, which is quite flaky and hard to maintain down the road. 

With the LinkedList approach, the extra space needed is a forward and backward reference which is 16 bytes total, plus the list entry object (12 bytes). So it comes down to about 50 bytes per data point. A 5-million operation run would then take about 250MB, and  a 10-million run takes 500MB, so on, so forth. This still seems reasonable to me. I will update the doc to reflect this.

2) Regarding combining HdrHistogram with Raw, I agree it can be very useful. Meanwhile, I'd like to retain the ability to use Raw directly and separately. So I am suggesting that we add a new measurement type called "HdrHistogram+Raw", and for this type, we'd initialize a TwoInOneMeasurement() object with these two measurement classes as you outlined above. This way, for people who only cares about RAW measurement, they can just use measurementtype="Raw", and for people who wants both histogram and RAW, they can use the "HdrHistogram+Raw" type. 

Regarding the percentiles calculation done by the Raw measurement class, it is still the most accurate  computation of percentiles, and it is more accurate than HdrHistogram, I'd like to retain it if possible. 

Let me know what you think about my suggestions above, once green lighted, I will go ahead and do the necessary code changes as documented here. 

Thanks so much for all your comments so far!

stfeng:
@allanbank I've made the code changes that incorporate your latest feedback. (see commit 8bfadae) Please take a look. Thanks.

allanbank:
@cmccoy This looks ready to merge to me.  It is probably going to conflict with #469. I know @stfeng is eager to get this merged so I am happy to do the post-#469 merge.

stfeng:
@allanbank - many thanks for your code review and comments, in addition to that, highly appreciate your efforts in handling the merge conflicts!

Issue #171, opened 2014-05-20
The JDBC database interface layer constructs incomplete scan query


arnaudsjs:
When a scan query is composed in the JDBC database interface layer, the resulting query has to following shape:

SELECT * FROM "TABLE_NAME" WHERE "PRIMARY_KEY" >= "START_KEY"

This implementation causes the scan operation to return more and more results when the amount of entries in the database raises during the benchmark. As a consequence the latencies of the scan operation raise too over time (this is how the bug was detected).

I suggest the implementation should be something like the following query:

SELECT * FROM "TABLE_NAME" WHERE "PRIMARY_KEY" >= "START_KEY" ORDER BY "PRIMARY_KEY" LIMIT "MAX_SCAN_LENGTH"

This implementation takes into account the maxscanlength property and only selects the first maxscanlength properties after the startkey (order by primary key). Experiments using the MySQL database show that this implementation makes the latencies of the scan operation constant over time.

busbey:
Any idea if this is still a problem @kruthar ?

kruthar:
@busbey, I'm looking into this one and a few other open JDBC issues. If I can reproduce this I'll look into a fix, but at this point I'm not sure if it is still an issue.

kruthar:
So, this is an interesting issue. First off, it does seem to be a problem still, increasing record count increases scan latency dramatically:

100 records: 		[SCAN], AverageLatency(us), 1482.5882352941176
1000 records: 		[SCAN], AverageLatency(us), 2339.490566037736
10000 records: 	[SCAN], AverageLatency(us), 6464.980392156863
100000 records: 	[SCAN], AverageLatency(us), 42759.489795918365

At first I thought why try and support a scan in SQL in the first place? SQL doesn't really do scans perse, I guess you can simulate one visually, but it is not the same as a scan in a NoSQL database like HBase. But...sounds like people are using it I guess.

In most NoSQL systems scans are useful for retrieving sets of data that are physically adjacent to each other. You just move to the start key and advance across adjacent records. To simulate grabbing adjacent keys you have to do like proposed above, GROUP BY and LIMIT. (This is another reason why it is not really a scan, you have to apply the GROUP BY to way more records then you are actually receiving back, just doesn't seem right.) Either way, that fix suggested will not work quite as expected: 

```mysql
mysql> select * from usertable where YCSB_KEY >= 'user4' group by YCSB_KEY limit 5;
+----------+-----------------+-----------------+-----------------+
[...]

PR #463, opened 2015-11-01
[jdbc] made scan operation unimplemented

kruthar:
It was determined that since SQL does not have relevant scan
functionality, that it should be made unimplemented in the jdbc-binding.
It is possible to simulate scan results in SQL but the performance
metrics would not be comparable to a database system with true scan
functionality. The results also could not gaurantee consistency across
certain YCSB property configurations.

If users truly want to do the scan operation, the old methods will always be in
commit history and they can implement it themselves, but as is, it is broken. 
See #171.


kruthar:
See my latest in #171.

I also think that since people are/were using this, it might be a little harsh to pull the rug out from underneath them and make it unimplemented. Let the users decide about contextual relevancy of the scan results as long as YCSB can provide performance metrics around performing a scan.

PR #464, opened 2015-11-01
[accumulo] Checkstyle cleanup and enforcement for the Accumulo binding.

allanbank:
Again, a lot of whitespace changes and the normal checkstyle type errors.

2 items of note:
* Deprecated QueueElement since it is not used.
* Moved some of the exception handling to common locations to remove empty catch blocks.


cmccoy:
+1

PR #468, opened 2015-11-02
[jdbc] added LIMIT to JDBC scan operation

### LINKED ISSUE #171: The JDBC database interface layer constructs incomplete scan query

kruthar:
Now properly enforcing a scan of the given number of documents in the
scan operation using the LIMIT modifier in the cached statement.

Fixes #171.

### DIFF
--- a/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
+++ b/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
@@ -304,8 +304,8 @@ public class JdbcDBClient extends DB implements JdbcDBClientConstants {
     select.append(scanType.tableName);
     select.append(" WHERE ");
     select.append(PRIMARY_KEY);
-    select.append(" >= ");
-    select.append("?;");
+    select.append(" >= ?");
+    select.append(" LIMIT ?;");
     PreparedStatement scanStatement = getShardConnectionByKey(key).prepareStatement(select.toString());
     if (this.jdbcFetchSize != null) scanStatement.setFetchSize(this.jdbcFetchSize);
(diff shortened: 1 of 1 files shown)


cmccoy:
I think to fulfill the intent of the scan operation there should be some ordering. Workload E (which uses scans) describes:

> Application example: threaded conversations, where each scan is for the posts in a given thread (assumed to be clustered by thread id)

Allowing the DB to return, say, any record from the latter half of the database doesn't seem to match that intent.

cmccoy:
Also, from the original [YCSB paper](https://www.cs.duke.edu/courses/fall13/compsci590.4/838-CloudPapers/ycsb.pdf):

> Scan records in order, starting at a randomly chosen record key. The number of records to scan is randomly chosen.

kruthar:
OK, added an ORDER BY clause to ensure record ordering by the primary key.

allanbank:
Looks good to me. 

@cmccoy - Any other concerns.

cmccoy:
Looks good. Thanks!

PR #470, opened 2015-11-03
Add posix compatibility to distribution.

elliottneilclark:
Without this people who have a large gid will error out with:

group id '1519764623' is too big ( > 2097151 )


cmccoy:
Thanks!

PR #473, opened 2015-11-03
[jdbc] JDBC module fields fix

### LINKED ISSUE #472: JDBC binding starts fields at field1
YCSB core starts field keys at 'field0', the JDBC module assumes that they start at 'field1'.

This is an issue when dataintegrity is turned on, you would not be able to properly match keys to values because the field identifiers do not match. Currently this is not a breaking issue because no verification is being done, but should probably be fixed anyway.

This is sort of an extension issue for #132.

kruthar:
This PR updates documentation and the update operation to synchronize the JDBC module's field names with YCSB core.

Fixes #472


cmccoy:
Thanks!

PR #477, opened 2015-11-04
[DynamoDB Client] Add the support for HASH_AND_RANGE primary key type

stfeng:
The default primary key type remains exactly the same as before: HASH type. And the default behavior and processing logic also remains *exactly* the same as before.

What's added is the support to benchmark a primary key type called HASH_AND_RANGE which is one of two supported primary key types of DynamoDB. Knowing the performance characteristics of the HASH_AND_RANGE key is important for some users because using hash_and_range keys the only practical way for them to get strongly consistent query across multiple items.

See documentation in conf/dynamodb.properties for more details.


cmccoy:
Could you rebase and squash these?

stfeng:
I tried "git rebase -i HEAD~5 (or HEAD~8)" and what showed up was a set of commits from everyone, of those, 2 or 3 are check-ins from me which I want to keep in the history. 

Those "merge blah..." commits which are showing up in the UI here actually do not show up in the git rebase commands, so there is no need to squash them there.

stfeng:
Anyone want to take a look at this? 

This is pretty straightforward incremental code change (and there is no change whatsoever in the default behavior), but I'd be happy to answer any question you might have regarding this PR.

PR #480, opened 2015-11-08
Couchbase cleanup

allanbank:
Two commits.

* First updates the driver for Couchbase to the latest in Maven Central.
* Second is the checkstyle updates.

### DIFF
--- a/couchbase/src/main/java/com/yahoo/ycsb/db/package-info.java
+++ b/couchbase/src/main/java/com/yahoo/ycsb/db/package-info.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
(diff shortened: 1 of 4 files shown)


couchbase/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
```
busbey: 2015 contributors?

busbey:
Besides the year in the package doc, looks good.

+1 post rebase/fix

kruthar:
:-1:  We moved the checkstyle line limit to 120: https://github.com/brianfrankcooper/YCSB/blob/master/checkstyle.xml#L118-L120. So, all the open checkstyle PR's probably need to be updated as they implement a line length of 80. I had planned on helping with a few of them at least, just not there yet

cmatser:
80 < 120.  What's the problem?  I was just thinking that it doesn't seem necessary to change a file simply for line length unless it's over 120.  How were you thinking the files should change @kruthar ?

kruthar:
Well, we may be getting down to personal preference here, but the whole reason I pushed the line length change to 120 was because I felt 80 made the code very cramped and actually hindered easy reading with so many line breaks.

So, in my opinion if lines don't have to be broken because of the line length restriction then they shouldn't be, unless the act of breaking them helps readability. There are a lot of instances in this PR where lines were broken because of the line length restriction at 80, not because the line break helped readability.

cmatser:
Ah, so you're proposing to back out the changes that were done simply to meet the 80 length criteria.  That's all good to me.  I thought for a moment, that you wanted to go through and make lines longer, just because you could.  :)

kruthar:
No, I would not suggest going in to make any unchanged lines longer unless they are just egregiously broken.

PR #483, opened 2015-11-08
[gemfire] Checkstyle updates for GemFire.

allanbank:


busbey:
could we do this on top of #388?

risdenk:
@allanbank updated this PR with #648.

PR #491, opened 2015-11-09
[tarantool] Checkstyle updates for the Tarantool binding.

allanbank:


tarantool/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
```
busbey: year and attribution are incorrect.

busbey:
+1 fine to fix copyright on package-info before merge.

bigbes:
Seems OK

risdenk:
Updated PR with #649

PR #493, opened 2015-11-09
Nosqldb re-added to the build and checkstyle cleanup

allanbank:
I found the client driver in Maven Central so I re-added the binding to the build.

### DIFF
--- a/nosqldb/src/main/java/com/yahoo/ycsb/db/package-info.java
+++ b/nosqldb/src/main/java/com/yahoo/ycsb/db/package-info.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
+ *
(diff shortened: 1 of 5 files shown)


nosqldb/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
```
busbey: Is this year and attribution correct?

busbey:
+1

nit on the year/attribution for the package docs can be fixed before merge.

verified the license of the client drivers as ASLv2 over here: http://www.oracle.com/technetwork/database/database-technologies/nosqldb/downloads/index.html

stfeng:
LGTM.

PR #498, opened 2015-11-13
[core] added return codes to VERIFY

### LINKED ISSUE #479: VERIFY measurements could be improved
When you set dataintegrity=true you receive [VERIFY] measurements at the end of workloads that have a proportion of reads.  These measurements are supposed to reflect whether the read info is coming back with expected values or not. Logic for that is in Coreworkload: https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java#L596-L611.

```
[VERIFY], Operations, 9.0
[VERIFY], AverageLatency(us), 1000.0
[VERIFY], MinLatency(us), 1000.0
[VERIFY], MaxLatency(us), 1000.0
[VERIFY], 95thPercentileLatency(us), 1000.0
[VERIFY], 99thPercentileLatency(us), 1000.0
[READ], Operations, 9.0
[READ], AverageLatency(us), 31631.11111111111
[READ], MinLatency(us), 1601.0
[READ], MaxLatency(us), 228735.0
[READ], 95thPercentileLatency(us), 228735.0
[READ], 99thPercentileLatency(us), 228735.0
[READ], Return=OK, 9
```

Currently the measurement is being tracked with three different (arbitrary) latency values, whereas actually reporting the status and getting "RETURN=OK 100" like reports would be much more useful as it would tell you exactly how many readings are incorrect.

This could probably be achieved by adding a line similar to https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/DBWrapper.java#L95 into verifyRow(). A little work would have to be done there to conform to the new Status readouts.

kruthar:
Changed measurement latency to actual latency time of the verifyRow step and now recording Status codes for verify operations.

Fixes #479

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
+++ b/core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
@@ -172,16 +172,6 @@ public class CoreWorkload extends Workload
    */
   private boolean dataintegrity;
 
-  /**
-   * Response values for data integrity checks.
-   * Need to be multiples of 1000 to match bucket offsets of
-   * measurements/OneMeasurementHistogram.java.
-   */
-  private final int DATA_INT_MATCH = 0;
-  private final int DATA_INT_DEVIATE = 1000;
-  private final int DATA_INT_UNEXPECTED_NULL = 2000;
-
-
 	/**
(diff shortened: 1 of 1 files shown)


cmccoy:
Thanks!

PR #500, opened 2015-11-18
[googledatastore] Add a Google Datastore client

stfeng:
Add a DB client for [Google Cloud Datastore](https://cloud.google.com/datastore/docs/concepts/overview?hl=en)

This Pull Request includes implementation of all CRUD operations (except scan), and integration into the rest of YCSB (pom.xml changes, etc).

### DIFF
--- a/googledatastore/README.md
+++ b/googledatastore/README.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2010 Yahoo! Inc., 2012 - 2015 YCSB contributors.
+Copyright (c) 2015 YCSB contributors.
 All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License"); you
--- a/googledatastore/README.md
+++ b/googledatastore/README.md
@@ -25,7 +25,8 @@ https://cloud.google.com/datastore/docs/concepts/overview?hl=en
     YCSB_HOME - YCSB home directory
     DATASTORE_HOME - Google Cloud Datastore YCSB client package files
 
-Please refer to https://github.com/brianfrankcooper/YCSB/wiki/Using-the-Database-Libraries for more information on setup.
+Please refer to https://github.com/brianfrankcooper/YCSB/wiki/Using-the-Database-Libraries
+for more information on setup.
 
 # Benchmark
 
@@ -36,5 +37,55 @@ Please refer to https://github.com/brianfrankcooper/YCSB/wiki/Using-the-Database
 
     $DATASTORE_HOME/conf/googledatastore.properties
 
-# FAQs
+# Details
+
+A. Configuration and setup:
+
+See this link for instructions about setting up Google Cloud Datastore and
+authentication:
+
+https://cloud.google.com/datastore/docs/getstarted/start_java/
+
+After you setup your environment, you will have 3 pieces of information ready:
+- datasetId,
+- service account email, and
+- a private key file in P12 format.
+
+These will be configured via corresponding properties in the googledatastore.properties file.
+
+B. EntityGroupingMode
+
+In Google Datastore, Entity Group is the unit in which the user can
+perform strongly consistent query on multiple items; Meanwhile, Entity group
+also has certain limitations in performance, especially with write QPS.
+
+We support two modes here:
+
+1. [default] One entity per group (ONE_ENTITY_PER_GROUP)
+
+In this mode, every entity is a "root" entity and sits in one group,
+and every entity group has only one entity. Write QPS is high in this
+mode (and there is no documented limitation on this). But query across
+multiple entities are eventually consistent.
+
+When this mode is set, every entity is created with no ancestor key (meaning
+the entity itself is the "root" entity).
+
+2. Multiple entities per group (MULTI_ENTITY_PER_GROUP)
+
+In this mode, all entities in one benchmark run are placed under one
+ancestor (root) node therefore inside one entity group. Query/scan
+performed on these entities will be strongly consistent but write QPS
+will be subject to documented limitation (current is at 1 QPS).
+
+Because of the write QPS limit, it's highly recommended that you rate
+limit your benchmark's test rate to avoid excessive errors.
+
+The goal of this MULTI_ENTITY_PER_GROUP mode is to allow user to
+benchmark and understand performance characteristics of a single entity
+group of the Google Datastore.
+
+While in this mode, one can optionally specify a root key name. If not
+specified, a default name will be used.
+
 
--- a/googledatastore/README.md
+++ b/googledatastore/README.md
@@ -1,0 +1,40 @@
+<!--
+Copyright (c) 2010 Yahoo! Inc., 2012 - 2015 YCSB contributors.
+All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+# Google Cloud Datastore Binding
+
+https://cloud.google.com/datastore/docs/concepts/overview?hl=en
+
+## Configure
+
+    YCSB_HOME - YCSB home directory
+    DATASTORE_HOME - Google Cloud Datastore YCSB client package files
+
+Please refer to https://github.com/brianfrankcooper/YCSB/wiki/Using-the-Database-Libraries for more information on setup.
+
+# Benchmark
+
+    $YCSB_HOME/bin/ycsb load googledatastore -P workloads/workloada -P googledatastore.properties
+    $YCSB_HOME/bin/ycsb run googledatastore -P workloads/workloada -P googledatastore.properties
+
+# Properties
+
+    $DATASTORE_HOME/conf/googledatastore.properties
+
+# FAQs
+
--- a/googledatastore/conf/googledatastore.properties
+++ b/googledatastore/conf/googledatastore.properties
@@ -17,6 +17,76 @@
 # Sample property file for Google Cloud Datastore DB client 
 
 ## Mandatory parameters
+# See this link for instructions about setting up Google Cloud Datastore and
+# authentication:
+# https://cloud.google.com/datastore/docs/getstarted/start_java/
+#
+# After you setup your environment, you will have 3 pieces of information ready:
+# datasetId, service account email, and a private key file. These must be
+# configured via the properties below
+#
+# googledatastore.datasetId=<string id of your dataset>
+# googledatastore.privateKeyFile=<full path to your private key file>
+# googledatastore.serviceAccountEmail=<Your service account email>
+
+# Google Cloud Datastore's read and update APIs do not support
+# reading or updating a select subset of properties for an entity.
+# (as of version v1beta2-rev1-3.0.2)
+# Therefore, it's recommended that you set writeallfields and readallfields
+# to true to get stable and comparable performance numbers.
+writeallfields = true
+readallfields = true
 
 ## Optional parameters
+#
+# Decides the consistency level of read requests. Acceptable value is:
+# EVENTUAL, STRONG (default is EVENTUAL)
+#
+# googledatastore.readConsistency=EVENTUAL
 
+# Decides how we group entities into entity groups. In Google Datastore,
+# Entity Group is the unit in which the user can perform strongly
+# consistent query on multiple items; Meanwhile, Entity group also has
+# certain limitations in performance, especially with write QPS.
+#
+# We support two modes here:
+#
+# 1. [default] One entity per group (ONE_ENTITY_PER_GROUP)
+#
+# In this mode, every entity is a "root" entity and sits in one group,
+# and every entity group has only one entity. Write QPS is high in this
+# mode (and there is no documented limitation on this). But query across
+# multiple entities are eventually consistent.
+#
+# When this mode is set, every entity is created with no ancestor key (meaning 
+# the entity itself is the "root" entity).
+#
+# 2. Multiple entities per group (MULTI_ENTITY_PER_GROUP)
+#
+# In this mode, all entities in one benchmark run are placed under one
+# ancestor (root) node therefore inside one entity group. Query/scan
+# performed on these entities will be strongly consistent but write QPS
+# will be subject to documented limitation (current is at 1 QPS).
+#
+# Because of the write QPS limit, it's highly recommended that you rate
+# limit your benchmark's test rate to avoid excessive errors.
+#
+# The goal of this MULTI_ENTITY_PER_GROUP mode is to allow user to
+# benchmark and understand performance characteristics of a single entity
+# group of the Google Datastore.
+#
+# While in this mode, one can optionally specify a root key name. If not
+# specified, a default name will be used.
+#
+# googledatastore.entityGroupingMode=ONE_ENTITY_PER_GROUP
+
+# If you set the googledatastore.entityGroupingMode property to 
+# MULTI_ENTITY_PER_GROUP, you can optionally specify the name of the root entity
+#
+# googledatastore.rootEntityName="YCSB_ROOT_ENTITY"
+
+# Strongly recommended to set to uniform.
+# requestdistribution = uniform
+
+# Enable/disable debug message, default is false.
+# googledatastore.debug = false
--- a/googledatastore/conf/googledatastore.properties
+++ b/googledatastore/conf/googledatastore.properties
@@ -1,0 +1,22 @@
+# Copyright (c) 2015 YCSB contributors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you
+# may not use this file except in compliance with the License. You
+# may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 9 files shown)


googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,334 @@
...
+      }
+    }
+
+    this.rootEntityName = getProperties().getProperty(
+        "googledatastore.rootEntityName", "YCSB_ROOT_ENTITY");
+
+    try {
+      // Setup the connection to Google Cloud Datastore with the credentials
+      // obtained from the configure.
+      DatastoreOptions.Builder options = new DatastoreOptions.Builder();
```
cmccoy: Can this use [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials#callingjava)?

stfeng: I think if we are running on GCE VM, we could use application default credentials if we create the VM with a special flag. ([here](https://cloud.google.com/datastore/docs/getstarted/start_java/) is the datastore doc about this.)

However, if we want this tool to be able to run from somewhere else other than a GCE VM, then we need to take a more generic approach to setup and use credentials (as shown in step 2 in the doc linked above), which is the approach I'm taking here.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,334 @@
...
+
+      // DatastoreException.getCode() returns an HTTP response code which we
+      // will bubble up to the user as part of the YCSB Status "name".
+      return new Status("ERROR-" + exception.getCode(), exception.getMessage());
+    }
+
+    if (response.getFoundCount() == 0) {
+      return new Status("ERROR-404", "Not Found, key is: " + key);
+    }
+
```
kruthar: Is there possibility of response.getFoundCount() > 1? This would be an unexpected result, but would get be marked as key not found.

kruthar:
The documentation in googledatastore.properties is great, but looking at other bindings the convention seems to be to document binding specific configuration in the bindings README.

kruthar:
Also, can you add checkstyle enforcement to your pom.xml and fix any issues that come up in the build? See here: https://github.com/brianfrankcooper/YCSB/blob/master/accumulo/pom.xml#L67-L90

googledatastore/conf/googledatastore.properties
```diff
@@ -0,0 +1,92 @@
...
+# authentication:
+# https://cloud.google.com/datastore/docs/getstarted/start_java/
+#
+# After you setup your environment, you will have 3 pieces of information ready:
+# datasetId, service account email, and a private key file. These must be
+# configured via the properties below
+#
+# googledatastore.datasetId=<string id of your dataset>
+# googledatastore.privateKeyFile=<full path to your private key file>
+# googledatastore.serviceAccountEmail=<Your service account email>
```
kruthar: GoogleDatastoreClient.java is actually expecting ```googledatastore.serviceAccount``` as the property name.

kruthar:
When testing this I received this:

```sh
com.yahoo.ycsb.DBException: I/O error connecting to the datastore: toDerInputStream rejects tag type 123
	at com.yahoo.ycsb.db.GoogleDatastoreClient.init(GoogleDatastoreClient.java:177)
	at com.yahoo.ycsb.DBWrapper.init(DBWrapper.java:63)
```

The fix was to simply use a P12 private key file instead of a Json key file. Is that a know limitation, and hence should be documented in the README or somewhere? Or was that just an issue with my particular setup?

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
stfeng: Good point! I've added a check for getFoundCount()>1 and return unexpected state from there. Since we only requested to lookup for one key, so we are only expecting one entity back.

googledatastore/conf/googledatastore.properties
stfeng: Yep, I agree this variable name should be serviceAccountEmail. And thanks for the catch!

stfeng:
@kruthar, Thanks for the great code review! I've incorporated all your comments into my code. Here are  answers to your questions:

1. Yes the Google Cloud Datastore needs a p12 private key file to authenticate. I added a note to README.md to clarify this.

2. Checkstyle: I've added the checkstyle plugin to pom.xml and fixed all issues (except one issue about TODO comment, which I'd like to keep there until the work is actually done. :-)

3. Moved documentation from .properties file to READMD.md

PTAL, thanks!

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,339 @@
...
+    try {
+      // Setup the connection to Google Cloud Datastore with the credentials
+      // obtained from the configure.
+      DatastoreOptions.Builder options = new DatastoreOptions.Builder();
+      Credential credential = DatastoreHelper.getServiceAccountCredential(
+          serviceAccountEmail, privateKeyFile);
+      logger.info("Using JWT Service Account credential.");
+      logger.info("DatasetID: " + datasetId + "\nService Account Email: " +
+          serviceAccountEmail + "\nPrivate Key File Path: " + privateKeyFile);
```
kruthar: Log statements should be greppable single line messages, this could just be broken up into three logger.info statements instead of using newlines.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,339 @@
...
+      throw new DBException("Security error connecting to the datastore: " +
+            exception.getMessage());
+
+    } catch (IOException exception) {
+      throw new DBException("I/O error connecting to the datastore: " +
+            exception.getMessage());
+    }
+
+    logger.info("Datastore client instance created:\n" +
+        datastore.toString());
```
kruthar: Same here, this could just be a single log line.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,339 @@
...
+  public Status read(String table, String key, Set<String> fields,
+          HashMap<String, ByteIterator> result) {
+    LookupRequest.Builder lookupRequest = LookupRequest.newBuilder();
+    lookupRequest.addKey(buildPrimaryKey(table, key));
+    lookupRequest.getReadOptionsBuilder().setReadConsistency(
+        this.readConsistency);
+    // Note above, datastore lookupRequest always reads the entire entity, it
+    // does not support reading a subset of "fields" (properties) of an entity.
+
+    logger.debug("Built lookup request as:\n" + lookupRequest.toString());
```
kruthar: And once more, remove the newline so this is a single line in the log output.

kruthar:
One last set of inline comments about multiline log statements. Otherwise I built this module successfully, and ran some simple load/read/update tests and things look good!

stfeng:
Done. Incorporated all comments, changed from newlines to one line output in the debug log. 

Thank you for the CR and taking the extra step to build, test and validate this module. Much appreciated!

stfeng:
@kruthar @cmccoy 

Is this ready to merge? Thanks!

kruthar:
Looks good to me.

busbey:
Could you update the last two commits to have a message that starts with the module that's changed? Maybe squash commits as well?

googledatastore/README.md
```diff
@@ -0,0 +1,91 @@
+<!--
+Copyright (c) 2010 Yahoo! Inc., 2012 - 2015 YCSB contributors.
```
busbey: Any reason this isn't just (c) 2015 contributors?

googledatastore/pom.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+Copyright (c) 2012 - 2015 YCSB contributors. All rights reserved.
```
busbey: Why 2012-2015?

googledatastore/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>2.15</version>
```
busbey: Do we not set the checkstyle version in the root pom?

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,339 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
busbey: Does Google need to retain a by name copyright here?

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,339 @@
...
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.Vector;
+
+import javax.annotation.Nullable;
+
+/**
+ * Google Cloud Datastore Client for YCSB.
+ * Author: stfeng@
```
busbey: No author tags please.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,339 @@
...
+    ONE_ENTITY_PER_GROUP,
+    MULTI_ENTITY_PER_GROUP
+  }
+
+  private static Logger logger =
+      Logger.getLogger(GoogleDatastoreClient.class);
+
+  // Read consistency defaults to "eventual" (this is the same as other
+  // DB client, such as DynamoDB). User can override this via configure.
+  private ReadConsistency readConsistency = ReadConsistency.EVENTUAL;
```
busbey: Apologies for not having made a guide for new bindings yet. YCSB clients should default to the strictest consistency available and document consistency tuning and tradeoffs in their readme.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
busbey: Please file a follow-on issue to enable falling back to try default creds when on GCE.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,339 @@
...
+          serviceAccountEmail, privateKeyFile);
+      logger.info("Using JWT Service Account credential.");
+      logger.info("DatasetID: " + datasetId + ", Service Account Email: " +
+          serviceAccountEmail + ", Private Key File Path: " + privateKeyFile);
+
+      datastore = DatastoreFactory.get().create(
+          options.credential(credential).dataset(datasetId).build());
+
+    } catch (GeneralSecurityException exception) {
+      throw new DBException("Security error connecting to the datastore: " +
```
busbey: Can we wrap causes in all these cases instead of just grabbing the message?

googledatastore/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2015 Google Inc. All rights reserved.
```
busbey: Same coment as above re: Google as named copyright rather than the general "YCSB contributors"

googledatastore/pom.xml
kruthar: There is checkstyle validation in the root pom, but the version is not parameterized to use in the modules. Adding the checkstyle enforcement in each module is really just a way to incrementally add verification to the modules, once they all have it and the entire project is cleaned up, it will probably be best to just at failOnViolation to the root and remove all enforcements at the module level as it is largely repetitious.

googledatastore/README.md
stfeng: Done. Changed this and everywhere else to (c) 2015 YCSB contributors.

googledatastore/pom.xml
stfeng: Right, when @kruthar asked me to add the checkstyle plugin into this binding specific pom.xml file, I had the same question, then I realized the parent/root checkstyle does not fail the build when there is violation, and there doesn't appear to be a way to parameterize this version of checkstyle plug-in to make it behave differently. 

This is the same way of configuration in other DB bindings, such as Accumulo. 

What's important here though is we are still sharing exactly the same set of checkstyle rules and configurations (../checkstyle.xml) with the rest of the YCSB code base.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
stfeng: I'm totally fine setting this default to STRONG here. Happy to do that.

Meanwhile, I also think we should set other DB clients default to conform with this guideline too. I've filed issue #509 to set the DynamoDB client's read consistency default level to STRONG so it becomes comparable to the rest of the YCSB clients at default setting.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
stfeng: Done. Filed #510

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
stfeng: Done.

stfeng:
@busbey, I've incorporated all your comments and pushed the changes. I've modified the commit message to all start with [googledatastore] and squashed a couple of commits. 

PTAL. Thanks.

stfeng:
@busbey Could you please take a look and let me know if there is anything else I need to do to get this merged? Thanks.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
busbey: Sorry for not making this clear earlier, by "wrap" I mean to use the [two argument constructor that takes a Throwable cause](https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/DBException.java#L40).

e.g.

```java
  throw new DBException("Security error connecting to the datastore", exception)
```

busbey:
the DBException comment above is the only issue I see.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
stfeng: Done.

stfeng:
Done. Changed the DBException() to take the cause. PTAL. Thanks.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
busbey: should be exception rather than exception.getCause

stfeng: Agree! Done.

stfeng:
Pushed the last commit (and squashed it into the previous one) that addressed the last comment. Thanks for all the reviews.

busbey:
thanks for the contribution!

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
k1xme: @stfeng can we somehow support both ADC and credential file authentication in there? Like adding a flag to switch between these 2 modes. This will be very handy.

googledatastore/src/main/java/com/yahoo/ycsb/db/GoogleDatastoreClient.java
```diff
@@ -0,0 +1,338 @@
...
+        result.put(name, new StringByteIterator(properties.get(name)
+            .getStringValue()));
+      }
+    }
+
+    return Status.OK;
+  }
+
+  @Override
+  public Status scan(String table, String startkey, int recordcount,
```
k1xme: Why don't we implement this method? Does YCSB not call it?

googledatastore/conf/googledatastore.properties
RogerKirby: What exactly is the googledatastore.datasetId=<string id of your dataset> referring to? I read on another forum that it is Project ID but when I update the properties file with the Project ID the following error is returned:

site.ycsb.DBException: Required property "datasetId" missing.
        at site.ycsb.db.GoogleDatastoreClient.init(GoogleDatastoreClient.java:113)
        at site.ycsb.DBWrapper.init(DBWrapper.java:86)
        at site.ycsb.ClientThread.run(ClientThread.java:91)
        at java.base/java.lang.Thread.run(Thread.java:832)

I've tried using the Project Number but to no avail.

PR #505, opened 2015-11-20
Add class OneMeasurementTimeSeriesOps to count ops per time (tps).

henrikingo:
Hi

I wanted YCSB to output the actual ops-per-second, not just summaries, so I've added a class and a configuration that does that. If you could merge this, would be great, so I don't have to patch YCSB every time I do testing.

Add class OneMeasurementTimeSeriesOps to count ops per time (tps).
Add option measurementtype = hdrhistogram+timeseriesops

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
+++ b/core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
@@ -41,6 +41,7 @@ public class Measurements {
     HDRHISTOGRAM_AND_HISTOGRAM,
     HDRHISTOGRAM_AND_RAW,
     TIMESERIES,
+    HDRHISTOGRAM_AND_TIMESERIES,
     RAW
   }
 
@@ -107,6 +108,10 @@ public class Measurements {
     {
       _measurementType = MeasurementType.TIMESERIES;
     }
+    else if (mTypeString.equals("hdrhistogram+timeseries"))
+    {
+      _measurementType = MeasurementType.HDRHISTOGRAM_AND_TIMESERIES;
+    }
     else if (mTypeString.equals("raw"))
     {
       _measurementType = MeasurementType.RAW;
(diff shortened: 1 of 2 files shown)


kruthar:
The ops per time unit is interesting functionality to add to the Timeseries measurement. I think with just a little more work it could be generalized to be useful in more cases. Currently you are only ever going to see the ops per time unit measurement with your added ```measurementtype=hdrhistogram+timeseries``` addition, whereas it could also be useful standalone via something like ```measurementtype=timeseriesops```.

Was the motivation to combine with hdrhistogram because of the percentile outputs are not present in timeseries? All the other measurements look to be just about the same (ex. AverageLatency(us)). If so it might be better to look into adding the percentile output into the new Timeseries measurement, that way there is less time spent measuring and less repetitive measurement output.

henrikingo:
Yes, I wanted the latencies as percentiles vs tps as a timeseries. As for your suggestions:

- it would of course be trivial to add a standalone timeseriesops option. Can do that (subject to below choices).
- can add summary stats to the TimeSeriesOps class too, but from my point of view combining existing classes seemed like a nice approach wrt code reuse.
 - I think it's worth discussing whether this should be a separate class or whether I should have instead added this functionality to the existing TimeSeries. In that case, TimeSeries would output two columns, one with average latency per time unit, one with ops per time unit.
 - If you agree that adding this to the existing TimeSeries class makes sense, I'd also like to modify the output so that the first column is in seconds rather than milliseconds, and make the interval configurable (e.g. do you want ops per second, per ten seconds, per minute...)

Let me know what you'd like and I'll modify accordingly.

kruthar:
Also, you can effectively accomplish this by using: 

```sh
-s -p statusInterval=1
```

Which would yield a status readout every 1 second that looks something like this:

```sh
2015-11-20 14:16:06:073 1 sec: 78885 operations; 78855 current ops/sec; est completion in 12 seconds [VERIFY: Count=39340, Max=997, Min=0, Avg=0.12, 90=0, 99=1, 99.9=1, 99.99=3] [READ: Count=39848, Max=168, Min=0, Avg=8.37, 90=11, 99=41, 99.9=66, 99.99=120] [UPDATE: Count=40295, Max=198, Min=0, Avg=8.54, 90=12, 99=43, 99.9=69, 99.99=117]
```

This readout gives you just about every measurement per the time interval you specify, and is already implemented. You could then stick with the hdrhistogram and get your overall results and percentiles at the end.

henrikingo:
Ah, I was not aware of that. That is actually preferable then. We can ignore this pull request.

kruthar:
Great! Glad I could help.

PR #514, opened 2015-11-27
Added support for Vim and Emacs in .gitignore

mbrukman:
This change ignores temporary files created by Vim and Emacs.


busbey:
Thanks!

Issue #267, opened 2015-05-29
adopt a code style guide and update checkstyle rules accordingly


busbey:
we have a ton of checkstyle warnings right now, and it looks like it might be because we don't have rules that reflect what we want for our style.

busbey:
what are folks feelings? I'm used to sun style guide modified for 2 space indents.

cmccoy:
I don't have strong opinions as long as it's consistent (and automatable?). I also think we should delay any reformatting until most PRs outstanding PRs are merged, to minimize merge conflicts.

busbey:
+1. Maybe we do this post 0.2.0?

cmccoy:
Sounds good.

kruthar:
Started to add checkstyle enforcement to the JDBC module. Realized that 80 char line length is pretty short. Requires some weird line breaks when dealing with long class names and makes self documenting variables and methods hard.

Is the 80 char length a hard and fast rule or would anyone second increasing it to something like 120?

kruthar:
@allanbank - any input? I know you have been working hard on enforcing the checkstyle.

allanbank:
@kruthar - I don't care what the standard is. My main concern id that needs for there to be one and it to be followed. After that it falls down to being pragmatic. 

Having said that, on the line length: my one concern would be the number of characters that Github displays in the side-by-side diff view (my preference).  That looks to be about 112 characters. I am not sure if that is a limit due to my monitor/resolution or something GitHu is defaulting to. There is extra space in the margins.

kruthar:
@allanbank, agreed. I'll stick to whatever the community decides, my two cents are that 80 is a bit short. Ex. looking at your diff for the JDBC binding, you've had to break a lot of lines because of length which actually makes reading the code more difficult.

I have a big screen, 15-inch retina, so it looks like lines like: https://github.com/brianfrankcooper/YCSB/pull/494/files#diff-564e0de8a77f56fa41dedad0949c793dL405 display without scrolling and that line is like 125 chars. But I may have a bigger than typical screen.

allanbank:
@kruthar - Since no one is opposed to the 120 want to put together a pull request?  There are several of the checkstyle commit that won't apply cleanly at this point so before I redo them I'd like the extra elbow room.

kruthar:
[...]

PR #517, opened 2015-11-30
[build] changed checkstyle linelength to 120

kruthar:
Discussion for this change is in #267. Abridged - 80 character limit is too short to be practical.

This should not invalidate any checkstyle enforcement PRs already merged since we are making the limit larger. I can assist in reevaluating the modules that have already had their checkstyle enforcement merged.

@allanbank - I can help with the open checkstyle PRs as well, just let me know.


allanbank:
:+1:

Issue #326, opened 2015-06-25
Memcached in YCSB framework


bigbes:
I can add Memcached, if it's interesting for someone. There are already two key-value in-memory stores (Redis and Tarantool), stores that can't use `scan` or stores that can operate with memcached protocol, so i think this'll be interesting.

busbey:
there was a PR for a memached implementation a while back. might be worth seeing if it's viable. If you can't find the individual PR for the implementation, I know there's one on #98 from the thumbtack fork, but the current state of me getting that to a mergable state has all of thumbtack reverted.

if you prefer starting fresh instead of finding that older code, that's fine by me as well.

paulie-g:
:+1: , definitely interesting and viable now that memcached supports excluding records from lru handling (5 years late). I don't know what is available memcached-client-wise in javaland, but keep in mind that choice of client, protocol (classic or binary) and server threading model are very significant for characterisation. UDP performs much better than TCP for a single-threaded server, but the trade-offs with udp+threading are unclear (to me). Also, If I had to guess, the ycsb client might well bottleneck before memcached so that's worth looking out for.

bigbes:
@paulie-g to be honest, while testing tarantool/redis http://tarantool.org/benchmark.html - i've ran into performance bottleneck as you may see on the graphics. It's needed to start a couple of YCSB clients (using https://github.com/brianfrankcooper/YCSB/wiki/Running-a-Workload-in-Parallel)

paulie-g:
Yes, that makes sense. How close to wire-speed did you get? I wouldn't be surprised if you managed to get to a place where bandwidth or packet rate were the limiting factor.

Configurable L4 and L7 protocols for the memcached tests would be great.

p.s. Nichego, u busurman-imperalistov s russkim huzhe chem u tebya s angliyskim ;)

-pasha

bigbes:
@paulie-g Problem wasn't in the bandwidth, but in CPU. Server was 16-cpu xeon... Maybe i did something wrong?

busbey:
We have a memcached binding now. Looks like the ref to this issue didn't get parsed as closing it.

PR #518, opened 2015-12-01
[memcached] Added memcached binding.

mbrukman:
The memcached support was extracted from PR #98 by @jbellis, with cleanups to
bring it in line with current APIs and style guide.

This PR also addresses issue #326.

@busbey, could you please review? Thanks!

### DIFF
--- a/memcached/README.md
+++ b/memcached/README.md
@@ -1,0 +1,97 @@
+<!--
+Copyright (c) 2015 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+# YCSB Memcached binding
+
+This section describes how to run YCSB on memcached.
+
+## 1. Install and start memcached service on the host(s)
+
+Debian / Ubuntu:
+
+    sudo apt-get install memcached
+
+RedHat / CentOS:
+
+    sudo yum install memcached
+
+## 2. Install Java and Maven
+
+See step 2 in [`../mongodb/README.md`](../mongodb/README.md).
+
+## 3. Set up YCSB
+
+Git clone YCSB and compile:
+
+    git clone http://github.com/brianfrankcooper/YCSB.git
+    cd YCSB
+    mvn -pl com.yahoo.ycsb:memcached-binding -am clean package
+
+## 4. Load data and run tests
+
+Load the data:
+
+    ./bin/ycsb load memcached -s -P workloads/workloada > outputLoad.txt
+
+Run the workload test:
+
+    ./bin/ycsb run memcached -s -P workloads/workloada > outputRun.txt
+
+## 5. memcached Connection Parameters
+
+A sample configuration is provided in
+[`conf/memcached.properties`](conf/memcached.properties).
+
+### Required params
+
+- `memcached.hosts`
+
+  This is a comma-separated list of hosts providing the memcached interface.
+  You can use IPs or hostnames. The port is optional and defaults to the
+  memcached standard port of `11211` if not specified.
+
+### Optional params
+
+- `memcached.shutdownTimeoutMillis`
+
+  Shutdown timeout in milliseconds.
+
+- `memcached.objectExpirationTime`
+
+  Object expiration time for memcached; defaults to `Integer.MAX_VALUE`.
+
+- `memcached.checkOperationStatus`
+
+  Whether to verify the success of each operation; defaults to true.
+
+- `memcached.readBufferSize`
+
+  Read buffer size, in bytes.
+
+- `memcached.opTimeoutMillis`
+
+  Operation timeout, in milliseconds.
+
+- `memcached.failureMode`
+
+  What to do with failures; this is one of `net.spy.memcached.FailureMode` enum
+  values, which are currently: `Redistribute`, `Retry`, or `Cancel`.
+
+You can set properties on the command line via `-p`, e.g.:
+
+    ./bin/ycsb load memcached -s -P workloads/workloada \
+        -p "memcached.hosts=127.0.0.1" > outputLoad.txt
--- a/memcached/conf/memcached.properties
+++ b/memcached/conf/memcached.properties
@@ -1,0 +1,52 @@
+# Copyright (c) 2015 YCSB contributors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you
+# may not use this file except in compliance with the License. You
+# may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing
+# permissions and limitations under the License. See accompanying
+# LICENSE file.
+
+#
+# Sample property file for Memcached Client
+
+## Mandatory parameters
+
+# A comma-separated list of memcached server endpoints, each being an IP or
+# hostname with an optional port; the port defaults to the memcached-standard
+# port of 11211 if not specified.
+#
+# memcached.hosts =
+
+## Optional parameters
+
+# Shutdown timeout in milliseconds.
+#
+# memcached.shutdownTimeoutMillis = 30000
+
+# Object expiration time for memcached; defaults to `Integer.MAX_VALUE`.
+#
+# memcached.objectExpirationTime = 2147483647
+
+# Whether to verify the success of each operation; defaults to true.
+#
+# memcached.checkOperationStatus = true
+
+# Read buffer size, in bytes.
+#
+# memcached.readBufferSize = 3000000
+
+# Operation timeout, in milliseconds.
+#
+# memcached.opTimeoutMillis = 60000
+
+# What to do with failures; this is one of `net.spy.memcached.FailureMode` enum
+# values, which are currently: `Redistribute`, `Retry`, or `Cancel`.
+#
+# memcached.failureMode = Redistribute
--- a/memcached/pom.xml
+++ b/memcached/pom.xml
@@ -1,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+Copyright (c) 2014-2015 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.yahoo.ycsb</groupId>
+    <artifactId>binding-parent</artifactId>
+    <version>0.6.0-SNAPSHOT</version>
+    <relativePath>../binding-parent</relativePath>
+  </parent>
+
+  <artifactId>memcached-binding</artifactId>
+  <name>memcached binding</name>
+  <groupId>com.yahoo.ycsb</groupId>
+  <packaging>jar</packaging>
+
+  <dependencies>
+    <dependency>
+      <groupId>log4j</groupId>
+      <artifactId>log4j</artifactId>
+      <version>1.2.17</version>
+    </dependency>
+    <dependency>
+      <groupId>com.yahoo.ycsb</groupId>
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.codehaus.jackson</groupId>
+      <artifactId>jackson-mapper-asl</artifactId>
+      <version>1.9.13</version>
+    </dependency>
+    <dependency>
+      <groupId>net.spy</groupId>
+      <artifactId>spymemcached</artifactId>
+      <version>2.11.4</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>2.15</version>
+        <configuration>
+          <consoleOutput>true</consoleOutput>
+          <configLocation>../checkstyle.xml</configLocation>
+          <failOnViolation>true</failOnViolation>
+          <failsOnError>true</failsOnError>
+        </configuration>
+        <executions>
+          <execution>
+            <id>validate</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>checkstyle</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-assembly-plugin</artifactId>
+        <version>${maven.assembly.version}</version>
+        <configuration>
+          <descriptorRefs>
+            <descriptorRef>jar-with-dependencies</descriptorRef>
+          </descriptorRefs>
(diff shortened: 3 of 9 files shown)


memcached/README.md
```diff
@@ -0,0 +1,45 @@
+# YCSB Memcached binding
+
+## Quick Start
+
+This section describes how to run YCSB on memcached.
+
+### 1. Install and start memcached service on the host(s)
+
+### 2. Install Java and Maven
+
```
kruthar: Can you expand a bit on 1 and 2? Maybe just a link to a relevant memcached installation page, and possible a copy and paste of Java and Maven installation instructions from the mongodb module: https://github.com/brianfrankcooper/YCSB/tree/master/mongodb

memcached/README.md
```diff
@@ -0,0 +1,45 @@
...
+Set the list of hosts in the workload you plan to run.
+
+- `memcached.hosts`
+
+This is a comma-separated list of hosts providing the memcached interface. You
+can use IPs or hostnames. The port is optional and defaults to the memcached
+standard port of `11211` if not specified.
+
+Optional parameters are listed in
+[`conf/memcached.properties`](conf/memcached.properties).
```
kruthar: Having an example conf file is great, in addition can you please document the memcached specific properties at the end of this README? The mongodb module also has a good example of how to do that.

memcached/README.md
```diff
@@ -0,0 +1,45 @@
+# YCSB Memcached binding
```
kruthar: README should also have licensing.

memcached/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+      <url>http://files.couchbase.com/maven2</url>
+    </repository>
+    <repository>
+      <id>repository.jboss.org</id>
+      <url>https://repository.jboss.org/nexus/content/groups/public</url>
+    </repository>
+  </repositories>
+
+  <build>
+    <plugins>
```
kruthar: Can you add the checkstyle enforcement plugin, build and correct any checkstyle notices? Here is an example: https://github.com/brianfrankcooper/YCSB/blob/master/mongodb/pom.xml

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
```diff
@@ -0,0 +1,239 @@
...
+    try {
+      GetFuture<Object> future = memcachedClient().asyncGet(key);
+      Object document = future.get();
+      if (document != null) {
+        fromJson((String) document, fields, result);
+      }
+      return Status.OK;
+    } catch (Exception e) {
+      logger.error("Error encountered for key: " + key, e);
+      return new Status("ERROR_READ", e.toString());
```
kruthar: Status codes are already going to be grouped by operation time ie. READ, UPDATE, etc. You are also logging the error so the user will see the exception in the output. So, it is not necessary to record a new status of "ERROR_READ", ```Status.ERROR``` should suffice.

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
```diff
@@ -0,0 +1,239 @@
...
+  public Status update(
+      String table, String key, HashMap<String, ByteIterator> values) {
+    key = createQualifiedKey(table, key);
+    try {
+      OperationFuture<Boolean> future =
+          memcachedClient().replace(key, objectExpirationTime, toJson(values));
+      return getReturnCode(future);
+    } catch (Exception e) {
+      logger.error("Error updating value with key: " + key, e);
+      return new Status("ERROR_UPDATE", e.toString());
```
kruthar: return ```Status.Error```

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
```diff
@@ -0,0 +1,239 @@
...
+  public Status insert(
+      String table, String key, HashMap<String, ByteIterator> values) {
+    key = createQualifiedKey(table, key);
+    try {
+      OperationFuture<Boolean> future =
+          memcachedClient().add(key, objectExpirationTime, toJson(values));
+      return getReturnCode(future);
+    } catch (Exception e) {
+      logger.error("Error inserting value", e);
+      return new Status("ERROR_INSERT", e.toString());
```
kruthar: return ```Status.Error```

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
```diff
@@ -0,0 +1,239 @@
...
+
+  @Override
+  public Status delete(String table, String key) {
+    key = createQualifiedKey(table, key);
+    try {
+      OperationFuture<Boolean> future = memcachedClient().delete(key);
+      return getReturnCode(future);
+    } catch (Exception e) {
+      logger.error("Error deleting value", e);
+      return new Status("ERROR_DELETE", e.toString());
```
kruthar: return ```Status.Error```

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+  }
+
+  public int query(String table, String key, int limit) {
```
kruthar: I do think it would be interesting to see query get implemented in some way in the YCSB framework. Currently it is not implemented, can this method be removed?

memcached/README.md
mbrukman: Are all the bindings not under the same license as YCSB itself (Apache 2.0)? I assume so, and I don't see any other license files in other directories (though I only sampled a few, but all the Java files I saw had Apache 2.0 header).

Is this really necessary? Even just saying that "this binding is licensed under Apache 2.0" may inadvertently imply that repo has a set of licenses, up to one for each binding, which is presumably not the message it should be sending.

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
mbrukman: Done.

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
mbrukman: Done.

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
mbrukman: Done.

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
mbrukman: Done.

memcached/src/main/java/com/yahoo/ycsb/db/MemcachedCompatibleClient.java
mbrukman: Sorry, looks like this is a holdover from the original PR; didn't notice it's not implemented in the base class and hence extraneous. Removed.

memcached/README.md
mbrukman: Done for step 1. For step 2, it seems like this is binding-agnostic, so presumably it should be documented in a single place, and every binding can just point to it.

In fact, even step 5 is the same for everything, just substituting the binding is sufficient.

What are your thoughts about moving the "install Java / Maven" to a common place and linking to it? If so, where should it be located: top-level README, perhaps?

memcached/README.md
kruthar: I was under the impression all files needed licensing. For .md files you have to look at the raw to see the licensing in the comments. Here's the example from mongodb again: https://raw.githubusercontent.com/brianfrankcooper/YCSB/master/mongodb/README.md

memcached/README.md
mbrukman: Done (copied from `conf/memcache.properties`).

memcached/README.md
mbrukman: Done.

memcached/pom.xml
mbrukman: Done.

memcached/README.md
mbrukman: To make progress on this PR, I've added a link to `../mongodb/README.md` and refer the user to follow step 2 in that doc. Let's consider moving Java / Maven instructions to a common place as a separate PR?

mbrukman:
@kruthar, thanks for the review! I think I've addressed all of your comments; please take another look.

kruthar:
uh-oh, looks like you might have rebased farther then you should have, you've replayed several commits. It looks like you should only have a single commit, but now you have 5.

mbrukman:
Rebased on `master`, re-pushed, back to a single commit. Sorry about that!

kruthar:
Last thing - It might be good to add a note in the README about how to add another MemcachedClient if someone ever wants to do so, and explain which memcached implementations the current client supports (I assume this is why you created the abstract class).

Other then that, I pulled, built and did some very basic testing and things look good to me. 

A second opinion before merge would be great.

mbrukman:
The base class was there with the original PR so I left it in, under the assumption that either the original author was going to use it, or there are other parts to the original PR that may want to use it, and I wanted to be as close to original PR to make it easy to do the large merge that @busbey is working on.

If it were up to me, I would probably allow using different Memcache client implementations such as [xmemcached](https://github.com/killme2008/xmemcached) instead of Spymemcached, but that would require a slightly different class structure (right now the signature is hardcoded to return a Spymemcached client).

mbrukman:
@kruthar, @busbey — I think I see what the issue here is: the YCSB style is to call every binding's main class `[software]Client`, which means this one should be called `MemcachedClient`. Spymemcached, the actual memcached client library, has a class by the same name. These classes cannot co-exist in the same Java file, since Java does not allow aliases or typedefs, so we have to split them up somehow.

Here are some alternatives I've come up with to make progress on this; let me know which you would prefer.

### Option 0.

Keep things as they are in this PR.

Pros:
* already done :-)

Cons:
* two classes

### Option 1.

Keep the main class in YCSB as `MemcachedClient`, wrap the Spymemcached class in another class.

Cons:
* still two classes

### Option 2.

Rename the main class in YCSB to, e.g., `MemcacheBenchmark`, use the `MemcachedClient` from Spymemcached as-is.

Pros:
* single Java file

Cons:
* breaks tradition with other YCSB bindings

### Option 3.

Rewrite this to enable substituting alternative implementations to Spymemcached: create an abstract interface, a factory class, and return Spymemcached as the only (current) implementation, but leave the possibility of adding another implementation backed by a different memcached client.

Pros:
* actually uses multiple classes to achieve something

Cons:
* even more than two files
* non-trivial amount of additional work
* not clear if anyone would benefit from any other clients than Spymemcached: to my knowledge, it's the most popular Java client for memcached

busbey:
the two classes are in different packages, so you can just use them as-is and refer to the one from spymemcached by its fully qualified name.

mbrukman:
@busbey — yes, but that only works if using wildcard imports:

```java
import net.spy.memcached.*;
public class MemcachedClient extends DB {
```

but explicit import does not work:

```java
import net.spy.memcached.MemcachedClient;
public class MemcachedClient extends DB {
```

and errors out with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project memcached-binding: Compilation failure
[ERROR] .../ycsb/memcached/src/main/java/com/yahoo/ycsb/db/MemcachedClient.java:[49,1] com.yahoo.ycsb.db.MemcachedClient is already defined in this compilation unit
```

I was under the impression that wildcard imports is bad style:

* [Stack Overflow discussion](http://stackoverflow.com/q/2067158)
* [Google's Java style guide](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)
* [OpenJDK style guide](http://cr.openjdk.java.net/~alundblad/styleguide/#toc-import-statements)

and assumed that checkstyle will disallow it (turns out, YCSB's usage of checkstyle does not, even though [it's supported](http://checkstyle.sourceforge.net/config_imports.html#AvoidStarImport)).

It turns out that other bindings in YCSB already have wildcard imports of the following variety:

```java
import java.net.*;
import java.util.*;
```

which suggests that it's OK to do in YCSB.

I've updated my PR to selectively use wildcard imports just to address this issue. Please take a look.

busbey:
It looks like you're already using the fully qualified name for the spymemcached, so importing it shouldn't be necessary at all.

mbrukman:
@busbey — hmmm, I didn't realize that would work. Thanks! Done.

memcached/pom.xml
```diff
@@ -0,0 +1,109 @@
...
+  <repositories>
+    <repository>
+      <id>files.couchbase.com</id>
+      <url>http://files.couchbase.com/maven2</url>
+    </repository>
+    <repository>
+      <id>repository.jboss.org</id>
+      <url>https://repository.jboss.org/nexus/content/groups/public</url>
+    </repository>
+  </repositories>
```
kruthar: Both of these repositories are not used and can be removed. Probably left over from a copy paste, or maybe from deciding which JSON framework to use.

mbrukman: Removed.

memcached/README.md
kruthar: That should be ok for now. I agree it should probably go into a single documented place. The top level README is an option.

mbrukman: OK, happy to do that as a follow-up PR. Anything else you'd like to see as part of this PR, or is it good to go now?

mbrukman: [PR 523](https://github.com/brianfrankcooper/YCSB/pull/523) moves Java + Maven instructions from MongoDB's README to the top-level README.

mbrukman:
@busbey, I think this PR is ready to merge. Please let me know if you have any other comments.

cmccoy:
@kruthar / @busbey - any other comments? This looks good to me.

kruthar:
Nothing more from me. :+1:

busbey:
I was planning to do another pass this evening to check for things. The only bits that I remember was the comment about containing Google copyrighted code (but not seeing any in the patch) and the copyright year since this was based on a prior PR.

mbrukman:
@busbey – Turns out, I labeled all files as "Copyright (c) 2015 YCSB contributors." even for the new files that I created, such as `package-info.java` and `README.md`; I've just updated those to be "Copyright (c) 2015 Google Inc." to be correct. That said, Google legal also told me that Google Inc. owns the copyright for the modifications that I made to the original PR, so even if it's not explicitly listed as the copyright owner for the entire file, they believe that Google Inc. should be listed in the NOTICE file.

Regarding the copyright year, I modified the copyright line to 2014-2015 for the files that came from the original PR: `MemcachedClient.java` and `pom.xml`.

busbey:
Normally we attribute copyright for changes via the author on git commits, and note it on the files via the "YCSB contributors" line. Generally, the project would prefer if Google could be covered under the banner of YCSB contributors.

If they're not, then putting them in the NOTICE file is fine so long as downstream folks can figure out which changes are attributed to Google.

memcached/README.md
```diff
@@ -0,0 +1,97 @@
+<!--
+Copyright (c) 2014-2015 Google Inc. All rights reserved.
```
busbey: is the year range here correct?

memcached/conf/memcached.properties
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2015 YCSB contributors. All rights reserved.
```
busbey: copyright correct?

busbey:
If we need to include attribution to Google on a per-file basis, then the files that are a mixture of old changes and new need to reflect both, similar to how we handle having both the old Yahoo! specific line and the contributors line on hte top-level README:

https://raw.githubusercontent.com/brianfrankcooper/YCSB/master/README.md

That said, it'd be much simpler for everyone involved if Google legal was fine with their rights being covered under "YCSB contributors".

If they aren't and we need some explicit mention, the difficulty in listing things just in NOTICE is there's no way for a downstream consumer who doesn't have access to git to turn even "contributions from @mbrukman (c) 2015 Google, Inc" into actual sections of code.

cmccoy:
@mbrukman - could you just add Google, Inc to `NOTICE.txt` in the root of the repository and leave the copyrights as "YCSB contributors" in the headers?

busbey:
wfm. maybe we can make a follow on to add a section in the NOTICE that explains "YCSB Contributors" with a link to a page that explains how to find people via git history.

memcached/conf/memcached.properties
mbrukman: Per discussion below, we're keeping it as `Copyright (c) 2015 YCSB contributors.` so yes, this is now (retroactively) correct.

memcached/README.md
mbrukman: You're right, this was not created in 2014. Updated to 2015, also updated it to be "YCSB contributors" per discussion below.

mbrukman:
I've updated the copyright headers to be `Copyright (c) <YEAR> YCSB contributors.` as agreed to above. I've fixed the years per above comments for files I created in 2015 that were not part of the original PR that this change is based on.

Also, the easy way of finding contributions that are (c) Google Inc. would be to look for commits by users with `@google.com` email addresses, which is more correct than saying "all contributions from GitHub user xyz are (c) Company Inc." because not all contributions under a single GitHub user id are owned by a single corporation.

busbey:
> Also, the easy way of finding contributions that are (c) Google Inc. would be to look for commits by users with @google.com email addresses, which is more correct than saying "all contributions from GitHub user xyz are (c) Company Inc." because not all contributions under a single GitHub user id are owned by a single corporation.

That's a really good point.

mbrukman:
Thank you @busbey, @kruthar, and @cmccoy for the reviews and feedback!

PR #523, opened 2015-12-05
Migrated Java and Maven docs to root.

mbrukman:
Moved installation instructions from MongoDB to top-level README per discussion in [PR 518](https://github.com/brianfrankcooper/YCSB/pull/518#discussion_r46354680).

Questions:

1. Should the default installation instructions use distributions' native packaging mechanisms, rather than direct users to download from Oracle and Apache Maven's websites?

  We can direct users for installation of the latest versions via tarballs/packages from Oracle and Maven, rather than specify URLs to specific versions here (which are almost always out-of-date).

  Specifying direct URLs implicitly suggests that YCSB requires this specific version of each dependency, which doesn't appear to be the case.

2. Is Oracle JDK the only supported one? What about OpenJDK?

We can resolve both of these questions separately from this PR, but I'm raising them here since these questions were prompted by making this change and pushing the instructions so prominently on the main top-level README.


busbey:
I'd rather we move the "java / maven" stuff into docs rather than the top README. Essentially, I'd like to see something like [our guide for windows](https://github.com/brianfrankcooper/YCSB/wiki/Prerequisites-for-Windows) for other platforms.

Related, I'd answer question number 1 with "definitely favor native packaging" unless it doesn't get us a sufficient version #.

for number 2, I believe our [travis-ci setup](https://travis-ci.org/brianfrankcooper/YCSB) currently tests oracle jdk and openjdk 1.7, so both ought to be fine. (we should probably start testing java 8 soon)

mbrukman:
@busbey — where would you like these docs and in what format?

The link you pointed to for Windows is on the wiki, but I recall you [saying earlier](https://github.com/brianfrankcooper/YCSB/pull/98#issuecomment-159506665) that you wanted to migrate from wiki to code such that you could get regular PRs for changes to the docs instead.

Assuming you meant the `doc` directory, it currently contains HTML, but I'm guessing that MarkDown is better for editing and reviewing, and we can generate HTML from it—even automatically, using GitHub functionality.

What do you think?

mbrukman:
@busbey — I moved the docs from `README.md` -> `doc/GettingStartedLinux.md` and link to the Windows and Linux docs from the top-level README, which is now much cleaner.

Would that work?

busbey:
Yeah, this approach wfm. Sorry for the lag.

doc/GettingStartedLinux.md
```diff
@@ -0,0 +1,81 @@
+# Getting Started on Linux
+
+1. Install Java and Maven
+
+  Go to [Java SE Downloads page](http://www.oracle.com/technetwork/java/javase/downloads/index.html)
```
busbey: Can we include a note about openjdk like we do on the windows page?

doc/GettingStartedLinux.md
```diff
@@ -0,0 +1,81 @@
...
+  Or install via `yum` or `apt-get`:
+
+  ```sh
+  sudo yum install java-devel
+  ```
+
+  Download and install [Maven](http://maven.apache.org/download.cgi), e.g.:
+
+  ```sh
+  wget http://ftp.heanet.ie/mirrors/www.apache.org/dist/maven/maven-3/3.1.1/binaries/apache-maven-3.1.1-bin.tar.gz
```
busbey: We should include a note about verifying the download.

i.e. https://www.apache.org/info/verification.html

doc/GettingStartedLinux.md
```diff
@@ -0,0 +1,81 @@
...
+  sudo tar xzf apache-maven-*-bin.tar.gz -C /usr/local
+  cd /usr/local
+  sudo ln -s apache-maven-* maven
+  sudo vi /etc/profile.d/maven.sh
+  ```
+
+  Add the following to `maven.sh`:
+
+  ```sh
+  export M2_HOME=/usr/local/maven
```
busbey: Is this still needed for Maven 3?

doc/GettingStartedLinux.md
```diff
@@ -0,0 +1,81 @@
...
+  export PATH=${M2_HOME}/bin:${PATH}
+  ```
+
+  Reload `bash` and test `mvn`:
+
+  ```sh
+  bash
+  mvn -version
+  ```
```
busbey: An example of what `mvn -version` outputs would help here.

doc/GettingStartedLinux.md
```diff
@@ -0,0 +1,81 @@
...
+  rpm -Uvh jdk-7u40-linux-x64.rpm
+  ```
+
+  Or install via `yum` or `apt-get`:
+
+  ```sh
+  sudo yum install java-devel
+  ```
+
+  Download and install [Maven](http://maven.apache.org/download.cgi), e.g.:
```
busbey: We don't need Maven for running from a release, right?

And we need python.

doc/GettingStartedLinux.md
```diff
@@ -0,0 +1,81 @@
...
+YCSB requires the use of Maven 3; if you use Maven 2, you may see [errors
+such as these](https://github.com/brianfrankcooper/YCSB/issues/406).
+
+To build the full distribution, with all database bindings:
+
+    mvn clean package
+
+To build a single database binding:
+
+    mvn -pl com.yahoo.ycsb:mongodb-binding -am clean package
```
busbey: It would help to end this with a pointer for where folks can find the built binary artifacts that we talk about above.

busbey:
Still working on this? If not, we'll try to find someone to handle the suggestions in a follow on PR.

mbrukman:
@busbey — yes, still working on it. Sorry for the delay; will get back to this soon.

busbey:
no worries. just send a ping when you're ready.

Issue #341, opened 2015-07-06
Improve experience running on CentOS 6


busbey:
CentOS 6 comes with Python 2.6, which *almost* works with our python bindings but not quite.

Possible solutions, in decreasing order of short-term preference:
* make the script work with python 2.6 (but keep working with 2.7)
* make a "what you need" section similar to #246 
* move off of python in favor of linux and windows specific shell glue

risdenk:
Closing this since this seems to be addressed by the above and addition of bin/yscb.sh script.

PR #526, opened 2015-12-08
[scripts] Improve errors in bin/ycsb

### LINKED ISSUE #345: improve error message when java isn't found
right now if there's no JAVA_HOME and java isn't in the path we get:

```
Traceback (most recent call last):
  File "./bin/ycsb", line 209, in <module>
    sys.exit(main())
  File "./bin/ycsb", line 205, in main
    return subprocess.call(ycsb_command)
  File "/usr/lib64/python2.6/subprocess.py", line 478, in call
    p = Popen(*popenargs, **kwargs)
  File "/usr/lib64/python2.6/subprocess.py", line 642, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.6/subprocess.py", line 1234, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

Which is pretty terrible. If there's no JAVA_HOME, do a test for the java command and give a better message. Probably the same thing should be done for where we call maven.

cmccoy:
* Handles missing "java" with an error message.
* Adds a check_output method for users running on Python 2.6.
* Prints an error when importing argparse fails (third party module on Python 2.6).

Partially addresses #341, fixes #345.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -16,13 +16,18 @@
 # LICENSE file.
 #
 
-import argparse
+import errno
 import fnmatch
 import io
 import os
 import shlex
 import sys
 import subprocess
+try:
+    import argparse
+except ImportError:
+    print >> sys.stderr, '[ERROR] argparse not found. Try installing it via "pip".'
+    raise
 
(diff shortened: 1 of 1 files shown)


allanbank:
:+1:

Issue #508, opened 2015-11-22
Make the YCSB load phase more robust


stfeng:
The YCSB load phase consists of inserting records into the DB. Currently, a single failed insertion to the DB would fail/abandon the entire load phase. For a DB service that has a 99.9% availability SLA (which isn't that bad), it can legitimately fail one insert operation out of 1000. For YCSB users, this means they will have a (very) hard time to load that DB with just a few thousands of records. (I had to deal with a DB with 99.9% availability and had a very hard time to load that DB with just 5000 records, the closest I got was 4535 successful insertions and then a single failure caused the entire load to fail...).

Repeating the load process won't help because the new load process will repeat the key sequence numbers from 0 - X and simply overwrite existing records instead of adding new ones.

A better solution is to make the load process more robust by retrying a failed insertion (the retry will use the same key).

The code executed during the load phase is CoreWorkload.java:doInsert() (Note: it's NOT the doTransactionInsert()), and currently doInsert() returns false immediately after it sees a failure from _db.insert().

I'm proposing we add some (very) simple retry behavior in doInsert(), we can retry up to N times (N configurable) with randomized wait time between retries (wait time can also be made configurable).

I have already made a private fix for this and it enabled me to load pretty much as many records to my DB as I wanted (up to 100,000) with the retry limit set at just 5. (yes, working with a DB of 99.9% availability, 5 retries will "fix" almost everything for you. :-)

stfeng:
Guys, let me know what your thoughts are, happy to submit a PR for this!

kruthar:
@stfeng I think configurable retry on load is a good idea. A couple of ideas to go along with your initial proposal:

* Configurable number of (subsequent, or maybe not subsequent) failures that would trigger a full stop to loading. Example of the need for this is if there truly is a DB issue, you don't want to have to sit through all the remaining retries before the load would finish.
* Default retry number should probably be 0 to conform to current behavior
* Warnings should be logged to stderr or stdout for each retry, along with reason for failure, to inform the user
* A failed insert should probably get a new Status, something to the effect of "Failed but retrying"

p.s. what DB are you using? I have done successful benchmarks on Couchbase and MongoDB with 100,000,000 records without much issue during loading. Then again there may be retry logic built into some DB drivers.

busbey:
I'm also curious what datastore binding you're using.

A given datastore shouldn't fail the load when a given insert fails, it should just count the number of failed responses.

[...]

PR #527, opened 2015-12-09
[Core] Add retry for insertion logic to make the load process more robust

stfeng:
The default behavior is NO RETRY, which is the current behavior.
    
But optionally user can enable retry if they are working with a DB backend that is better tolerated by allowing a few retries.

This is for issue #508 

Notes for reviewers: The first commit is purely formatting change only, the second commit is the change.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
+++ b/core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
@@ -296,6 +296,16 @@ public class CoreWorkload extends Workload {
    */
   public static final String HOTSPOT_OPN_FRACTION_DEFAULT = "0.8";
 
+  /**
+   * How many times to retry when insertion of a single item to a DB fails.
+   */
+  public static final String INSERTION_RETRY_LIMIT_DEFAULT = "0";
+
+  /**
+   * On average, how long to wait between the retries, in seconds.
+   */
+  public static final String INSERTION_RETRY_INTERVAL_DEFAULT = "3";
+
   IntegerGenerator keysequence;
 
   DiscreteGenerator operationchooser;
@@ -311,6 +321,9 @@ public class CoreWorkload extends Workload {
   boolean orderedinserts;
 
   int recordcount;
+
+  int insertionRetryLimit;
+  int insertionRetryInterval;
 
   private Measurements _measurements = Measurements.getMeasurements();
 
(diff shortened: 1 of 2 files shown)


stfeng:
@busbey, @kruthar, please take a look when you get a chance. Thanks!

stfeng:
Guys, would you have time to take a look at this soon? I will be offline in about 5 calendar days, would like to wrap this up before then if possible. Thanks!

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -1,11 +1,11 @@
 /**
- * Copyright (c) 2010 Yahoo! Inc. All rights reserved. 
- * 
+ * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
```
busbey: note to other reviewers, this copyright header is wrong but the problem is covered by #313.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -525,244 +559,238 @@ private String buildDeterministicValue(String key, String fieldkey) {
...
+   * this function must be thread safe. However, avoid synchronized, or the threads will block waiting
+   * for each other, and it will be difficult to reach the target throughput. Ideally, this function would
+   * have no side effects other than DB operations.
+   */
+  public boolean doInsert(DB db, Object threadstate) {
+    int keynum = keysequence.nextInt();
+    String dbkey = buildKeyName(keynum);
+    HashMap<String, ByteIterator> values = buildValues(dbkey);
+
+    Status status;
```
busbey: If you do this in DBWrapper instead of CoreWorkload, it'll work with arbitrary workloads (not that we have a huge variety ATM, but still).

If you'd prefer to keep this limited to the CoreWorkload, then please update the property to start with a prefix to indicate it's just the core workload (i.e. 'core_workload.').

busbey:
overall looks straight forward. I'll check back again later this evening.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
stfeng: After thinking about this for a bit, I'd prefer to keep the retry in the CoreWorkload class because the need for retry is largely driven by the design of the workload, e.g., a future workload may be able to ignore failed insertions during the load phase (and subsequently avoid touching those records that were not inserted during its run phase). It'd be quite hard to do such for the core workload in its current form, because doing so would require state tracking across runs in the random key number generator. A simpler and more robust way for the core workload is to just retry those failed insertions.

I like the idea of prefixing the retry limit property with "core_workload" so it differentiates from others. I just pushed a change for that. Thanks for the suggestion!

stfeng:
Pushed the change that incorporates CR feedback, rebased, and squashed. 

PTAL. Thanks!

busbey:
Thanks for the contribution!

stfeng:
Happy to contribute! Likewise, thank you for your code reviews, discussions, and contributions to the community here!

PR #528, opened 2015-12-10
[kudu] Bump Kudu dependency to 0.6.0

jdcryans:
This version and the previous one are wire compatible, the only code
chance reflects a naming change in the client API.


busbey:
please start commit message with [kudu]

Issue #532, opened 2015-12-14
[Cassandra] cassandra-binding major refactor


kruthar:
All the Thrift-based clients (all clients except CassandraCQLClient) in the cassandra-binding are essentially the same. They have no major functional differences (except for better property support in CassandraClient10). They are all built against the same dependencies and so to me it seems that they are all redundant.

Dredging back through commit history, it looks like the multiple Cassandra clients used to be built against different versions of the Cassandra driver and so were necessary. #64 brought a major paradigm shift in how the project was built and it seems that that is when they really started to just be copies of each other.

Posting this issue to see if anyone sees any reason why they are different that I may have missed. Otherwise they should be consolidated into a single client. This may help ease confusion between which Cassandra clients do what and are for what versions.

I'd be happy to take this on, just want to make sure I did not miss any reason why they are actually different and necessary as they are now.

busbey:
Related question: do we have anyone willing to test against the earlier versions of Cassandra that need to use (or even _can_ use) the older Thrift-base clients?

If not, should we just deprecate them and remove them in e.g. 6 months?

kruthar:
I had planned on testing against 7, 8, and 10 as the binding suggests. But I agree, deprecating might also make sense, most if not all people are going to be using the cassandra2-binding now.

busbey:
It might be worth asking on the cassandra user list, since they're more likely to have visibility into current prevalence.

kruthar:
I received a few responses from folks on the Cassandra users mailing list: http://www.mail-archive.com/user@cassandra.apache.org/msg45162.html. Looks like they don't expect many people to be using Cassandra versions in 0.X or 1.X.

So, my thoughts are to follow @busbey suggestion and plaster a deprecation warning on the cassandra-binding README, give it a stated amount of time, then remove it. Really no point in spending much effort on fixing that binding.

I would think keeping the cassandra2-binding named as it is, self documenting, and clearly state the versions it is compatible with at the top of the README. The current README suggests it is good for 2.1+, I want to test and see if it is actually good for 2.0 also.

PR #536, opened 2015-12-17
[cassandra2] added better property handling

### LINKED ISSUE #534: [Cassandra] cassandra2-cql fails with threadcount of 1 on Cassandra 2.0
Was testing to see if the ```cassandra2-cql``` would work against Cassandra 2.0 so that after deprecating the cassandra-binding we would have a single client that worked on Cassandra 2.X.

The only issue I found has to do with different connection pool defaults between protocol versions.

You can see here: https://datastax.github.io/java-driver/2.1.8/features/native_protocol/#compatibility-matrix that running the 2.1.8 driver against Cassandra 2.0.X uses protocol V2 whereas Cassandra 2.1.X uses V3.

Currently the maxConnections is being set to threadcount: https://github.com/brianfrankcooper/YCSB/blob/master/cassandra2/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java#L147-L150. With the default threadcount of 1, the V2 default of 2 core connections overruns the max before even starting the test: https://datastax.github.io/java-driver/2.1.8/features/pooling/#pool-size. Resulting in errors similar to:

```
com.yahoo.ycsb.DBException: java.lang.IllegalArgumentException: Core connections for LOCAL hosts must be less than max (2 > 1)
	at com.yahoo.ycsb.db.CassandraCQLClient.init(CassandraCQLClient.java:172)
	at com.yahoo.ycsb.DBWrapper.init(DBWrapper.java:77)
Caused by: java.lang.IllegalArgumentException: Core connections for LOCAL hosts must be less than max (2 > 1)
	at com.datastax.driver.core.PoolingOptions.checkConnectionsPerHostOrder(PoolingOptions.java:521)
	at com.datastax.driver.core.PoolingOptions.setProtocolVersion(PoolingOptions.java:459)
	at com.datastax.driver.core.Cluster$Manager.init(Cluster.java:1450)
	at com.datastax.driver.core.Cluster.getMetadata(Cluster.java:382)
	at com.yahoo.ycsb.db.CassandraCQLClient.init(CassandraCQLClient.java:159)
	... 2 more
```

The number of YCSB threads is not directly linked to the number of connections with Cassandra. The driver manages the connections and autoscales the number of connections to manage the current load. That being said the fix seems simple enough, don't set the maxConnections transparently. Leave the defaults as they are and expose the PoolingOptions connection parameters to the user for fine tuning.

kruthar:
This PR actually fixes two issues, both related to property handling in the ```cassandra2-cql``` client.

1. In the spirit of keeping the clients as baseline as possible, I removed the magic number updates to read and connect timeouts to the cluster. Instead I exposed these properties to the user.
2. No longer setting the max connections to threadcount. As I discussed in #534, YCSB threads are not directly related with Cassandra Connections. I also exposed these settings to the user for fine tuning.

This PR would fix #534 and allows the ```cassandra2-cql``` client to be used on all Cassandra 2.X versions making the deprecation of the cassandra-binding in #532 a little easier.

### DIFF
--- a/cassandra2/README.md
+++ b/cassandra2/README.md
@@ -71,3 +71,10 @@ For keyspace `ycsb`, table `usertable`:
   * Default value is `ONE`
   - Consistency level for reads and writes, respectively. See the [DataStax documentation](http://docs.datastax.com/en/cassandra/2.0/cassandra/dml/dml_config_consistency_c.html) for details.
   * *Note that the default setting does not provide durability in the face of node failure. Changing this setting will affect observed performance.* See also `replication_factor`, above.
+
+* `cassandra.maxconnections`
+* `cassandra.coreconnections`
+  * Defaults for max and core connections can be found here: https://datastax.github.io/java-driver/2.1.8/features/pooling/#pool-size. Cassandra 2.0.X falls under protocol V2, Cassandra 2.1+ falls under protocol V3.
+* `cassandra.connecttimeoutmillis`
+* `cassandra.readtimeoutmillis`
+  * Defaults for connect and read timeouts can be found here: https://docs.datastax.com/en/drivers/java/2.0/com/datastax/driver/core/SocketOptions.html.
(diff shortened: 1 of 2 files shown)


cmccoy:
+1. Thanks, @kruthar!

PR #537, opened 2015-12-17
[cassandra] added large deprecation message

### LINKED ISSUE #532: [Cassandra] cassandra-binding major refactor

kruthar:
I chose 3 month deprecation period, sounds like no one is really using Cassandra 0.X or 1.X anymore. Also, this binding has caused some confusion on which to use for different versions of Cassandra (most Cassandra issues start with 'try using ```cassandra2-cql```), so best to remove it sooner rather than later.

Fixes #532

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -205,6 +205,13 @@ def main():
 
     # Classpath set up
     binding = args.database.split("-")[0]
+
+    # Deprecation message for the entire cassandra-binding
+    if binding == "cassandra":
+        warn("The 'cassandra-7', 'cassandra-8', 'cassandra-10', and "
+             "cassandra-cql' clients are deprecated. If you are using "
+             "Cassandra 2.X try using the 'cassandra2-cql' client instead.")
+
     if is_distribution():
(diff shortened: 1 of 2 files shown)


busbey:
can we add a message to bin/ycsb when the binding is used as well?

kruthar:
added deprecation message to bin/ycsb

busbey:
:+1:

PR #551, opened 2015-12-22
[core] Generator refactoring

vfunshteyn:
This change set refactors generators to achieve better compile-time safety and simpler API, introducing use of generics. There's also some other minor unrelated cleanup.


kruthar:
@vfunshteyn - thanks for the PR.

Can you explain a little more on why these changes are necessary? 

These changes are fairly significant on pieces of core functionality for YCSB. Most of the generators do not have tests written for them. I know that means that we can't really be positive if they are 100% correct right now, but without tests in place we also cannot be sure that this PR does not introduce bugs in the generators.

I'd actually like to see tests written for the generators this PR effects first, so that we can ensure this doesn't break things.

Don't get me wrong, code safety and best practices are good contributions, but at this point it seems difficult to ensure that this is a safe PR.

vfunshteyn:
I would have no problem with this PR getting rejected (and in fact anticipate that), as both outside contributions to OS projects and their acceptance are 100% voluntary. I am perfectly happy keeping them in my fork forever. The changes in this PR are definitely not necessary in the strict sense of the word. But is this project only accepting absolutely necessary changes at this point anyway? 

The reason for this PR is that after I forked the project, I realized there is a lot of room for improvement by refactoring old crufty code that still exists all over the place. Typically, as I work with existing code, I like to clean things up as I go where appropriate. There are lots of examples I can give here, but the generator package appeared to be a relatively well-contained, low hanging fruit in terms of benefiting from use of features that have been standard in the language for quite a few years, so it's certainly a good start to introduce sensible refactoring.

Also, I am not sure how closely you have examined the code, but these changes are completely non-functional, i.e. no existing behavior was altered by them - numeric generators remain as such, and so are string ones. Not positive what you mean by difficulty ensuring safety of this PR but it's not any less safe than the code it supersedes; all you have to do is run your standard workload scenarios and it will be quite obvious if something is broken. Which is basically what I did.

Granted, custom workloads that exist out there and were written to the old generator API may be broken - but that is a compile time failure, and this PR is meant for the dev line anyway, without affecting existing tags/releases.

busbey:
We accept plenty of requests. I think @kruthar is just concerned about our lack of testing around expected behavior. I agree with the concern generally, but I'm hesistent to hold us back while waiting on getting some.

busbey:
The holidays will slow down checking on this request; apologies if this ends up sitting for a bit.

Do you happen to have any kind of perf comparison?

vfunshteyn:
I don't have perf comparisons but I also don't see how this PR would make any difference to performance as it introduces no additional processing logic.

busbey:
The existing code is pretty bad, so a perf change should be easy. ;)

Changing away from Vector, for example. Not sure if we hit any autoboxing overhead before/after the shift to Integer.

Anywho. Thanks for putting up a PR. I'll try to give some feedback before it gets too stale.

busbey:
One minor change, could you update the commit message to start with "[core]" ?

vfunshteyn:
Good point about minor perf improvements - I just pushed another small change that should definitely compare favorable in performance (though it'd likely be hardly noticeable compared to overall DB transaction latency).

kruthar:
@vfunshteyn - please don't get me wrong, I'm not categorically rejecting your PR.

>But is this project only accepting absolutely necessary changes at this point anyway?

Unnecessary changes are as the name implies unnecessary. But I was asking so I could get a feel for what this PR accomplishes, not making any judgments as to necessity at that point. As you and @busbey  pointed out a lot of parts of the code base are bad and so PRs to clean things up are important.

Although there were no functional changes, there were changes, and at this point we have no way to test the generators other than simple workload tests. Even those, unless planned out to do so would not test every generator.

As you can see, most of the generators are 4 years old, it is sort of taken for granted that they work. My two cents are simply that we have a way to test them before changing them. Testing the generators before this PR is the only way to ensure (according to the tests) that things are good before and after. This is as good a coding practice as clean code. But, I'm not the only one with a voice so if they are tested and approved by multiple people then it is what it is and we can move forward.

core/src/main/java/com/yahoo/ycsb/generator/FileGenerator.java
```diff
@@ -36,20 +36,21 @@
...
 	{
+	  filename = _filename;
 		try {
-			filename = _filename;
 			File file = new File(filename);
 			FileInputStream in = new FileInputStream(file);
 			reader = new BufferedReader(new InputStreamReader(in));
 		} catch(IOException e) {
-			System.err.println("Exception: " + e);
+		    throw new RuntimeException(e);
```
risdenk: Are these errors caught somewhere else? Is there a reason that exceptions were just printed instead of being thrown previously?

core/src/main/java/com/yahoo/ycsb/generator/FileGenerator.java
```diff
@@ -57,15 +58,14 @@ public synchronized String nextString()
 			System.err.println("NullPointerException: " + filename + ':' + current);
 			throw e;
 		} catch(IOException e) {
-			System.err.println("Exception: " + e);
-			return null;
-		}
+		  throw new RuntimeException(e);		}
```
risdenk: Are these errors caught somewhere else? Is there a reason that exceptions were just printed instead of being thrown previously?

core/src/main/java/com/yahoo/ycsb/generator/FileGenerator.java
```diff
@@ -73,16 +73,12 @@ public String lastString()
 	/**
 	 * Reopen the file to reuse values.
 	 */
-	public synchronized void reloadFile()
+	public synchronized void reloadFile() throws IOException
```
risdenk: Are these errors caught somewhere else? Is there a reason that exceptions were just printed instead of being thrown previously?

core/src/main/java/com/yahoo/ycsb/generator/Generator.java
```diff
@@ -18,20 +18,28 @@
...
-	public abstract String lastString();
+	public abstract V lastValue();
+	
+	public final String nextString() {
+	  return nextValue().toString();
+	}
+	
+	public final String lastString() {
+	  return lastValue().toString();
+	}
```
risdenk: Can nextvalue() or lastValue() ever be null? will this throw a NPE in that case?

core/src/main/java/com/yahoo/ycsb/generator/CounterGenerator.java
```diff
@@ -32,23 +32,18 @@
...
-	/**
-	 * If the generator returns numeric (integer) values, return the next value as an int. Default is to return -1, which
-	 * is appropriate for generators that do not return numeric values.
-	 */
-	public int nextInt() 
+	@Override
+  public Integer nextValue() 
 	{
-		int ret = counter.getAndIncrement();
-		setLastInt(ret);
```
risdenk: any reason that setLastInt call was removed?

core/src/main/java/com/yahoo/ycsb/generator/CounterGenerator.java
```diff
@@ -32,23 +32,18 @@
 	public CounterGenerator(int countstart)
 	{
 		counter=new AtomicInteger(countstart);
-		setLastInt(counter.get()-1);
```
risdenk: any reason that setLastInt call was removed?

core/src/main/java/com/yahoo/ycsb/generator/ExponentialGenerator.java
```diff
@@ -62,17 +60,6 @@ public ExponentialGenerator(double percentile, double range)
 
 	/****************************************************************************************/
 	
-	/** 
-	 * Generate the next item. this distribution will be skewed toward lower integers; e.g. 0 will
-	 * be the most popular, 1 the next most popular, etc.
-	 * @param itemcount The number of items in the distribution.
-	 * @return The next item in the sequence.
-	 */
```
risdenk: why was this JavaDoc removed? Couldn't it be merged with nextValue() for this class?

core/src/main/java/com/yahoo/ycsb/generator/CounterGenerator.java
```diff
@@ -32,23 +32,18 @@
...
 	}
 	
-	/**
-	 * If the generator returns numeric (integer) values, return the next value as an int. Default is to return -1, which
-	 * is appropriate for generators that do not return numeric values.
-	 */
```
risdenk: Why was this JavaDoc removed could it be merged with nextValue()?

risdenk:
@vfunshteyn - I left some feedback. Overall it looks pretty good I'll probably look it over again in the next week or so. Can you squash your two commits and update the commit message to start with "[core]" like  @busbey said?

core/src/main/java/com/yahoo/ycsb/generator/FileGenerator.java
vfunshteyn: I think the reason was just bad code. It doesn't make sense for this method to complete normally in case of an exception. Same comment for nextValue() too. I'll change it to close the reader as a try with resource statement.

core/src/main/java/com/yahoo/ycsb/generator/CounterGenerator.java
vfunshteyn: `setLastInt()` no longer exists. When it did, calling it was pointless anyway because it was mutating an otherwise unused (in this context) value in base class. Last value is always one behind the current one here.

core/src/main/java/com/yahoo/ycsb/generator/CounterGenerator.java
vfunshteyn: It was redundant, and also incorrect about the default value.

core/src/main/java/com/yahoo/ycsb/generator/CounterGenerator.java
risdenk: right but setLastValue now exists in NumberGenerator and isn't called? If you call getLastString it calls lastValue() which won't be populated by this CounterGenerator since setLastValue is what sets the lastValue?

core/src/main/java/com/yahoo/ycsb/generator/Generator.java
vfunshteyn: Yes, there's an implied contract here; numeric generators in existence will not return a `null` due to boxing of primitives. I will fix the string generators to comply with this assumption though.

core/src/main/java/com/yahoo/ycsb/generator/ExponentialGenerator.java
```diff
@@ -80,9 +67,10 @@ public int nextInt()
 	 * @param itemcount The number of items in the distribution.
```
risdenk: Looks like this param `itemcount` isn't valid anymore (and wasn't valid before?)

core/src/main/java/com/yahoo/ycsb/generator/CounterGenerator.java
vfunshteyn: It doesn't make sense to call `setLastValue` here. This implementation implicitly ties (and did before my changes too) the last value to current one, as opposed to depending on the side effect of calling `setLastValue` (setting a field in base class).

risdenk: ah yea missed that connection. thanks.

core/src/main/java/com/yahoo/ycsb/generator/Generator.java
risdenk: is there a reason to have nextString and lastString at all?

vfunshteyn: Strictly speaking, no. I guess I just wanted to maintain some level of backward compatibility. But we could drop these altogether if that feels cleaner.

core/src/main/java/com/yahoo/ycsb/generator/DiscreteGenerator.java
```diff
@@ -82,22 +81,12 @@ public String nextString()
...
 	 * Return the previous string generated by the distribution; e.g., returned from the last nextString() call. 
 	 * Calling lastString() should not advance the distribution or have any side effects. If nextString() has not yet 
 	 * been called, lastString() should return something reasonable.
 	 */
-	public String lastString()
+	@Override
+    public String lastValue()
 	{
 		if (_lastvalue==null)
 		{
```
risdenk: the line below this should probably be _lastvalue=nextValue();

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -609,20 +610,25 @@ public boolean doInsert(DB db, Object threadstate) {
...
-    } else if (op.compareTo("UPDATE") == 0) {
-      doTransactionUpdate(db);
-    } else if (op.compareTo("INSERT") == 0) {
-      doTransactionInsert(db);
-    } else if (op.compareTo("SCAN") == 0) {
-      doTransactionScan(db);
-    } else {
-      doTransactionReadModifyWrite(db);
-    }
+    switch (operationchooser.nextString()) {
```
risdenk: this can probably be .nextValue() as well.

busbey:
I'd like to merge this. Anyone have concerns still? Could we address them in a follow-up PR?

risdenk:
@busbey I think it looks good. Any minor things could be addressed with followups.

:+1:

busbey:
let's hope for the best on merge then :)

busbey:
thanks for working on this and your patience, @vfunshteyn

PR #554, opened 2015-12-25
Null pointers should not be dereferenced

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S2259 - Null pointers should not be dereferenced
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2259
Kirill Vlasov


busbey:
The currentHM shouldn't be null at that place in the logic. If it is, something is wrong and we should signal so rather than skip doing any work.

How about we put an assert in?

busbey:
Please start commit messages with the area of the code impacted. In this case "[accumulo]"

kirill-vlasov:
The commit was updated.

accumulo/src/main/java/com/yahoo/ycsb/db/AccumuloClient.java
```diff
@@ -276,11 +277,14 @@ public Status scan(String t, String startkey, int recordcount,
         }
         result.add(currentHM);
       }
+
+      Preconditions.checkNotNull(currentHM, "Unexpected scanner behaviour - currentHM is null");
```
busbey: To use this, we'll have to add a dependency on guava to the accumulo binding (because Accumulo might stop including it). Can we switch to a vanilla assert instead?

PR #558, opened 2015-12-26
[mongodb] Update to use BATCHED_OK Status for batched inserts.

### LINKED ISSUE #522: MongoDB batching > 1 not accurate
Hi folks (@allanbank),

I looked at the mongo source and saw there is a batching argument in there. If you set it to > 1 it will write it into a list and only if the batch value is reached will write it to mongo.

After thinking about it, isn't that incorrect? Since YCSB will "measure" the writing into an array list and not account for the actual response time which is when the end of batch is done? If so, I think this should be removed because it skews the results.

What do you think?

allanbank:
This ensures that users do not confuse the results from batched
and non-batched operations.

Closes #522


cmccoy:
+1

PR #559, opened 2015-12-28
Primitives should not be boxed just for "String" conversion

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S2131 - Primitives should not be boxed just for "String" conversion
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2131
Kirill Vlasov

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/generator/ScrambledZipfianGenerator.java
+++ b/core/src/main/java/com/yahoo/ycsb/generator/ScrambledZipfianGenerator.java
@@ -123,7 +123,7 @@ public class ScrambledZipfianGenerator extends IntegerGenerator
 		
 		for (int i=0; i<1000000; i++)
 		{
-			System.out.println(""+gen.nextInt());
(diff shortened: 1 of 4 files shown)


kruthar:
Looks good to me, thanks @kirill-vlasov.

kruthar:
Oh - actually, it looks like you have the wrong commit message?

kirill-vlasov:
Thanks! Updated.

core/src/main/java/com/yahoo/ycsb/generator/ScrambledZipfianGenerator.java
```diff
@@ -123,7 +123,7 @@ public static void main(String[] args)
 		
 		for (int i=0; i<1000000; i++)
 		{
-			System.out.println(""+gen.nextInt());
+			System.out.println(gen.nextInt());
```
busbey: Should this be nextString instead?

PR #560, opened 2015-12-28
Class variable fields should not have public accessibility

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:ClassVariableVisibilityCheck - Class variable fields should not have public accessibility
https://dev.eclipse.org/sonar/coding_rules#q=squid:ClassVariableVisibilityCheck
Kirill Vlasov


kruthar:
Looks good to me, thanks @kirill-vlasov

busbey:
I think this points to a larger problem. None of these things should be assuming the CoreWorkload is the one in use.

Maybe we need to refactor so that init gives the set of tables that will be used?

risdenk:
This was addressed as part of the checkstyle changes to core #895. Thanks for the contribution @kirill-vlasov!

Issue #562, opened 2015-12-30
[orientdb] always scanning to end of table.


kruthar:
While creating test cases for the orientdb-binding I found that the orientdb scan operation is always scanning to the end of the table instead of applying recordcount.

At first glance this looks like an easy fix, just break from looping over the OIndexCursor when we have the correct number of records. PR to follow.

Issue #539, opened 2015-12-17
orientdb README tells folks to clone a third party repo


busbey:
the orientdb README tells folks to download a fork repo:

> git clone git://github.com/nuvolabase/YCSB.git

we should not recommend this. :)

kruthar:
@busbey - can we edit this issue to be a general 'things to fix in the orientdb client' conversation. I have been messing around with it and there are several things that likely need to be updated and fixed. What I have found so far, in no particular order:

1. Update README with better getting started instructions, YCSB repo, etc.
2. Update driver version - current version does not allow remote connections (further on this below)
3. Create tests - OrientDB has an inmemory option so tests should be possible.
4. General code overview - this database isn't in the tested list, so need to give it a look through and see what needs to happen to promote to the tested and verified list.

First bit of discussion: Currently there is a default url set to use a plocal address. I see two issues with this:

1. Should we allow benchmarking of embedded or inmemory databases? What do the benchmark results mean in that case? It doesn't make sense to me.
2. Related, I don't think there should be a default url, in this case it points to in embedded database which seems wrong.

My thought is that all connections need to be remote.

busbey:
sure, fine by me. I only kept it tightly scoped at first to make sure the very wrong bit could be fixed quickly.

In general,  embedded datastores are fine I think. It's totally legit to decide between e.g. using leveldb (or sqlite) and a local single node mongodb deployment in a simple application. I agree, however, that our docs should be geared primarily towards remote when a given datastore has such a mode.

+1 on no default url, folks should have to conscientiously choose to be in local mode if there is a remote mode.

risdenk:
@kruthar How much of this issue is left to address?

kruthar:
@risdenk - everything is taken care of except for updating the README with better getting started instructions. Building from source is actually still required to get all the major changes we just pushed to the orientdb binding. Before the next release we should commit a final set of instructions to the orientdb README with instructions on how to get started with a prebuilt released version.

risdenk:
@kruthar Can this issue be closed?

PR #567, opened 2015-12-31
[orientdb] fixed end table scan issue

kruthar:
This fixes the end table scanning brought up in #562.

While working on this I found some things that should probably be marked as known issues.

1. First is a performance issue, after testing with a few different recordcounts I found that iterateEntriesMajor() return time is directly linked with number of records in the table. So, a relatively small scan size will start to perform worse on larger tables. The better solution would have been to use iterateEntriesBetween(), problem is it takes a start and end key which is not how the current scan framework is set up. It may be a good idea to think about updating the scan operation to pass a recordcount as well as a corresponding endkey. This would benefit more bindings then OrientDB I believe. There are a few things that complicate that idea, but I'll save that for that discussion.
2. The OIndexCursor implementation used here seems to be broken. It does not track the entries properly when iterating so when you get to the end of the iterator you get a ```NullPointerException```. For more info on why, here is the issue I posted: https://github.com/orientechnologies/orientdb/issues/5541. With a sufficiently high recordcount, +10000, this doesn't happen too often. I choose to not try and catch it, you never know if a legitimate null will pop up sometime. False negatives seemed better to me then false positives.

Should known issues go into the orientdb README? anywhere else?

### DIFF
--- a/orientdb/README.md
+++ b/orientdb/README.md
@@ -23,7 +23,7 @@ This section describes how to run YCSB on OrientDB running locally.
 
 Clone the YCSB git repository and compile:
 
-    git clone git://github.com/nuvolabase/YCSB.git
+    git clone https://github.com/brianfrankcooper/YCSB.git
     cd YCSB
     mvn clean package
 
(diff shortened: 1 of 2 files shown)


busbey:
The orientdb readme sounds like the right place.

kruthar:
OK, added issues to README. Subsequently fixed the README issue in #539.

busbey:
+1. We should probably request this get cherry picked to 0.6.0's staging branch.

kruthar:
I'm good either way, I have a couple more PR's coming for orientdb, as well as @allanbank's checkstyle PR which still needs to be merged. Hopefully the orient db can join the tested list by the next release.

PR #569, opened 2016-01-03
[orientdb] added test cases

kruthar:
Added basic operational test cases for the orientdb binding. Should give a little assurance to future cleanup work.

### DIFF
--- a/orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
+++ b/orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
@@ -1,0 +1,236 @@
+/**
+ * Copyright (c) 2015 YCSB contributors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+package com.yahoo.ycsb.db;
+
+import com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx;
+import com.orientechnologies.orient.core.dictionary.ODictionary;
+import com.orientechnologies.orient.core.record.ORecordInternal;
+import com.orientechnologies.orient.core.record.impl.ODocument;
+import com.yahoo.ycsb.ByteIterator;
+import com.yahoo.ycsb.DBException;
+import com.yahoo.ycsb.StringByteIterator;
+
+import org.junit.*;
+import java.util.*;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by kruthar on 12/29/15.
+ */
+public class OrientDBClientTest {
+  // TODO: This must be copied because it is private in OrientDBClient, but this should defer to table property.
+  private static final String CLASS = "usertable";
+  private static final int FIELD_LENGTH = 32;
+  private static final String FIELD_PREFIX = "FIELD";
+  private static final String KEY_PREFIX = "user";
+  private static final int NUM_FIELDS = 3;
+  private static final String TEST_DB_URL = "memory:test";
+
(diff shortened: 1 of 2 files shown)


kruthar:
So, this was a weird failure, seem that whatever location clojars was could not be reached.

I reran the build and things are green.

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
```diff
@@ -0,0 +1,228 @@
+package com.yahoo.ycsb.db;
```
busbey: File needs a license header.

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
```diff
@@ -0,0 +1,228 @@
...
+
+      Properties p = new Properties();
+      // TODO: Extract the property names into final variables in OrientDBClient
+      p.setProperty("orientdb.url", TEST_DB_URL);
+
+      orientDBClient.setProperties(p);
+      orientDBClient.init();
+      orientDBConnection = new ODatabaseDocumentTx(TEST_DB_URL).open("admin","admin");
+      orientDBDictionary = orientDBConnection.getMetadata().getIndexManager().getDictionary();
+    } catch (DBException e) {
```
busbey: If you just let this exception propogate out of the method, junit will take care of logging the details.

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
```diff
@@ -0,0 +1,228 @@
...
+    if (orientDBConnection != null) {
+      orientDBConnection.close();
+    }
+
+    try {
+      if (orientDBClient != null) {
+        orientDBClient.cleanup();
+      }
+    } catch (DBException e) {
+      e.printStackTrace();
```
busbey: Same comment as on the setup method re: letting the exception propogate.

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
kruthar: Indeed it does. Added.

kruthar:
@busbey - thanks for the tip, exceptions now propogating

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+
+  /*
+      This is a copy of buildDeterministicValue() from core:com.yahoo.ycsb.workloads.CoreWorkload.java.
+      That method is neither public nor static so we need a copy.
+   */
+  private String buildDeterministicValue(String key, String fieldkey) {
```
busbey: Any thoughts on this copying vs using something like powermock to access the method vs making the original visible for testing?

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
```diff
@@ -0,0 +1,236 @@
+/**
+ * Copyright (c) 2015 YCSB contributors. All rights reserved.
```
busbey: 2016?

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
busbey: Thinking about this more, I agree that copying makes the most sense right now. It probably points to a need to come up with a common strategy for these kinds of tests, but that's not an issue for this PR.

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
kruthar: I actually wrote this in 2015, but the PR came in 2016, does one matter over the other?

busbey: Wrote in 2015 means 2015 is correct. Just wanted to double check.

busbey:
+1

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
kruthar: Agreed on your last point, needs deeper thought on consolidated testing strategy in the future.

kruthar:
Thanks @busbey

PR #570, opened 2016-01-04
[accumulo] Accumulo binding does not work in 0.5.0 release

busbey:
YCSB with Accumulo doesn't work in 0.5.0. 

The datastore binding caches a connection to the Accumulo client and handles set up only when the table being interacted with changes. The changes from #464 accidentally used the cached table name rather than the newly proposed table name.

On the first operation, the cached table name is blank which causes the Accumulo driver to error out.


busbey:
with this update in place I can perform a trivial insert of a few records. @cmccoy mind if I update the 0.5.0 release notes to call out this as a known issue?

cmccoy:
@busbey - sure, go ahead!

busbey:
okay, I updated the 0.5.0 release notes.

if someone has a chance for a quick review, I'd like to get this in quick so we can get one more 0.6.0 RC. /cc @bigbes

cmccoy:
+1

Issue #571, opened 2016-01-04
[orientdb] fails with multiple threads when database does not already exists


kruthar:
I tested with:

```bin/ycsb load orientdb -P workloads/workloada -p orientdb.url=plocal:ycsb -threads 2```

my database did not already exists and I recieived this:

```
OrientDB database not found, create fresh db
OrientDB database not found, create fresh db
Could not initialize OrientDB connection pool for Loader: com.orientechnologies.orient.core.exception.ODatabaseException: Cannot create database 'ycsb'
com.orientechnologies.orient.core.exception.ODatabaseException: Cannot create database 'ycsb'
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.create(ODatabaseDocumentTx.java:452)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.create(ODatabaseDocumentTx.java:361)
	at com.yahoo.ycsb.db.OrientDBClient.init(OrientDBClient.java:98)
	at com.yahoo.ycsb.DBWrapper.init(DBWrapper.java:77)
	at com.yahoo.ycsb.ClientThread.run(Client.java:289)
Caused by: com.orientechnologies.orient.core.exception.OStorageException: Cannot create new storage 'plocal:ycsb' because it is not closed
	at com.orientechnologies.orient.core.storage.impl.local.OAbstractPaginatedStorage.create(OAbstractPaginatedStorage.java:263)
	at com.orientechnologies.orient.core.storage.impl.local.paginated.OLocalPaginatedStorage.create(OLocalPaginatedStorage.java:113)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.create(ODatabaseDocumentTx.java:383)
	... 4 more
DBWrapper: report latency for each error is false and specific error codes to track for latency are: []
com.orientechnologies.orient.core.exception.ODatabaseException: Database 'plocal:ycsb' is closed
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.checkOpeness(ODatabaseDocumentTx.java:2930)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.getMetadata(ODatabaseDocumentTx.java:793)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.getMetadata(ODatabaseDocumentTx.java:121)
	at com.orientechnologies.orient.core.record.impl.ODocument.setClassName(ODocument.java:1958)
[...]

PR #572, opened 2016-01-04
[orientdb] version bump to 2.1.8

kruthar:
Bumping the version required a few support fixes to make things work:

1. There were some OrientDB API changes around the use of ```ORecordInternal```, not sure why that was being used in the first place, changed to ```ORecord```.
2. Made ```db``` reference protected so that it could be used by the test class. The way that Maven was running the tests (with multiple threads) was hitting a lot of issues with OrientDB thread safety. This was the easiest way around that.
3. The way the test class calls the clients ```cleanup()``` method required a oneliner to deal with the threads.
4. Part of the known issue about the broken iterator is that the iterator starts at index 1 instead of 0. I added a temporary fix in the test class so that the test could still pass.

### DIFF
--- a/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
+++ b/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2012 - 2015 YCSB contributors. All rights reserved.
+ * Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You
--- a/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
+++ b/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
@@ -25,12 +25,13 @@
 package com.yahoo.ycsb.db;
 
 import com.orientechnologies.orient.core.config.OGlobalConfiguration;
+import com.orientechnologies.orient.core.db.ODatabaseRecordThreadLocal;
 import com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx;
 import com.orientechnologies.orient.core.db.record.OIdentifiable;
 import com.orientechnologies.orient.core.dictionary.ODictionary;
 import com.orientechnologies.orient.core.index.OIndexCursor;
 import com.orientechnologies.orient.core.intent.OIntentMassiveInsert;
-import com.orientechnologies.orient.core.record.ORecordInternal;
+import com.orientechnologies.orient.core.record.ORecord;
 import com.orientechnologies.orient.core.record.impl.ODocument;
 import com.yahoo.ycsb.ByteIterator;
 import com.yahoo.ycsb.DB;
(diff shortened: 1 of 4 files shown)


orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
```diff
@@ -42,10 +41,9 @@
   private static final int NUM_FIELDS = 3;
   private static final String TEST_DB_URL = "memory:test";
 
-  private static ODatabaseDocumentTx orientDBConnection = null;
-  private static ODictionary<ORecordInternal<?>> orientDBDictionary;
+  private static ODictionary<ORecord> orientDBDictionary;
   private static OrientDBClient orientDBClient = null;
-
+
```
busbey: nit: extra whitespace.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -115,6 +116,9 @@ public void init() throws DBException {
 
   @Override
   public void cleanup() throws DBException {
+    // Set this thread's db reference (needed for thread safety in testing)
+    ODatabaseRecordThreadLocal.INSTANCE.set(db);
```
busbey: any particular reason this happens in cleanup, rather than e.g. setup?

orientdb/src/test/java/com/yahoo/ycsb/db/OrientDBClientTest.java
```diff
@@ -223,8 +216,15 @@ public void scanTest() {
 
     // Check the resultVector is the correct size
     assertEquals("Assert the correct number of results rows were returned", resultRows, resultVector.size());
+
+    /**
+     * Part of the known issue about the broken iterator in orientdb is that the iterator
```
busbey: great comment! Can we include a link to the relevant orientdb issue?

kruthar: Added a link to the issue, also noted the dependency version this is for.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: Yes, this was a weird. I was not able to nail down exactly what is happening. When you run the tests, I am creating a OrientDBClient instance by hand and calling its ```init()```, ```cleanup()```, and operational methods manually. Two threads are used, what I don't understand is exactly what is happening on each. I can say that ```cleanup()``` seems to be running on a different thread then ```init()```, using ```ODatabaseRecordThreadLocal.INSTANCE.set(db);``` in ```init()``` did not fix the problem. Only from ```cleanup()``` did it fix the problem.

kruthar:
White space removed :smile:

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
busbey: we have #571 to cover correcting concurrency in the orientdb binding, so sounds fine.

busbey:
+1. does this update driver work with the older orientdb version(s)? just want to make sure we release note things correctly.

kruthar:
So, currently ```orientdb``` only supports memory and plocal database types. Both of which don't actually require an installed database, they are locally created via the OrientDB Java driver. I have a PR coming to allow remote connections as well, will be sure to note supported versions with that.

PR #574, opened 2016-01-06
Adding Apache Solr Support

ghaughian:
Since ElasticSearch is supported I thought it would be beneficial to have Solr supported also. I'm sure there are people out there who would love to compare head-to-head these 2 search engine technologies.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -78,6 +78,7 @@ DATABASES = {
     "orientdb"     : "com.yahoo.ycsb.db.OrientDBClient",
     "redis"        : "com.yahoo.ycsb.db.RedisClient",
     "s3"           : "com.yahoo.ycsb.db.S3Client",
+    "solr"         : "com.yahoo.ycsb.db.SolrClient",
     "tarantool"    : "com.yahoo.ycsb.db.TarantoolClient",
     "voldemort"    : "com.yahoo.ycsb.db.VoldemortClient"
 }
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -141,6 +141,11 @@ LICENSE file.
     </dependency>
     <dependency>
       <groupId>com.yahoo.ycsb</groupId>
+      <artifactId>solr-binding</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.yahoo.ycsb</groupId>
       <artifactId>tarantool-binding</artifactId>
       <version>${project.version}</version>
     </dependency>
--- a/solr/README.md
+++ b/solr/README.md
@@ -1,0 +1,110 @@
+<!--
+Copyright (c) 2016 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+## Quick Start
+
+This section describes how to run YCSB on Solr running locally. 
+
+### 1. Set Up YCSB
+
+Clone the YCSB git repository and compile:
+
+    git clone git://github.com/brianfrankcooper/YCSB.git
+    cd YCSB
+    mvn -pl com.yahoo.ycsb:solr-binding -am clean package
+
+### 2. Set Up Solr
+
+There must be a running Solr instance with a core/collection pre-defined and configured. 
+- See this [API](https://cwiki.apache.org/confluence/display/solr/CoreAdmin+API#CoreAdminAPI-CREATE) reference on how to create a core.
+- See this [API](https://cwiki.apache.org/confluence/display/solr/Collections+API#CollectionsAPI-api1) reference on how to create a collection in SolrCloud mode.
+
+The `conf/schema.xml` configuration file present in the core/collection just created must be configured to handle the expected field names during benchmarking.
+Below illustrates a sample from a schema config file that matches the default field names used by the ycsb client:
+
+	<field name="id" type="string" indexed="true" stored="true" required="true" multiValued="false"/>
+	<field name="field0" type="text_general" indexed="true" stored="true"/>
+	<field name="field1" type="text_general" indexed="true" stored="true"/>
+	<field name="field2" type="text_general" indexed="true" stored="true"/>
+	<field name="field3" type="text_general" indexed="true" stored="true"/>
+	<field name="field4" type="text_general" indexed="true" stored="true"/>
+	<field name="field5" type="text_general" indexed="true" stored="true"/>
+	<field name="field6" type="text_general" indexed="true" stored="true"/>
+	<field name="field7" type="text_general" indexed="true" stored="true"/>
+	<field name="field8" type="text_general" indexed="true" stored="true"/>
+	<field name="field9" type="text_general" indexed="true" stored="true"/>
+
+If running in SolrCloud mode ensure there is an external Zookeeper cluster running.
+- See [here](https://cwiki.apache.org/confluence/display/solr/Setting+Up+an+External+ZooKeeper+Ensemble) for details on how to set up an external Zookeeper cluster.
+- See [here](https://cwiki.apache.org/confluence/display/solr/Using+ZooKeeper+to+Manage+Configuration+Files) for instructions on how to use Zookeeper to manage your core/collection configuration files.
+
+### 3. Run YCSB
+    
+Now you are ready to run! First, load the data:
+
+    ./bin/ycsb load solr -s -P workloads/workloada -p table=<core/collection name>
+
+Then, run the workload:
+
+    ./bin/ycsb run solr -s -P workloads/workloada -p table=<core/collection name>
+
+For further configuration see below: 
+
+### Default Configuration Parameters
+The default settings for the Solr node that is created is as follows:
+	
+- `solr.cloud` 
+  - A Boolean value indicating if Solr is running in SolrCloud mode. If so there must be an external Zookeeper cluster running also.
+  - Default value is `false` and therefore expects solr to be running in stand-alone mode.
+
+- `solr.base.url` 
+  - The base URL in which to interface with a running Solr instance in stand-alone mode
+  - Default value is `http://localhost:8983/solr
+
+- `solr.commit.within.time`
+  - The max time in ms to wait for a commit when in batch mode, ignored otherwise
+  - Default value is `1000ms`
+
+- `solr.batch.mode`
+  - Indicates if inserts/updates/deletes should be commited in batches (frequency controlled by the `solr.commit.within.time` parameter) or commit 1 document at a time.
+  - Default value is `false`
+
+- `solr.zookeeper.hosts`
+  - A list of comma seperated host:port pairs of Zookeeper nodes used to manage SolrCloud configurations.
+  - Must be passed when in [SolrCloud](https://cwiki.apache.org/confluence/display/solr/SolrCloud) mode.
+  - Default value is `localhost:2181`
+
+### Custom Configuration
+If you wish to customize the settings used to create the Solr node
+you can created a new property file that contains your desired Solr 
+node settings and pass it in via the parameter to 'bin/ycsb' script. Note that 
+the default properties will be kept if you don't explicitly overwrite them.
+
+Assuming that we have a properties file named "myproperties.data" that contains 
+custom Solr node configuration you can execute the following to
+pass it into the Solr client:
+
+    ./bin/ycsb run solr -P workloads/workloada -P myproperties.data -s
+
+If you wish to use SolrCloud mode ensure a Solr cluster is running with an
+external zookeeper cluster and an appropriate collection has been created.
+Make sure to pass the following properties as parameters to 'bin/ycsb' script.
+
+	solr.cloud=true
+	solr.zookeeper.hosts=<zkHost2>:<zkPort1>,...,<zkHostN>:<zkPortN>
+
+
--- a/solr/pom.xml
+++ b/solr/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+Copyright (c) 2012 - 2016 YCSB contributors. 
+All rights reserved. 
+
+Licensed under the Apache License, Version 2.0 (the "License"); you 
+may not use this file except in compliance with the License. You 
+may obtain a copy of the License at 
+	
+http://www.apache.org/licenses/LICENSE-2.0 
+
+Unless required by applicable law or agreed to in writing, software 
+distributed under the License is distributed on an "AS IS" BASIS, 
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or 
+implied. See the License for the specific language governing 
+permissions and limitations under the License. See accompanying 
+LICENSE file. 
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>com.yahoo.ycsb</groupId>
+		<artifactId>binding-parent</artifactId>
+		<version>0.7.0-SNAPSHOT</version>
+		<relativePath>../binding-parent</relativePath>
+	</parent>
+
+	<artifactId>solr-binding</artifactId>
+	<name>Solr Binding</name>
+	<packaging>jar</packaging>
+
+	<dependencies>
+		<dependency>
+			<groupId>com.yahoo.ycsb</groupId>
+			<artifactId>core</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		
+		</dependency>
+		<dependency>
+			<groupId>org.apache.solr</groupId>
+			<artifactId>solr-solrj</artifactId>
+			<version>${solr.version}</version>
+		</dependency>
+		
+	</dependencies>
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-checkstyle-plugin</artifactId>
+				<version>2.15</version>
+				<configuration>
+					<consoleOutput>true</consoleOutput>
+					<configLocation>../checkstyle.xml</configLocation>
+					<failOnViolation>true</failOnViolation>
+					<failsOnError>true</failsOnError>
+				</configuration>
+				<executions>
+					<execution>
+						<id>validate</id>
+						<phase>validate</phase>
+						<goals>
+							<goal>checkstyle</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+</project>
--- a/solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
+++ b/solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
@@ -1,0 +1,327 @@
+/**
+ * Copyright (c) 2016 YCSB contributors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+package com.yahoo.ycsb.db;
+
+import com.yahoo.ycsb.ByteIterator;
+import com.yahoo.ycsb.DB;
+import com.yahoo.ycsb.DBException;
+import com.yahoo.ycsb.Status;
+import com.yahoo.ycsb.StringByteIterator;
+
+import org.apache.solr.client.solrj.SolrQuery;
+import org.apache.solr.client.solrj.impl.CloudSolrClient;
+import org.apache.solr.client.solrj.impl.HttpSolrClient;
+import org.apache.solr.client.solrj.response.QueryResponse;
+import org.apache.solr.client.solrj.response.UpdateResponse;
+import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.common.SolrDocument;
+import org.apache.solr.common.SolrDocumentList;
+import org.apache.solr.common.SolrInputDocument;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map.Entry;
+import java.util.Properties;
+import java.util.Set;
+import java.util.Vector;
+
+/**
+ * Solr client for YCSB framework.
+ *
+ * <p>
+ * Default properties to set:
+ * </p>
+ * <ul>
+ * See README.md
+ * </ul>
+ *
+ */
+public class SolrClient extends DB {
+
+  public static final String DEFAULT_CLOUD_MODE = "false";
+  public static final String DEFAULT_BATCH_MODE = "false";
+  public static final String DEFAULT_ZOOKEEPER_HOSTS = "localhost:2181";
+  public static final String DEFAULT_SOLR_BASE_URL = "http://localhost:8983/solr";
+  public static final String DEFAULT_COMMIT_WITHIN_TIME = "1000";
+
+  private org.apache.solr.client.solrj.SolrClient client;
+  private Integer commitTime;
+  private Boolean batchMode;
+
+  /**
+   * Initialize any state for this DB. Called once per DB instance; there is one DB instance per
+   * client thread.
+   */
+  @Override
+  public void init() throws DBException {
+    Properties props = getProperties();
+    commitTime = Integer
+        .parseInt(props.getProperty("solr.commit.within.time", DEFAULT_COMMIT_WITHIN_TIME));
+    batchMode = Boolean.parseBoolean(props.getProperty("solr.batch.mode", DEFAULT_BATCH_MODE));
+
+    // Check if Solr cluster is running in SolrCloud or Stand-alone mode
+    Boolean cloudMode = Boolean.parseBoolean(props.getProperty("solr.cloud", DEFAULT_CLOUD_MODE));
+    System.err.println("Solr Cloud Mode = " + cloudMode);
+    if (cloudMode) {
+      System.err.println("Solr Zookeeper Remote Hosts = "
+          + props.getProperty("solr.zookeeper.hosts", DEFAULT_ZOOKEEPER_HOSTS));
+      client = new CloudSolrClient(
+          props.getProperty("solr.zookeeper.hosts", DEFAULT_ZOOKEEPER_HOSTS));
+    } else {
+      client = new HttpSolrClient(props.getProperty("solr.base.url", DEFAULT_SOLR_BASE_URL));
+    }
+  }
+
+  @Override
+  public void cleanup() throws DBException {
+    try {
+      client.close();
+    } catch (IOException e) {
+      throw new DBException(e);
+    }
+  }
+
+  /**
+   * Insert a record in the database. Any field/value pairs in the specified values HashMap will be
+   * written into the record with the specified record key.
+   *
+   * @param table
+   *          The name of the table
+   * @param key
+   *          The record key of the record to insert.
+   * @param values
+   *          A HashMap of field/value pairs to insert in the record
+   * @return Zero on success, a non-zero error code on error. See this class's description for a
+   *         discussion of error codes.
+   */
+  @Override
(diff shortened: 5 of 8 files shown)


solr/README.md
```diff
@@ -0,0 +1,77 @@
...
+custom Solr node configuration you can execute the following to
+pass it into the Solr client:
+
+    ./bin/ycsb run solr -P workloads/workloada -P myproperties.data -s
+
+If you wish to use SolrCloud mode ensure a Solr cluster is running with an
+external zookeeper cluster and an appropriate collection has been created.
+Make sure to pass the following properties as parameters to 'bin/ycsb' script.
+
+	cloud.mode=true
```
kruthar: This seems to clash with above, is the property ```cloud.mode``` or ```solr.cloud```, The latter is prefered as it refers to the binding it belongs to in the name.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+   *          The name of the table
+   * @param key
+   *          The record key of the record to delete.
+   * @return Zero on success, a non-zero error code on error. See this class's description for a
+   *         discussion of error codes.
+   */
+  @Override
+  public Status delete(String table, String key) {
+    try {
+      UpdateResponse res = (cloudMode) ? cloudClient.deleteById(table, key, 1000)
```
kruthar: Since the maxcommittime is a performance attribute, can you pull it into a variable and expose it to the user as a workload property, with the default being what you are using here?

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+          : httpClient.deleteById(table, key, 1000);
+      if (res.getStatus() > 0) {
+        throw new Exception("");
```
kruthar: There might be a better way to dealing with bad states here. If res.getStatus() gives back any useful information, you could return different status codes that have more meaning than Status.ERROR. There are some predefined status codes here: https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/Status.java, you can also create more specific ones if necessary.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+ *
+ */
+public class SolrClient extends DB {
+
+  public static final String DEFAULT_COLLECTION_NAME = "ycsb";
+  public static final String DEFAULT_CLOUD_MODE = "false";
+  public static final String DEFAULT_REMOTE_ZOOKEEPER_HOST = "localhost:2181";
+  public static final String DEFAULT_REMOTE_SOLR_HOST = "localhost:8983";
+  private CloudSolrClient cloudClient;
+  private HttpSolrClient httpClient;
```
kruthar: HttpSolrClient and CloudSolrClient both extend SolrClient. Can you just store a SolrClient instance, instantiate it with either HttpSolrClient or CloudSolrClient? This way you can remove all the ```if cloudMode do this else that lines``` and do the action regardless on the SolrClient instance.

If there are places you need to distinguish you can cast.

solr/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+			<groupId>org.apache.solr</groupId>
+			<artifactId>solr-solrj</artifactId>
+			<version>${solr-version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.solr</groupId>
+			<artifactId>solr-core</artifactId>
+			<version>${solr-version}</version>
+		</dependency>
+		<dependency>
```
kruthar: Is this dependency used? Were there supposed to be tests added?

solr/README.md
```diff
@@ -0,0 +1,77 @@
+<!--
+Copyright (c) 2012 YCSB contributors. All rights reserved.
```
kruthar: This should probably be 2016?

solr/pom.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2012 - 2015 YCSB contributors. All rights reserved. Licensed
```
kruthar: Update to 2016?

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
+/**
+ * Copyright (c) 2012 YCSB contributors. All rights reserved.
```
kruthar: Update 2016?

solr/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
```
kruthar: Update 2016?

kruthar:
Can you update all the license headers to include 2016?

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+      SolrInputDocument doc = new SolrInputDocument();
+
+      doc.addField("id", key);
+      for (Entry<String, String> entry : StringByteIterator.getStringMap(values).entrySet()) {
+        doc.addField(entry.getKey(), entry.getValue());
+      }
+
+      if (cloudMode) {
+        cloudClient.add(table, doc);
+        cloudClient.commit(table);
```
risdenk: Explicit commits for each record will result in bad performance. Typically commits should be done within a certain amount of time or have the Solr server do auto soft/hard commits. This is different than a standard DB. It should be made configurable?

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+  @Override
+  public Status delete(String table, String key) {
+    try {
+      UpdateResponse res = (cloudMode) ? cloudClient.deleteById(table, key, 1000)
+          : httpClient.deleteById(table, key, 1000);
+      if (res.getStatus() > 0) {
+        throw new Exception("");
+      }
+      return Status.OK;
+    } catch (Exception e) {
```
risdenk: This catches all exceptions?

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+      if (cloudMode) {
+        cloudClient.add(table, doc);
+        cloudClient.commit(table);
+      } else {
+        httpClient.add(table, doc);
+        httpClient.commit(table);
+      }
+
+      return Status.OK;
+    } catch (Exception e) {
```
risdenk: should be more explicit about what exceptions are being caught

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+
+      if ((response.getResults() != null) && (response.getResults().getNumFound() > 0)) {
+        for (String field : response.getResults().get(0).getFieldNames()) {
+          result.put(field, new StringByteIterator(
+              String.valueOf(response.getResults().get(0).getFirstValue(field))));
+        }
+        return Status.OK;
+      }
+      return Status.NOT_FOUND;
+    } catch (Exception e) {
```
risdenk: avoid catching all exceptions be more explicit

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+          cloudClient.add(table, ClientUtils.toSolrInputDocument(updatedDoc));
+          cloudClient.commit(table);
+        } else {
+          httpClient.add(table, ClientUtils.toSolrInputDocument(updatedDoc));
+          httpClient.commit(table);
+        }
+        return Status.OK;
+      }
+      return Status.NOT_FOUND;
+    } catch (Exception e) {
```
risdenk: avoid catching all exceptions be more explicit

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
busbey: the default in ycsb, to allow comparisons across systems, is to do a commit per record. you are correct that this is terrible for most systems.

other bindings make it configurable (hbase, cassandra, mongo, etc) and then just default to commit-per-transaction with instructions in their README for folks who can allow for client side batching.

solr/README.md
```diff
@@ -0,0 +1,77 @@
...
+Clone the YCSB git repository and compile:
+
+    git clone git://github.com/brianfrankcooper/YCSB.git
+    cd YCSB
+    mvn clean package
+
+### 2. Set Up Solr
+
+There must be a running Solr instance with a core/collection already
+defined with an appropriate schema.xml file.
```
madrob: What is the appropriate schema.xml file for the ycsb test? Is this based on the workloads? It is unclear what kinds of types/fields I need defined for the test to run.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,316 @@
...
+      if ((response.getResults() != null) && (response.getResults().getNumFound() > 0)) {
+        final SolrDocument updatedDoc = response.getResults().get(0);
+
+        for (Entry<String, String> entry : StringByteIterator.getStringMap(values).entrySet()) {
+          updatedDoc.setField(entry.getKey(), entry.getValue());
+        }
+
+        if (cloudMode) {
+          cloudClient.add(table, ClientUtils.toSolrInputDocument(updatedDoc));
+          cloudClient.commit(table);
```
madrob: Same comment about explicit commits in `insert` applies here.

ghaughian:
Addressed all of your comments. Some very helpful stuff and a few clear oversights on my part. Cheers!

bin/ycsb
```diff
@@ -78,6 +78,7 @@ DATABASES = {
     "orientdb"     : "com.yahoo.ycsb.db.OrientDBClient",
     "redis"        : "com.yahoo.ycsb.db.RedisClient",
     "s3"           : "com.yahoo.ycsb.db.S3Client",
+    "solr"         : "com.yahoo.ycsb.db.SolrDBClient",
```
risdenk: Nit but should this be SolrClient? Solr to me isn't a DB per say and to stick with existing convention of NameClient.

solr/src/main/conf/schema.xml
```diff
@@ -0,0 +1,613 @@
...
+      Unless this field is marked with required="false", it will be a required field
+   -->
+ <uniqueKey>id</uniqueKey>
+
+
+  <!-- copyField commands copy one field to another at the time a document
+        is added to the index.  It's used either to index the same field differently,
+        or to add multiple fields to the same field for easier/faster searching.  -->
+
+   <copyField source="field0" dest="text"/>
```
risdenk: Are these copyFields necessary?

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+  private Boolean batchMode;
+
+  /**
+   * Initialize any state for this DB. Called once per DB instance; there is one DB instance per
+   * client thread.
+   */
+  @Override
+  public void init() throws DBException {
+    Properties props = getProperties();
+    // String collectionName = props.getProperty("table", DEFAULT_COLLECTION_NAME);
```
risdenk: Dead code?

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+ * <li>solr.batch.mode
+ * <li>zookeeper.host.list
+ * </ul>
+ *
+ * @author Gerard Haughian
+ *
+ */
+public class SolrDBClient extends DB {
+
+  public static final String DEFAULT_COLLECTION_NAME = "ycsb";
```
risdenk: Is this still used?

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+    Boolean cloudMode = Boolean.parseBoolean(props.getProperty("solr.cloud", DEFAULT_CLOUD_MODE));
+    System.out.println("Solr Cloud Mode = " + cloudMode);
+    if (cloudMode) {
+      System.out.println("Solr Zookeeper Remote Hosts = "
+          + props.getProperty("zookeeper.host.list", DEFAULT_REMOTE_ZOOKEEPER_HOST));
+      client = new CloudSolrClient(
+          props.getProperty("zookeeper.host.list", DEFAULT_REMOTE_ZOOKEEPER_HOST));
+    } else { 
+      client = new HttpSolrClient(
+          "http://" + props.getProperty("solr.host.list", DEFAULT_REMOTE_SOLR_HOST) + "/solr");
```
risdenk: Might make sense to have the property be the solr base url which includes the http:// and /solr instead of appending to make the base url. These could conceivably (unlikely though) be different for clusters (ie: https).

bin/ycsb
busbey: all of the other bindings just use the name of the data storage layer. e.g. elastic search.

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+   *          The name of the table
+   * @param key
+   *          The record key of the record to delete.
+   * @return Zero on success, a non-zero error code on error. See this class's description for a
+   *         discussion of error codes.
+   */
+  @Override
+  public Status delete(String table, String key) {
+    try {
+      UpdateResponse response = client.deleteById(table, key, commitTime);
```
risdenk: commitWithin here but not an if statement check and explicit commit like other places?

solr/pom.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved. Licensed
```
busbey: formatting is bad here. please try to match the other binding pom files wrt where new lines go.

solr/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+		<artifactId>binding-parent</artifactId>
+		<version>0.7.0-SNAPSHOT</version>
+		<relativePath>../binding-parent</relativePath>
+	</parent>
+
+	<artifactId>solr-binding</artifactId>
+	<name>Solr Binding</name>
+	<packaging>jar</packaging>
+	<properties>
+		<solr-version>5.4.0</solr-version>
```
busbey: this already got set in the top level.

bin/ycsb
ghaughian: SolrClient conflicts with the name of an object in the org.apache.solr library that's used as an abstraction through which all communication with a Solr server is routed. I needed to rename it to something.

solr/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2016, Yahoo!, Inc. All rights reserved.
```
busbey: YCSB contributors, I think.

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+ * <ul>
+ * <li>table = ycsb
+ * <li>solr.cloud = false
+ * <li>solr.host.list
+ * <li>solr.commit.within.time
+ * <li>solr.batch.mode
+ * <li>zookeeper.host.list
+ * </ul>
+ *
+ * @author Gerard Haughian
```
busbey: no author tags please. we rely on git for attribution.

solr/src/main/conf/schema.xml
```diff
@@ -0,0 +1,613 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
```
busbey: is this copied verbatim from somewhere? if not, it should follow the header format we use rather than that for contributing to an ASF project.

Also, should this be in src/main/resources? Where do we expect it to end up in the solr-specific convenience binary tarball?

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+import java.util.Map.Entry;
+import java.util.Properties;
+import java.util.Set;
+import java.util.Vector;
+
+/**
+ * Solr client for YCSB framework.
+ *
+ * <p>
+ * Default properties to set:
```
busbey: if these all get covered in the README, just refer folks there.

solr/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Defaults Configuration
+The default settings for the Solr node that is created is as follows:
+	
+	'table'/core/collection name = 'ycsb'		// standard ycsb property
+	'solr.cloud' mode = 'false'					// expects solr to be running in stand-alone mode
+	'solr.host.list' = 'localhost:8983'			// assumes Solr is running locally on the standard port
+	'solr.commit.within.time' = '1000'			// max time in ms to wait for a commit when in batch mode
+	'solr.batch.mode' = 'false'					// commits 1 document at a time (when set to false)
+	'zookeeper.host.list' = 'localhost:2181'	// used when in SolrCloud mode only
```
busbey: can we confirm that the zookeeper parameter matches other bindings that require including the same info? Accumulo, I think?

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+   *          The name of the table
+   * @param key
+   *          The record key of the record to write.
+   * @param values
+   *          A HashMap of field/value pairs to update in the record
+   * @return Zero on success, a non-zero error code on error. See this class's description for a
+   *         discussion of error codes.
+   */
+  @Override
+  public Status update(String table, String key, HashMap<String, ByteIterator> values) {
```
risdenk: A thought here: Solr supports updating documents without doing a query/update manually. https://cwiki.apache.org/confluence/display/solr/Updating+Parts+of+Documents and here how to do it with SolrJ http://yonik.com/solr/atomic-updates/

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+public class SolrDBClient extends DB {
+
+  public static final String DEFAULT_COLLECTION_NAME = "ycsb";
+  public static final String DEFAULT_CLOUD_MODE = "false";
+  public static final String DEFAULT_BATCH_MODE = "false";
+  public static final String DEFAULT_REMOTE_ZOOKEEPER_HOST = "localhost:2181";
+  public static final String DEFAULT_REMOTE_SOLR_HOST = "localhost:8983";
+  public static final String DEFAULT_COMMIT_WITHIN_TIME = "1000";
+
+  private SolrClient client;
```
busbey: if this is the only place where the solr object is used, you could instead refer to it by the full qualified package name here and then you wouldn't have to worry about your class name colliding.

busbey:
It would be helpful if, before this gets merged, you could squash all the commits into a single commit and then include in the commit message "[solr]" at the beginning.

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+  public Status scan(String table, String startkey, int recordcount, Set<String> fields,
+      Vector<HashMap<String, ByteIterator>> result) {
+    try {
+      Boolean returnFields = false;
+      String[] fieldList = null;
+      if (fields != null) {
+        returnFields = true;
+        fieldList = fields.toArray(new String[fields.size()]);
+      }
+      SolrQuery query = new SolrQuery();
```
risdenk: Does this query on a scan need to be ordered by the id field? Does it matter if the scan doesn't come back ordered? More a question of convention for YCSB implementation of scans.

risdenk: NVM doesn't look like Elasticsearch does ordering here either.

solr/src/main/conf/schema.xml
ghaughian: It is copied from a sample Apache Solr schema.xml file.

It's not actually required for the tarball at all; it's only there for users to make use and refer to as they configure their Solr servers separate from ycsb. So, would that make src/main/resources the best place for it then?

solr/README.md
ghaughian: In Accumulo the parameter is called "accumulo.zooKeepers" and is the exact same format as that required here.

solr/src/main/conf/schema.xml
busbey: can we point users to Apache Solr then?

distribution/pom.xml
```diff
@@ -139,6 +139,11 @@ LICENSE file.
       <artifactId>s3-binding</artifactId>
       <version>${project.version}</version>
     </dependency>
+        <dependency>
```
madrob: nit: whitespace

solr/src/main/conf/schema.xml
ghaughian: It has been copied but changed (a lot) to support the default field names that the ycsb client uses i.e. "field0" - "field9". It is being included to help illustrate the importance and connection between Solr's schema.xml file and ycsb's expected field names as an example only. This was added in response to a previous comment regarding the clarity and important of the schema.xml file in Step 2 of the README.

I might replace the file with an original simplified snippet that likely won't need reference to any license if that makes things simpler?

solr/README.md
madrob: SolrCloud installations typically chroot under /solr in zookeeper, should specify that here.

solr/src/main/conf/schema.xml
```diff
@@ -0,0 +1,613 @@
...
+ or located where the classloader for the Solr webapp can find it.
+
+ This example schema is the recommended starting point for users.
+ It should be kept correct and concise, usable out-of-the-box.
+
+ For more information, on how to customize this file, please see
+ http://wiki.apache.org/solr/SchemaXml
+
+ PERFORMANCE NOTE: this schema includes many optional features and should not
+ be used for benchmarking.  To improve performance one could
```
madrob: Given this caveat, this seems like a bad schema to ship as the example. Probably worth the effort to either start with one of the minimal example schema, or pare this one down quite a bit.

solr/src/main/java/com/yahoo/ycsb/db/SolrDBClient.java
```diff
@@ -0,0 +1,339 @@
...
+  public void cleanup() throws DBException {
+    try {
+      client.close();
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Insert a record in the database. Any field/value pairs in the specified values HashMap will be
```
madrob: IIRC, method level JavaDoc can get inherited by default on an @Override. Copy and pasting the same content doesn't add value, so I think we can drop this.

solr/src/main/conf/schema.xml
vfunshteyn: The field name/quantity selection happens above the `DB` implementation layer (this code), namely in workload code. Why would we couple schema definition here to what the default workload does by default? Someone might create another workload config with a different field number - wouldn't this schema get out of sync?

Also, does Solr really requires a fixed set of user-level field definitions?

ghaughian: Solr does require a schema file that matches the data/fields intended to be
indexed/searched.

This schema file is being included for *example* purposes *only*, for the
sole purpose of illustrating by *example* what a schema file would look
like for the default fields used by the YCSB client. It is *not* expected
to be used for all possible benchmarking studies nor supported in any way.

Perhaps the location of this sample file is throwing people off??

solr/README.md
busbey: Well that name from Accumulo isn't reusable. :(

I'd say just make sure your property starts with "solr." And that'll suffice.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,327 @@
...
+  @Override
+  public void init() throws DBException {
+    Properties props = getProperties();
+    commitTime = Integer
+        .parseInt(props.getProperty("solr.commit.within.time", DEFAULT_COMMIT_WITHIN_TIME));
+    batchMode = Boolean.parseBoolean(props.getProperty("solr.batch.mode", DEFAULT_BATCH_MODE));
+
+    // Check if Solr cluster is running in SolrCloud or Stand-alone mode
+    Boolean cloudMode = Boolean.parseBoolean(props.getProperty("solr.cloud", DEFAULT_CLOUD_MODE));
+    System.out.println("Solr Cloud Mode = " + cloudMode);
```
busbey: This should be on stderr.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,327 @@
...
+    if (cloudMode) {
+      System.out.println("Solr Zookeeper Remote Hosts = "
+          + props.getProperty("zookeeper.host.list", DEFAULT_REMOTE_ZOOKEEPER_HOST));
```
busbey: Debug messages should be on stderr.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,327 @@
...
+    } else {
+      client = new HttpSolrClient(props.getProperty("solr.base.url", DEFAULT_SOLR_BASE_URL));
+    }
+  }
+
+  @Override
+  public void cleanup() throws DBException {
+    try {
+      client.close();
+    } catch (IOException e) {
```
busbey: Shouldn't we propogate this exception?

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
ghaughian: Really? I see several other client bindings use stdout for similar types of messages

busbey: Yes, unfortunately lots of fixing still to do. ;)

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
ghaughian: We should. I have updated the source code

busbey:
Does ES have sufficient tooling that we could include a test?

Anything else, @kruthar, @madrob, @risdenk?

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,327 @@
...
+      query.setParam("fq", "id:[ " + startkey + " TO * ]");
+      if (returnFields) {
+        query.setFields(fieldList);
+      }
+      query.setRows(recordcount);
+      final QueryResponse response = client.query(table, query);
+
+      HashMap<String, ByteIterator> entry;
+
+      for (SolrDocument hit : response.getResults()) {
```
risdenk: nit: pull response.getResults() into a variable? multiple calls to getResults.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,327 @@
...
+        fieldList = fields.toArray(new String[fields.size()]);
+      }
+      SolrQuery query = new SolrQuery();
+      query.setQuery("id:" + key);
+      if (returnFields) {
+        query.setFields(fieldList);
+      }
+      final QueryResponse response = client.query(table, query);
+
+      if ((response.getResults() != null) && (response.getResults().getNumFound() > 0)) {
```
risdenk: nit: pull response.getResults() into a variable? multiple calls to getResults.

solr/src/main/java/com/yahoo/ycsb/db/SolrClient.java
```diff
@@ -0,0 +1,327 @@
...
+      responseStatus = Status.NOT_FOUND;
+      break;
+    case 500:
+      responseStatus = Status.ERROR;
+      break;
+    case 503:
+      responseStatus = Status.SERVICE_UNAVAILABLE;
+      break;
+    default:
+      responseStatus = Status.OK;
```
risdenk: Should the default response status be OK? We should explicitly check for good status and not report a false OK.

busbey:
Looks good to me. Could you please squash the commits and make the commit message start with "[solr]"?

ghaughian:
Awesome, thanks! Commits have been squashed into 1 commit

busbey:
If folks want to take a final look, I'll be merging this in ~5-6 hours barring feedback.

Thanks for all your work on this @ghaughian!

risdenk:
Looks good to me. Thanks @ghaughian

risdenk:
> Does Solr have sufficient tooling that we could include a test?

@busbey I have some working tests for this that I just created PR #583 for

risdenk:
@ghaughian - There are a few comments from @madrob on PR #583 that should be addressed here. They include:
* https://github.com/brianfrankcooper/YCSB/pull/583#discussion_r49488802
* https://github.com/brianfrankcooper/YCSB/pull/583#discussion_r49489253
* https://github.com/brianfrankcooper/YCSB/pull/583#discussion_r49488475

ghaughian:
@risdenk addressing those now; thanks.

PR #577, opened 2016-01-06
[orientdb] properties cleanup

kruthar:
Cleaner handling of properties.
Updated documentation of available properties.
Removed default database url.

Making progress towards #539

### DIFF
--- a/orientdb/README.md
+++ b/orientdb/README.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2012 YCSB contributors. All rights reserved.
+Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License"); you
 may not use this file except in compliance with the License. You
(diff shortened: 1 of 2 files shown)


orientdb/README.md
```diff
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2012 YCSB contributors. All rights reserved.
+Copyright (c) 2016 YCSB contributors. All rights reserved.
```
busbey: Is everything from previous years gone?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -48,13 +48,6 @@
...
  * OrientDB client for YCSB framework.
  * 
- * Properties to set:
- * 
- * orientdb.url=local:C:/temp/databases or remote:localhost:2424 <br>
- * orientdb.database=ycsb <br>
- * orientdb.user=admin <br>
- * orientdb.password=admin <br>
- * 
  * @author Luca Garulli
```
busbey: Can you remove the author tag while you're here?

orientdb/README.md
kruthar: oops. fixed.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: For my own knowledge, should author tags not exist at all? I think I am guilty of auto generated author tags as well.

busbey: I'd prefer not; I don't know if anyone else feels strongly about it.

Essentially, I think of author tags as a "ask this person if there are issues" and I'd prefer we get folks in the habit of asking the community rather than a named individual.

busbey:
+1

PR #578, opened 2016-01-07
Added log4j properties for mongodb to not show debug output as default

Mike-Petersen:
Currently there is a lot of debug output from the mongodb module. By adding this log4j.properties file the DEBUG output is not displayed by default.


mongodb/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2012 YCSB contributors. All rights reserved.
```
busbey: 2016 should be the year

busbey:
Please edit the commit message to start with "[mongodb]"

Mike-Petersen:
I changed the commit message and the year

busbey:
Thanks!

PR #581, opened 2016-01-12
[orientdb] Added remote connection capabilties

kruthar:
Had to implement a separate init path for remote connections using OServerAdmin. The old, local init path is essentially the same as before.

OrientDB does not support the iterator methods that scan is built off of for remote connections, so had to check for that and record NOT_IMPLEMENTED. Made a note of that in the README as well.

### DIFF
--- a/orientdb/README.md
+++ b/orientdb/README.md
@@ -42,16 +42,20 @@ See the next section for the list of configuration parameters for OrientDB.
 ## OrientDB Configuration Parameters
 
 * ```orientdb.url``` - (required) The address to your database.
-    * Supported storage types: memory, plocal
+    * Supported storage types: memory, plocal, remote
     * EX. ```plocal:/path/to/database```
 * ```orientdb.user``` - The user to connect to the database with.
     * Default: ```admin```
 * ```orientdb.password``` - The password to connect to the database with.
     * Default: ```admin```
-* ```orientdb.newdb``` - Create the database if it does not exists, or overwrite the database if it does exist.
+* ```orientdb.newdb``` - Overwrite the database if it already exists.
+    * Only effects the ```load``` phase.
     * Default: ```false```
+* ```orientdb.remote.storagetype``` - Storage type of the database on remote server
+    * This is only required if using a ```remote:``` connection url
 
 ## Known Issues
 
 * There is a performance issue around the scan operation. This binding uses OIndex.iterateEntriesMajor() which will return unnecessarily large iterators. This has a performance impact as the recordcount goes up. There are ideas in the works to fix it, track it here: [#568](https://github.com/brianfrankcooper/YCSB/issues/568).
 * The OIndexCursor used to run the scan operation currently seems to be broken. Because of this, if the startkey and recordcount combination on a particular operation were to cause the iterator to go to the end, a NullPointerException is thrown. With sufficiently high record counts, this does not happen very often, but it could cause false negatives. Track that issue here: https://github.com/orientechnologies/orientdb/issues/5541.
+* Iterator methods needed to perform scans are Unsupported in the OrientDB API for remote database connections and so will return NOT_IMPLEMENTED status if attempted.
--- a/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
+++ b/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
@@ -17,11 +17,14 @@
 
 package com.yahoo.ycsb.db;
 
-import com.orientechnologies.orient.core.config.OGlobalConfiguration;
+import com.orientechnologies.common.exception.OException;
+import com.orientechnologies.orient.client.remote.OEngineRemote;
+import com.orientechnologies.orient.client.remote.OServerAdmin;
 import com.orientechnologies.orient.core.db.ODatabaseRecordThreadLocal;
 import com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx;
 import com.orientechnologies.orient.core.db.record.OIdentifiable;
 import com.orientechnologies.orient.core.dictionary.ODictionary;
+import com.orientechnologies.orient.core.exception.ODatabaseException;
 import com.orientechnologies.orient.core.index.OIndexCursor;
 import com.orientechnologies.orient.core.intent.OIntentMassiveInsert;
 import com.orientechnologies.orient.core.record.ORecord;
@@ -32,6 +35,7 @@ import com.yahoo.ycsb.DBException;
 import com.yahoo.ycsb.Status;
 import com.yahoo.ycsb.StringByteIterator;
 
+import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map.Entry;
 import java.util.Properties;
@@ -46,6 +50,7 @@ public class OrientDBClient extends DB {
   private static final String             CLASS = "usertable";
   protected ODatabaseDocumentTx             db;
   private ODictionary<ORecord> dictionary;
+  private boolean isRemote = false;
 
   private static final String URL_PROPERTY = "orientdb.url";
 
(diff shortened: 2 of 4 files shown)


orientdb/README.md
```diff
@@ -42,16 +42,19 @@ See the next section for the list of configuration parameters for OrientDB.
...
+    * Supported storage types: memory, plocal, remote
     * EX. ```plocal:/path/to/database```
 * ```orientdb.user``` - The user to connect to the database with.
     * Default: ```admin```
 * ```orientdb.password``` - The password to connect to the database with.
     * Default: ```admin```
-* ```orientdb.newdb``` - Create the database if it does not exists, or overwrite the database if it does exist.
+* ```orientdb.newdb``` - Overwrite the database if it already exists.
     * Default: ```false```
+* ```orientdb.remote.storagetype``` - Storage type of the database on remote server
```
busbey: This is surprising.

orientdb/README.md
```diff
@@ -42,16 +42,19 @@ See the next section for the list of configuration parameters for OrientDB.
...
 * ```orientdb.url``` - (required) The address to your database.
-    * Supported storage types: memory, plocal
+    * Supported storage types: memory, plocal, remote
     * EX. ```plocal:/path/to/database```
 * ```orientdb.user``` - The user to connect to the database with.
     * Default: ```admin```
 * ```orientdb.password``` - The password to connect to the database with.
     * Default: ```admin```
-* ```orientdb.newdb``` - Create the database if it does not exists, or overwrite the database if it does exist.
+* ```orientdb.newdb``` - Overwrite the database if it already exists.
```
busbey: Probably worth a note that you should only be setting this in the load phase.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -68,42 +76,70 @@ public void init() throws DBException {
...
-          db.drop();
-          db.create();
-        }
-      } else {
-        System.out.println("OrientDB database not found, create fresh db");
-        db.create();
-      }
-
-      System.out.println("OrientDB connection created with " + url);
+    System.out.println("OrientDB loading database url = " + url);
```
busbey: Can we move these messages to stderr while we're here?

orientdb/README.md
kruthar: So, there really is NO reason this should be set during the run phase, that I can think of. It will always kill the workload because the database will be empty.

Would adding a note of "Only effects load phase", and then only honoring this setting during the load phase be acceptable?

orientdb/README.md
kruthar: Yes, a remote connection can be used to connect to a long living plocal or memory database on the server. Memory databases stored on the server don't die like local memory databases do. This being the case you could have two databases with the same name but different types, and so when doing any administration of remote databases you have to know the storage type.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: Sure, I can do that for now. I have plans to move the whole binding to slf4j, which I think the entire project should move to, but it can be tackled in increments.

orientdb/README.md
busbey: Sounds good.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -46,6 +49,7 @@
   private static final String             CLASS = "usertable";
   protected ODatabaseDocumentTx             db;
   private ODictionary<ORecord> dictionary;
+  private boolean isRemote = false;
```
risdenk: Should this default to true since other bindings default to remote?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -68,42 +76,70 @@ public void init() throws DBException {
...
 
-    } catch (Exception e1) {
-      System.err.println("Could not initialize OrientDB connection pool for Loader: " + e1.toString());
-      e1.printStackTrace();
-      return;
+        server.close();
+        db = new ODatabaseDocumentTx(url).open(user, password);
+      } catch (IOException | OException e) {
+        System.err.println(String.format("Error interfacing with %s", url));
+        e.printStackTrace();
```
risdenk: Should this wrap the error and throw a DBException?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -68,42 +76,70 @@ public void init() throws DBException {
...
+            db.create();
+          }
+        } else {
+          System.err.println("OrientDB database not found, creating fresh db");
+          db.create();
+        }
+      } catch (Exception e1) {
+        System.err.println("Could not initialize OrientDB connection pool for Loader: " + e1.toString());
+        e1.printStackTrace();
+        return;
```
risdenk: Should this wrap the error and throw a DBException?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -68,42 +76,70 @@ public void init() throws DBException {
...
+          if (newdb) {
+            System.err.println("OrientDB dropping and recreating fresh db.");
+            db.drop();
+            db.create();
+          }
+        } else {
+          System.err.println("OrientDB database not found, creating fresh db");
+          db.create();
+        }
+      } catch (Exception e1) {
```
risdenk: Any reason for the catch all Exception and not specific exceptions?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -68,42 +76,70 @@ public void init() throws DBException {
...
+        System.err.println("Could not initialize OrientDB connection pool for Loader: " + e1.toString());
+        e1.printStackTrace();
+        return;
+      }
     }
 
+    System.err.println("OrientDB connection created with " + url);
+    dictionary = db.getMetadata().getIndexManager().getDictionary();
+    if (!db.getMetadata().getSchema().existsClass(CLASS))
+      db.getMetadata().getSchema().createClass(CLASS);
```
risdenk: missing brackets for if?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: Whether or not the connection is a remote connection is purely dependent on the url provided by the user, so really there is not default. ```isRemote``` is defaulted to false because it is only set to true if the user provides a remote connection url, in any other case (user provides a local url) it should be false.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: Yes, it should, thanks for catching this, it was not done correctly initially either but in both code paths throwing a DBException is the right way to go.

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: Purely a formatting change. Although this is marked as an addition, it is original code. Don't worry, it'll get fixed with the checkstyle PR :smile:.

orientdb/README.md
kruthar: Done

kruthar:
Made a few edits:

* Added a note in README about newdb
* No longer honoring newdb if in the transaction phase. Had to modify core to push the ```dotransactions``` property to the workload as that information wasn't currently available downstream.
* Propagating DBExceptions in init()
* Found it unnecessary to split the url, no just passing the server url which is safer.

risdenk:
Looks good to me.

busbey:
+1

PR #489, opened 2015-11-09
[orientdb] Checkstyle updates for the OrientDB binding.

allanbank:

PR #585, opened 2016-01-18
Orientdb checkstyle

kruthar:
* Rabased pull/489/head to master
* Backed out unnecessary line breaks for 120 char limit
* Enforced checkstyle on new code
  * Was required to make db variable private, updated in tests
  * Removed TODO

Closes #489

### DIFF
--- a/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
+++ b/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
@@ -47,8 +47,8 @@ import java.util.Vector;
  */
 public class OrientDBClient extends DB {
 
-  private static final String             CLASS = "usertable";
-  protected ODatabaseDocumentTx             db;
+  private static final String CLASS = "usertable";
+  protected ODatabaseDocumentTx db;
   private ODictionary<ORecord> dictionary;
   private boolean isRemote = false;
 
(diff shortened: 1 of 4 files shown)


orientdb/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2014, Yahoo!, Inc. All rights reserved.
```
busbey: Is this year correct?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -291,4 +327,11 @@ public Status scan(String table, String startkey, int recordcount, Set<String> f
     }
     return Status.ERROR;
   }
+
+  /**
+   * Protected Access method to db variable for unit testing.
+   **/
+  protected ODatabaseDocumentTx getDB() {
```
busbey: Should this be package protected so that the test can get it, buy a subclass wouldn't necessarily?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -71,7 +71,8 @@
   private static final String ORIENTDB_DOCUMENT_TYPE = "document";
 
   /**
-   * Initialize any state for this DB. Called once per DB instance; there is one DB instance per client thread.
+   * Initialize any state for this DB. Called once per DB instance; there is one
```
busbey: I think these are all just the comments from the DB interface. Can we just leave them out and have the javadoc get inherited?

kruthar: Yes, I'd prefer that also

kruthar:
All comments addressed.

busbey:
+1

PR #586, opened 2016-01-18
Orientdb cleanup

kruthar:
Two things fixed here:
* Added logging
* Moved Intent to a property, before it was always being called as massiveinsert which, is not always a good thing to do. It should be at the control of the user.

### DIFF
--- a/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
+++ b/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
@@ -27,6 +27,8 @@ import com.orientechnologies.orient.core.dictionary.ODictionary;
 import com.orientechnologies.orient.core.exception.ODatabaseException;
 import com.orientechnologies.orient.core.index.OIndexCursor;
 import com.orientechnologies.orient.core.intent.OIntentMassiveInsert;
+import com.orientechnologies.orient.core.intent.OIntentMassiveRead;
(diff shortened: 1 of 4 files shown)


orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -134,13 +147,22 @@ public void init() throws DBException {
...
 
-    System.err.println("OrientDB connection created with " + url);
+    log.info("OrientDB connection created with " + url);
     dictionary = db.getMetadata().getIndexManager().getDictionary();
     if (!db.getMetadata().getSchema().existsClass(CLASS)) {
       db.getMetadata().getSchema().createClass(CLASS);
     }
 
-    db.declareIntent(new OIntentMassiveInsert());
+    if (intent.equals(ORIENTDB_MASSIVEINSERT)) {
```
risdenk: Should these be set automatically per type of action? Like insert should be massiveinsert, read should be massiveread, etc? Or is it assumed that these settings will be set when someone runs a workload?

kruthar: No, I don't think they should be set automatically, that is actually why I moved this to a user property. It is an optimization, not a necessary action. Prior to this change MassiveInsert was being set no matter what which didn't much make since for workloads that weren't inserting.

risdenk:
:+1: looks good to me.

Issue #322, opened 2015-06-25
HBase client should use a single Connection object


busbey:
per [the javadocs for HBase Connection](http://hbase.apache.org/apidocs/org/apache/hadoop/hbase/client/Connection.html), we should be making and sharing a single Connection for the application and then creating Table instance per DB instance. As is, we create a ZK connection per thread and tons of threads.

bijugs:
In the current [core/Client.java](https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/Client.java) code, [_db.cleanup()](https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/Client.java#L364-L368) is called on each thread which prevents from having a single db connection object across multiple threads. This prevents db like ``hbase`` not being able to use a single connection object across threads which is a good practice and also aligns with how applications are written. This may be applicable to other databases as well. 

If we [move out the logic to call the cleanup()](https://github.com/bijugs/YCSB/blob/single-conn/core/src/main/java/com/yahoo/ycsb/Client.java#L364-L378) method after all the thread execution in Client.java (here)[https://github.com/bijugs/YCSB/blob/single-conn/core/src/main/java/com/yahoo/ycsb/Client.java#L916-L925], then DBs like ``hbase`` can work with a [single connection object](https://github.com/bijugs/YCSB/blob/single-conn/hbase098/src/main/java/com/yahoo/ycsb/db/HBaseClient.java#L66).

**Note** that the quick hack referred in the write-up is to make the conversation easier.

bijugs:
@busbey . Regarding your query about stats from using single hbase connection and multi-connections in PR #591  , here are some from **hbase10** load and run with 10 threads and 1 million recordcount/operation count. The test was run a 4 RS VM cluster with hbase 0.98 on it.

|**hbase10**	|load|Multi -Conn|	Single-Conn
-------------------|------|----------------|--------------------
[OVERALL]|	 RunTime(ms)|	935029|	915693
[OVERALL]|	 Throughput(ops/sec)|	1069.485545|	1092.069067
[CLEANUP]|	 Operations|	20|	20
[CLEANUP]|	 AverageLatency(us)|	93268.8|	19950.7
[CLEANUP]|	 MinLatency(us)|	5|	5296
[CLEANUP]|	 MaxLatency(us)|	359679|	290815
[...]

PR #350, opened 2015-07-10
[hbase] Add kerberos authentication for HBase binding

uijin:
Continue unfinished pull request #333. Rebase on master branch, and add Kerberos authentication to HBaseClient10.java .

PR #591, opened 2016-01-20
[hbase] Changes to support secured HBase cluster 

bijugs:
Changes in this pull request include

**For hbase098**
- Use HBase ``connection`` object to create ``table`` object
- Changes to make YCSB hbase tests run against a kerberized hbase cluster
- Update to README.txt to document the new parameters

**For hbase10**
- Changes to make YCSB hbase tests run against a kerberized hbase cluster

If users like to use **kerberos** keytab to make the connection to secure hbase cluster instead of using the ``user-id`` of the user running the test they can be passed using ``principal=keytab-principal-name`` and ``keytab=keytab=file`` properties.

**Update:** After the conversation in issue #322 , changes were made to **hbase098** and **hbase10** to use single HBase connection across all the threads when running a multi-thread test.

### DIFF
--- a/hbase098/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
+++ b/hbase098/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
@@ -50,6 +50,7 @@ import java.util.Properties;
 import java.util.Random;
 import java.util.Set;
 import java.util.Vector;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * HBase client for YCSB framework
@@ -59,11 +60,12 @@ public class HBaseClient extends com.yahoo.ycsb.DB
     // BFC: Change to fix broken build (with HBase 0.20.6)
     //private static final Configuration config = HBaseConfiguration.create();
     private static final Configuration config = HBaseConfiguration.create(); //new HBaseConfiguration();
+    private static final AtomicInteger THREAD_COUNT = new AtomicInteger(0);
 
     public boolean _debug=false;
 
     public String _table="";
-    public HConnection _hConn=null;
+    private static HConnection _hConn=null;
     public HTableInterface _hTable=null;
     public String _columnFamily="";
     public byte _columnFamilyBytes[];
@@ -112,7 +114,12 @@ public class HBaseClient extends com.yahoo.ycsb.DB
             }
         }
         try {
-            _hConn = HConnectionManager.createConnection(config);
+            THREAD_COUNT.getAndIncrement();
+            synchronized(THREAD_COUNT) {
+              if (_hConn == null){
+                _hConn = HConnectionManager.createConnection(config);
+              }
+            }
         } catch (IOException e) {
             System.err.println("Connection to HBase was not successful");
             throw new DBException(e);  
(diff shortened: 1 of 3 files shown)


busbey:
please start commit messages with just then name of the module impacted (i.e. "[hbase098]" rather than "changes to hbase098"

busbey:
>For hbase098
>
> - Use a single HBase connection object across all the threads which is recommended

Do you have some perf numbers to show the actual impact?

Why only do this change in the 0.98 driver?

busbey:
If you're going to add properties for the principal and keytab, please document them

- In the relevant binding README
- In the workload template file

Are these changes needed to run on a kerberos cluster? I was under the impression the one merely needed to kinit before running YCSB and the hbase driver would correctly use the cached tickets?

hbase098/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
```diff
@@ -94,7 +99,23 @@ public void init() throws DBException
...
+        if ( (getProperties().getProperty("principal")!=null) && (getProperties().getProperty("keytab")!=null) ){
+            try {
+                UserGroupInformation.loginUserFromKeytab(getProperties().getProperty("principal"), getProperties().getProperty("keytab"));
+            } catch (IOException e) {
+                System.err.println("Keytab file is not readable or not found");
+                throw new DBException(e);
+            }
+        }
+        try {
+            _hConn = HConnectionManager.createConnection(config);
```
busbey: where does this get closed? should either ref count or register a shutdown hook

busbey:
Note also that #350 has been waiting for verification of essentially the same keytab functionality.

bijugs:
Thanks @busbey for the review and feedback.
- please start commit messages with just then name of the module impacted (i.e. "[hbase098]" rather than "changes to hbase098"
  - Will change the commit message to start with module name.
- Why only do this change in the 0.98 driver?
  - hbase10 is already using ``connection`` object to create ``table`` objects and I don't have a 0.94 environment to test the changes.
- If you're going to add properties for the principal and keytab, please document them
  - Will document them in the module specific readme
- Are these changes needed to run on a kerberos cluster?
  - We are testing against a Kerberized cluster from a node outside the cluster. These changes should not impact the current behavior
- Do you have some perf numbers to show the actual impact? & where does this get closed?
  - Since [core/Client.java](https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/Client.java) calls [_db.cleanup()](https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/Client.java#L367), at this time single connection object will not work. To complete this PR will close the connection on the cleanup() method. Will raise another request to update the [core/Client.java] so that DBs which can work with single connection object shared among multiple threads will be able to leverage it.

risdenk:
@busbey and @bijugs There is a related issue #322 for HBase single connection object. Just wanted to cross reference it here.

bijugs:
@busbey Changes were made for the review comments.

@risdenk Thanks for pointing out the existing issue. Will update with my comments there so that we can have a conversation on the proposed change.

cmatser:
:+1: from me

busbey:
Thanks for the contribution!

PR #592, opened 2016-01-22
Dynamodb API update

yuyantingzero:
Run workloada from an external vm to dynamodb, the performance difference between  dynamodb v1 and v2 api. Results:
percentile,	v1 read,	v1 write,	v2 read,	v2 write
  0.1,	34775.9,  	36097,	35244,	36357
  0.5,	35869.5,          36843,	36658,	37602
0.95,	39209.1,  	  42328.75,	41423,	43277.9
0.99,	54295.1,  	68858.76,	56915,	63764.16

### DIFF
--- a/dynamodb/src/main/java/com/yahoo/ycsb/db/DynamoDBClient.java
+++ b/dynamodb/src/main/java/com/yahoo/ycsb/db/DynamoDBClient.java
@@ -21,19 +21,16 @@ import com.amazonaws.AmazonServiceException;
 import com.amazonaws.ClientConfiguration;
 import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.PropertiesCredentials;
-import com.amazonaws.services.dynamodb.AmazonDynamoDBClient;
-import com.amazonaws.services.dynamodb.model.AttributeValue;
-import com.amazonaws.services.dynamodb.model.AttributeValueUpdate;
-import com.amazonaws.services.dynamodb.model.DeleteItemRequest;
-import com.amazonaws.services.dynamodb.model.DeleteItemResult;
-import com.amazonaws.services.dynamodb.model.GetItemRequest;
-import com.amazonaws.services.dynamodb.model.GetItemResult;
-import com.amazonaws.services.dynamodb.model.Key;
-import com.amazonaws.services.dynamodb.model.PutItemRequest;
-import com.amazonaws.services.dynamodb.model.PutItemResult;
-import com.amazonaws.services.dynamodb.model.ScanRequest;
-import com.amazonaws.services.dynamodb.model.ScanResult;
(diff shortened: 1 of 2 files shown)


yuyantingzero:
@stfeng , PTAL

stfeng:
Code LGTM. Thanks for doing the performance regression validation to make sure the new API does not perform worse than the old one. (Your results show they are within ~0.3% of each other)

stfeng:
Please take a look at the Travis CI build error though...

stfeng:
Thanks for taking care of the Travis build issue. All green now. 

One final minor thing: Can you edit both of your commits to start with [DynamoDB]? We always prefix commits with [DB-binding-name], or [core] (if the code is in YCSB core).

dynamodb/src/main/java/com/yahoo/ycsb/db/DynamoDBClient.java
```diff
@@ -337,20 +333,20 @@ public Status delete(String table, String key) {
...
+    private Map<String, AttributeValue> createPrimaryKey(String key) {
+        Map<String, AttributeValue> k = new HashMap<String, AttributeValue>();
         if (primaryKeyType == PrimaryKeyType.HASH) {
-          k = new Key().withHashKeyElement(new AttributeValue().withS(key));
+          k.put(primaryKeyName, new AttributeValue().withS(key));
         } else if (primaryKeyType == PrimaryKeyType.HASH_AND_RANGE) {
-          k = new Key()
-                .withHashKeyElement(new AttributeValue().withS(hashKeyValue))
-                .withRangeKeyElement(new AttributeValue().withS(key));
+          k = new HashMap<String, AttributeValue>();
```
risdenk: Does k need to be a new hashmap here? Looks like it is initialized on line 343 already.

risdenk:
@yuyantingzero does dynamodb use v1 anymore? Will these changes make testing some versions of DynamoDB not possible?

busbey:
as a point of clarification for @risdenk's comment, making it so we can't test earlier versions of DynamoDB isn't a blocker. It just means we have to weigh the benefit of maintaining that ability against the cost of breaking this into a new binding module (see the HBase modules or Cassandra modules as an example).

stfeng:
@risdenk, the DynamoDB v1 client API has been deprecated since Apr 2013 when AWS released the 1.4.2 Java SDK. Upgrading to v2 client API is a long overdue task for this DB binding. ;)

risdenk:
Great thanks @stfeng

yuyantingzero:
Thanks for the comments, @risdenk , @stfeng , @busbey . Fixed the commit message and addressed comment on line 347.

risdenk:
Looks good to me :+1:

stfeng:
LGTM, merging.

Issue #340, opened 2015-07-06
Accumulo binding errors during cleanup


busbey:
While running through RC3 for #266 for Accumulo, my workload A load got errors during cleanup for several threads. All looked like one of the two following:

```
com.yahoo.ycsb.DBException: org.apache.accumulo.core.client.MutationsRejectedException: # constraint violations : 0  security codes: {}  # server errors 0 # exceptions 1
        at com.yahoo.ycsb.db.AccumuloClient.cleanup(AccumuloClient.java:105)
        at com.yahoo.ycsb.DBWrapper.cleanup(DBWrapper.java:74)
        at com.yahoo.ycsb.ClientThread.run(Client.java:273)
Caused by: org.apache.accumulo.core.client.MutationsRejectedException: # constraint violations : 0  security codes: {}  # server errors 0 # exceptions 1
        at org.apache.accumulo.core.client.impl.TabletServerBatchWriter.checkForFailures(TabletServerBatchWriter.java:537)
        at org.apache.accumulo.core.client.impl.TabletServerBatchWriter.close(TabletServerBatchWriter.java:354)
        at org.apache.accumulo.core.client.impl.BatchWriterImpl.close(BatchWriterImpl.java:54)
        at com.yahoo.ycsb.db.AccumuloClient.cleanup(AccumuloClient.java:102)
        ... 2 more
Caused by: org.apache.accumulo.core.client.impl.ThriftTransportPool$TransportPoolShutdownException
        at org.apache.accumulo.core.client.impl.ThriftTransportPool.getCache(ThriftTransportPool.java:656)
        at org.apache.accumulo.core.client.impl.ThriftTransportPool.returnTransport(ThriftTransportPool.java:526)
        at org.apache.accumulo.core.util.ThriftUtil.returnClient(ThriftUtil.java:118)
        at org.apache.accumulo.core.client.impl.TabletServerBatchWriter$MutationWriter.sendMutationsToTabletServer(TabletServerBatchWriter.java:930)
        at org.apache.accumulo.core.client.impl.TabletServerBatchWriter$MutationWriter.access$1600(TabletServerBatchWriter.java:616)
        at org.apache.accumulo.core.client.impl.TabletServerBatchWriter$MutationWriter$SendTask.send(TabletServerBatchWriter.java:801)
        at org.apache.accumulo.core.client.impl.TabletServerBatchWriter$MutationWriter$SendTask.run(TabletServerBatchWriter.java:765)
[...]

PR #596, opened 2016-01-25
Accumulo cleanup and tests

busbey:
Correct some outstanding debt in the accumulo binding and add some tests.

- binding should be in its own java package
- remove the producer-consumer-via-zookeeper implementation that isn't used
- close out #340 
- add AccumuloMiniCluster based tests


risdenk:
@busbey looks good to me. It would be nice if there were a common test framework for testing workloads.

Side question: Is there a plan to move bindings to their own packages? Should this be noted in #245?

busbey:
in general bindings should be in their own package, so I think mentioning it in the guidelines is a good idea.

risdenk:
@busbey Added a note to the guidelines.

:+1: from me for this PR.

PR #598, opened 2016-01-27
Cleanup use of checkstyle and instructions

busbey:
- remove defunct BUILD text
- switch to checkstyle:check because it has defaults that match our use better than checkstyle:checkstyle
- default to enforcing for everything and rely on inherited configs instead of repeating
- on modules that don't pass, skip fail on violations using properties
- remove the checkstyle rule for TODO flagging (seems it wasn't getting enforced by checkstyle:checkstyle)


kruthar:
License updates on just about everything :smile:

busbey:
good catch! updated files that were modified but not deleted.

kruthar:
Checking to make sure this is expected output.

Running either of these displays checkstyle errors, but does not fail execution:

```
mvn test
mvn package
```

Running either of these displays checkstyle errors and fails execution:

```
mvn checkstyle:checkstyle
mvn checkstyle:check
```

busbey:
Maybe?

The first two are using the configured execution. On modules where we have errors, the property is set to not fail.

The latter two run the goal directly, which won't use our configured checkstyle.xml file. So more things in the "should be passing" modules fail.

You can verify the first two are checking correctly by editing the checkstyle.xml file and e.g. picking a really short line length.

busbey:
bump?

kruthar:
+1 - looks good to me, things are compiling. It'll be nice when all the checkstyle changes are done and the only checkstyle artifacts are at the top level.

PR #600, opened 2016-01-27
[jdbc] solidified field ordering

### LINKED ISSUE #132: update method in JdbcDBClient.java
In update method of JdbcDBClient.java, following lines of code get values from a map in order of entries:
      for (Map.Entry<String, ByteIterator> entry : values.entrySet()) {
        updateStatement.setString(index++, entry.getValue().toString());
      }
Since, Map does not guarantee order of entries, though, it doesn't make a big difference for usertable as all the columns are of the same type, varchar and random values are passed, however, while verification, corresponding values are not reflected . It would be good to get the corresponding values from the map for desired fields.

Following is a excerpt from update() method: 

public int update(String tableName, String key, HashMap<String, ByteIterator> values) {
........
.......
 PreparedStatement updateStatement = cachedStatements.get(type);
 ....
......    
      int index = 1;
      for (Map.Entry<String, ByteIterator> entry : values.entrySet()) {
        updateStatement.setString(index++, entry.getValue().toString());
      }
      updateStatement.setString(index, key);

kruthar:
Changes to account for the proper field ordering. Now correctly inserting and updating field values so that they can be verified when using ```dataintegrity=true```. Fixes #132.

Main change here is that the StatementType is now built with an ordered listing of the fields for the statement. Some statements do not require this (scan, read, delete) and so just get an empty string. Also now able to be more precise in tests.

Each insert and update operation does have a little work to do to parse the fields for the operation. I did some performance comparisons and found the difference is not negligible, but also not dramatic or consistently slower.

||original|fixed|
|---|---|---|
|```[INSERT], AverageLatency(us)```|8748.63398|9805.35678|
|```[SCAN], AverageLatency(us)```|9463.342140738743|9258.540857848042|
|```[READ], AverageLatency(us)```|9570.516845735508|9105.605827304706|
|```[UPDATE], AverageLatency(us)```|9959.60669122772|9073.723949453415|

### DIFF
--- a/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
+++ b/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
@@ -84,6 +84,27 @@ public class JdbcDBClient extends DB {
   private ConcurrentMap<StatementType, PreparedStatement> cachedStatements;
 
   /**
+   * Ordered field information for insert and update statements.
+   */
+  private static class OrderedFieldInfo {
+    private String fieldKeys;
+    private List<String> fieldValues;
+
+    OrderedFieldInfo(String fieldKeys, List<String> fieldValues) {
+      this.fieldKeys = fieldKeys;
+      this.fieldValues = fieldValues;
+    }
+
+    String getFieldKeys() {
+      return fieldKeys;
+    }
+
+    List<String> getFieldValues() {
+      return fieldValues;
+    }
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


cmatser:
Regarding the performance impact, as long as this mimics typical real-world behavior (which I believe this does) I don't have a problem with it.  In fact, I would argue that that is what YCSB should do.  It'd be very easy to speed up performance by disregarding the data returned, but that wouldn't be a real-world measurement.  So, thanks for working on this and taking the time to do the performance comparison.
:+1: for this PR

PR #609, opened 2016-02-02
[orientdb] put db conn stuff in sync method

### LINKED ISSUE #571: [orientdb] fails with multiple threads when database does not already exists

kruthar:
All db connection and creation logic is the same, it is just pulled into
a synchronized method so that multiple threads run through the
connection process one at a time.

Fixes #571

### DIFF
--- a/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
+++ b/orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
@@ -80,30 +80,35 @@ public class OrientDBClient extends DB {
   private static final String ORIENTDB_MASSIVEREAD = "massiveread";
   private static final String ORIENTDB_NOCACHE = "nocache";
 
-  private final Logger log = LoggerFactory.getLogger(getClass());
+  private static final Logger LOG = LoggerFactory.getLogger(OrientDBClient.class);
 
-  @Override
-  public void init() throws DBException {
-    Properties props = getProperties();
-
+  /**
+   * This method abstracts the administration of OrientDB namely creating and connecting to a database.
+   * Creating a database needs to be done in a synchronized method so that multiple threads do not all try
+   * to run the creation operation simultaneously, this ends in failure.
+   *
+   * @param props Workload properties object
+   * @return a usable ODatabaseDocumentTx object
+   * @throws DBException
+   */
+  private static synchronized ODatabaseDocumentTx initDB(Properties props) throws DBException {
     String url = props.getProperty(URL_PROPERTY);
     String user = props.getProperty(USER_PROPERTY, USER_PROPERTY_DEFAULT);
     String password = props.getProperty(PASSWORD_PROPERTY, PASSWORD_PROPERTY_DEFAULT);
     Boolean newdb = Boolean.parseBoolean(props.getProperty(NEWDB_PROPERTY, NEWDB_PROPERTY_DEFAULT));
     String remoteStorageType = props.getProperty(STORAGE_TYPE_PROPERTY);
-    String intent = props.getProperty(INTENT_PROPERTY, INTENT_PROPERTY_DEFAULT);
-    Boolean dotransactions = Boolean.parseBoolean(
-        props.getProperty(DO_TRANSACTIONS_PROPERTY, DO_TRANSACTIONS_PROPERTY_DEFAULT));
+    Boolean isrun = Boolean.parseBoolean(props.getProperty(DO_TRANSACTIONS_PROPERTY, DO_TRANSACTIONS_PROPERTY_DEFAULT));
+
+    ODatabaseDocumentTx dbconn;
 
     if (url == null) {
       throw new DBException(String.format("Required property \"%s\" missing for OrientDBClient", URL_PROPERTY));
     }
 
-    log.info("OrientDB loading database url = " + url);
+    LOG.info("OrientDB loading database url = " + url);
 
     // If using a remote database, use the OServerAdmin interface to connect
     if (url.startsWith(OEngineRemote.NAME)) {
-      isRemote = true;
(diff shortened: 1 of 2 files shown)


orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -80,30 +80,26 @@
   private static final String ORIENTDB_MASSIVEREAD = "massiveread";
   private static final String ORIENTDB_NOCACHE = "nocache";
 
-  private final Logger log = LoggerFactory.getLogger(getClass());
-
-  @Override
-  public void init() throws DBException {
-    Properties props = getProperties();
+  private static final Logger LOG = LoggerFactory.getLogger(OrientDBClient.CLASS);
```
busbey: Should this be a lower case ".class"?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
```diff
@@ -80,30 +80,26 @@
...
 
+  private static synchronized ODatabaseDocumentTx initDB(Properties props) throws DBException {
```
busbey: Please add javadoc explaining why this needs to be synchronized (it's not obvious because the various clients don't appear to share connection objects).

I think it's because of the automatic table creation?

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: Nice catch. OrientDBClient.CLASS="usertable".

orientdb/src/main/java/com/yahoo/ycsb/db/OrientDBClient.java
kruthar: done

risdenk:
Would there be issues here if multiple YCSB clients are run at once? This will only help fix multiple threads in one client correct?

Does the initdb need to create the db? Is it possible to make it necessary to have the db already exist as a prerequisite? Like the other bindings ie: Solr and Cassandra

busbey:
We could add a note to the README for the folks who use multiple Client invocations, letting them know how to create it in advance?

kruthar:
@risdenk 

>Would there be issues here if multiple YCSB clients are run at once? This will only help fix multiple threads in one client correct?

Correct.

> Does the initdb need to create the db? Is it possible to make it necessary to have the db already exist as a prerequisite? Like the other bindings ie: Solr and Cassandra

So, the OrientDB client is the only one that I am aware of that actually creates a database for you. All others require it to be present. I kept this functionality because it was there from the initial creation of the client. I'm ok with removing it, or just writing a note about that specific scenario.

busbey:
Either a note or removal sounds fine by me.

Could also move the creation into a utility class that the README references. I think I've seen a couple of those.

kruthar:
I decided to keep the db creation functionality in place since 1) it works and 2) mongodb does "create" a database if it doesn't already exist, although in a much more passive way.

I added a note about it to the README

busbey:
+1

PR #172, opened 2014-05-24
YCSB w/t Couchbase support

maisaengineering:
I was hoping someone from CouchBase would add CouchBase support for YCSB. Apparently missing still.
This fork adds CouchBase support to YCSB. README should help in getting environment up and running.

I am ok if you reject this as this is about 6months old work.

-Chandan Benjaram

PR #610, opened 2016-02-03
[couchbase] implemented scan for couchbase

kruthar:
Adapted from #172

### DIFF
--- a/couchbase/src/main/java/com/yahoo/ycsb/db/CouchbaseClient.java
+++ b/couchbase/src/main/java/com/yahoo/ycsb/db/CouchbaseClient.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013 Yahoo! Inc. All rights reserved.
+ * Copyright (c) 2013 - 2016 YCSB contributors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You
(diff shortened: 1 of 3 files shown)


couchbase/src/main/java/com/yahoo/ycsb/db/CouchbaseClient.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013 Yahoo! Inc. All rights reserved.
+ * Copyright (c) 2013 - 2016 Yahoo! Inc. All rights reserved.
```
busbey: 2013 Yahoo! Inc, 201? - 2016 YCSB Contributors

busbey:
how do you want to handle attribution of authorship given the basis on #172?

kruthar:
I'm not sure, I guess what I could do is start with the commit that this came from and then add a commit of my own to get things in the final state. That may introduce a bunch of irrelevant changes for one commit, but that way the original author has a commit in history. Would that work?

busbey:
Sure, that would work. We could also do what some ASF projects do and have the author for the commit set to the original author and then list you in the commit message with

```
Amending-Author: First Last <email@example.com>
```

I'm fine with either approach.

kruthar:
OK - license fixed, and cherry picked relevant changes from the original PR as the basis for this one.

busbey:
+1

PR #616, opened 2016-02-07
[solr] tune down log output during tests.

busbey:
the Solr tests are particularly chatty at the default log settings. this PR keeps them at INFO messages from the YCSB classes, but reduces them to ERROR from the Solr / jetty/ zk ones.


kruthar:
+1, works for me

risdenk:
:+1: looks good

PR #620, opened 2016-02-11
Add Geode driver

sbawaska:
This is a retry of #388.


busbey:
please update the most recent set of commits so their commit message starts with the component impacted, e.g. "[geode]".

geode/README.md
```diff
@@ -0,0 +1,67 @@
+<!--
+Copyright (c) 2014 - 2015 YCSB contributors. All rights reserved.
```
busbey: some of these edits are 2016, so the year range should include it.

geode/pom.xml
```diff
@@ -47,10 +47,4 @@ LICENSE file.
     </dependency>
   </dependencies>
   
-  <repositories>
-    <repository>
-      <id>gemstone</id>
-      <url>http://dist.gemstone.com.s3.amazonaws.com/maven/release</url>
```
busbey: :clap:

geode/src/main/java/com/yahoo/ycsb/db/GeodeClient.java
```diff
@@ -64,23 +64,23 @@
  * @author Swapnil Bawaskar (sbawaska at vmware)
```
busbey: while you're here, would you mind clearing out @author tags? we rely on git commit metadata for authorship attribution.

busbey:
I'd say this looks good to go, once the couple of nits above are addressed.

Thanks for sticking with getting this in, and kudos to the Geode community on their first release (and some impressive API compatibility)

Note to whomever ends up managing the next release: this should be labeled as a breaking change.

sbawaska:
Thanks for taking a look Sean! I have incorporated your feedback.

PR #625, opened 2016-02-15
Orientdb scan fix

kruthar:
Fixed two issues in the scan operation for OrientDBClient:

1. orientdb client was not handling fields being null which happens when you use ```readallfields=true```. I handled this the way a lot of other clients handle it, which is to just not iterate through the results.
2. There was a known issue where the scan operation would throw a NullPointerException if the scan range ran over the end of the record key range. This was because of a bad mixing of iterator implementations. This has now been fixed.


risdenk:
:+1: 

Should an issue be opened about bindings skipping reading all fields?

busbey:
There is one. It's named something like "wrapper must drain all client returned iterators"

Issue #624, opened 2016-02-15
Release version 0.7.0


risdenk:
For the next release, 0.7.0, the following items have already been merged:

* Apache Solr binding was added: #574, #583
* GemFire -> Apache Geode (incubating) binding: #620
* OrientDB improvements: #567, #569, #572, #577, #581, #585, #586, #609, #625
* HBase Kerberos support and use single connection: #591
* Accumulo improvements: #570, #596
* JDBC improvements: #588, #600
* Couchbase scan implementation: #610
* MongoDB improvements: #522, #578
* DynamoDB version increase: #592
* Elasticsearch version increase: #552
* Code formatting/cleanup: #555, #556, #557, #559, #561, #564, #565, #566

Breaking changes:
* gemfire binding changed to geode

Bindings to test:
* Solr [newly added]
* GemFire -> Geode [newly added]
* OrientDB [fixes]
* HBase [fixes]
* Accumulo [fixes]
* JDBC [fixes]
* Couchbase [fixes]
* MongoDB [fixes]
* DynamoDB [version bump]
* Elasticsearch [version bump]

Release plan:
* RC1 - Tues Feb 16th
* 0.7.0 - by Fri Feb 26th

risdenk:
RC1 is out: https://github.com/brianfrankcooper/YCSB/releases/tag/0.7.0-RC1

risdenk:
First draft of release notes (a bit early but figured I'd get a start on it): https://gist.github.com/risdenk/2c739719a815c4e74c64

busbey:
Excellent! thanks for driving this Kevin.

I can take care of Accumulo and Windows platform testing.

* @ghaughian could you test out the Release Candidate to make sure it works with SOLR?
* @sbawaska could you check out the Release Candidate to make sure the new geode binding works?

risdenk:
@kruthar Can you take care of testing OrientDB, JDBC, and Couchbase bindings?

sbawaska:
The Geode binding looks good.

busbey:
I tried building on windows and ran into #617 again. submitted PR #626 to skip the Accumulo tests on windows. We should cherry-pick that back for the 0.7.0 release to make sure building from source on windows works.

risdenk:
@busbey Merged #626. I'll cherry-pick that commit back to 0.7.0 and respin the RC.

risdenk:
@busbey had to back out #626. When it is fixed I can merge it and respin the RC.

[...]

PR #629, opened 2016-02-22
[orientdb] fixed scan with readallfields

kruthar:
I was handling the null fields case with ```readallfields=true``` very poorly. The result was that it had to do no work with ```readallfields=true``` and so the performance was drastically different in a counterintuitive way.

This PR fixes that. @risdenk can we pull this into 0.7.0?


risdenk:
Tagging the release issue #624

risdenk:
:+1: Thanks @kruthar

PR #641, opened 2016-03-04
[core] added real version printout

kruthar:
Not sure if this is the best way, but it seems ok to me. Opens up any other project details that you may want to expose to the client.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Client.java
+++ b/core/src/main/java/com/yahoo/ycsb/Client.java
@@ -771,16 +771,7 @@ public class Client
       double targetperthread=((double)target)/((double)threadcount);
       targetperthreadperms=targetperthread/1000.0;
     }
-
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -809,6 +800,22 @@ public void run()
     //load the workload
     ClassLoader classLoader = Client.class.getClassLoader();
 
+    try {
+      Properties projectProp = new Properties();
+      projectProp.load(classLoader.getResourceAsStream("project.properties"));
+      System.out.println("YCSB Client " + projectProp.getProperty("version"));
```
busbey: can we put all of these on System.err instead of System.out?

kruthar:
changed printouts to System.err

PR #651, opened 2016-03-09
[hbase10] Fix the init() method to properly construct a single…

manolama:
Connection object. Previously N objects were instantiated because the method didn't
check to see if the connection was null.

### DIFF
--- a/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
+++ b/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
@@ -51,7 +51,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.Vector;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * HBase 1.0 client for YCSB framework.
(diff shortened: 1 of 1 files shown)


hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -132,9 +132,11 @@ public void init() throws DBException {
     }
 
     try {
-      THREAD_COUNT.getAndIncrement();
       synchronized(THREAD_COUNT) {
-        connection = ConnectionFactory.createConnection(config);
+        THREAD_COUNT.getAndIncrement();
```
busbey: this change makes it so that all of our accesses for THEAD_COUNT are now in synchronized blocks, so it need not be an AtomicLong.

Would you mind changing it to a normal Long with a comment that it must be accessed within a synchronized block?

busbey: I guess synchronize on the Connection object, since that will still be a proper Object shared wherever the THREAD_COUNT is.

busbey:
Thanks for chasing this down! This'll be a big help.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
manolama: Fixed it up so that:
``private static final AtomicInteger THREAD_COUNT = new AtomicInteger(0)``
is now
`` private static Integer threadCount = 0``

and locking on that since we have to have a non-null object to synchronize on. And check style forced the name change :(

busbey:
thanks again!

Issue #653, opened 2016-03-15
[core] Feature Request - Time series workloads


manolama:
In the past few years we've seen a growing number of database systems for storing time series data. DBs like OpenTSDB, Druid, InfluxDB along with updates to standbys like Graphite and Ganglia. We'll likely see even more specialized DBs for IOT use cases. 

I've started playing around with creating time series workloads that would enable load testing these services. So far it looks like it will be pretty straight-foward to implement in that the YCSB API's don't have to change. Rather we can just pass timestamps and numeric values in the field list and let the DB implementations handle the data properly. 

So would this be a welcome addition that other folks could use? Any thoughts on it? Thanks!

kruthar:
:+1:  I think that functionality would definitely be interesting. If you can find a way to slide it into the current DB interface then all the better.

manolama:
Cool, I'll start a whole bunch of small PRs tied to this issue then. Thanks!

busbey:
This would be a great addition. Thanks for taking it on!

manolama:
Happy to help! It's about time Yahoo contributed again.

PR #656, opened 2016-03-17
[core] Add a UnixEpochTimestampGenerator

manolama:
Tied to generating time series workloads: https://github.com/brianfrankcooper/YCSB/issues/653

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
+++ b/core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
@@ -1,0 +1,178 @@
+/**
+ * Copyright (c) 2016 YCSB contributors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+package com.yahoo.ycsb.generator;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A generator that produces Unix epoch timestamps in seconds, milli, micro or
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
```diff
@@ -0,0 +1,188 @@
+/**
+ * Copyright (c) 2016 Yahoo! Inc. All rights reserved.
```
risdenk: Instead of Yahoo! Inc. This should be `Copyright (c) 2016 YCSB contributors. All rights reserved.`

Same goes for the test file.

manolama:
Fixed up the copyrights, thanks!

busbey:
Would you mind updating the commit message to start with the module impacted? In this case it'd be something like "[core] ..."

core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+ * }</pre>
+ * A constructor is also present for setting an explicit start time. 
+ * Negative intervals are supported as well for iterating back in time.
+ * <p>
+ * WARNING: This generator is not thread safe and should not called from multiple
+ * threads.
+ */
+public class UnixEpochTimestampGenerator extends Generator<Long> {
+  /** Name and default value for the timestamp interval property. */
+  public static final String TIMESTAMP_INTERVAL_PROPERTY = "timestamp_interval";
```
busbey: Where are these properties getting used? What are they for?

manolama: They'll be used by the TimeseriesWorkload class once it's ready. I figured it'd be good to stash them here as they could be used by multiple workloads. Or I could just move them there if you think that's better.

manolama:
Commit message updated, thanks!

core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
busbey: Other workloads might want to have multiple timestamps, which would then make these properties confusing. Probably best to keep them in the TimeseriesWorkload.

Thanks!

manolama: Sounds good, yanked those guys out and I'll toss 'em in with the TS worklaod.

busbey:
thanks!

PR #657, opened 2016-03-17
[core] Add byte to double and long conversion methods to Utils.

manolama:
These will be used for #653 to write timestamps and numeric values
to the ByteIterator maps.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Utils.java
+++ b/core/src/main/java/com/yahoo/ycsb/Utils.java
@@ -1,5 +1,5 @@
 /**                                                                                                                                                                                
- * Copyright (c) 2010 Yahoo! Inc. All rights reserved.                                                                                                                             
+ * Copyright (c) 2010 Yahoo! Inc., 2016 YCSB contributors. All rights reserved.                                                                                                                             
  *                                                                                                                                                                                 
  * Licensed under the Apache License, Version 2.0 (the "License"); you                                                                                                             
  * may not use this file except in compliance with the License. You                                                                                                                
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/yahoo/ycsb/TestUtils.java
```diff
@@ -0,0 +1,124 @@
+/**
+ * Copyright (c) 2016 Yahoo! Inc. All rights reserved.
```
risdenk: Instead of Yahoo! Inc. This should be `Copyright (c) 2016 YCSB contributors. All rights reserved.`

risdenk:
The copyright for the top of `Utils.java` should be Copyright (c) 2010 Yahoo! Inc., 2016 YCSB contributors. All rights reserved.`

manolama:
Roger that, tweaked the copyrights.

busbey:
Please update the commit message to start with the impacted module, e.g. "[core] ..."

core/src/main/java/com/yahoo/ycsb/Utils.java
```diff
@@ -113,4 +113,64 @@ public static long FNVhash64(long val)
...
+        return bytes;
+      }
+      
+      /**
+       * Parses the byte array into a double.
+       * The byte array MUST be 8 bytes long and have been encoded using 
+       * {@link #doubleToBytes}.
+       * @param bytes The byte array to parse, at least 8 bytes.
+       * @return A double value read from the byte array.
+       * @throws IllegalArgumentException if the byte array is not 8 bytes wide.
```
busbey: This description should match the actual check.

core/src/test/java/com/yahoo/ycsb/TestUtils.java
```diff
@@ -0,0 +1,124 @@
...
+    bytes = new byte[] { 63, -16, 0, 0, 0, 0, 0, 0, 42, 42, 42 };
+    assertEquals(Utils.bytesToDouble(bytes), 1, 0.0001);
+  }
+  
+  @Test (expectedExceptions = NullPointerException.class)
+  public void bytesToLongNull() throws Exception {
+    Utils.bytesToLong(null);
+  }
+  
+  @Test (expectedExceptions = IndexOutOfBoundsException.class)
```
busbey: Should we have a similar test for proper exceptions from the double method?

In support: the current javadocs say "must be 8 bytes long" but the code would allow e.g. a length 9 array.

core/src/test/java/com/yahoo/ycsb/TestUtils.java
```diff
@@ -0,0 +1,124 @@
...
+  public void bytesToLongTooShort() throws Exception {
+    Utils.bytesToLong(new byte[] { 0, 0, 0, 0, 0, 0, 0 });
+  }
+  
+  /**
+   * Since this version of TestNG doesn't appear to have an assertArrayEquals,
```
busbey: Ugh, really? We should make sure we have a ticket about moving to JUnit.

core/src/test/java/com/yahoo/ycsb/TestUtils.java
manolama: Yeah, sorry I missed that one, added the double test in there. 

And I updated the Javadocs about the array too long issue, wasn't sure if we should just throw or parse the first 8 so I'm parsing for now.

busbey:
thanks for the improvement!

PR #670, opened 2016-03-25
[rados] Development of RADOS binding of Ceph FS

jaemyoun:
Development of RADOS binding of Ceph File System
Even though Ceph FS is not NoSQL database, RADOS, Ceph utility, supports object-based storage.

For Ceph benchmark for cloud services. I implemented RADOS binding using [rados-java library](https://github.com/ceph/rados-java) expect the scan() function.


pom.xml
```diff
@@ -93,6 +93,8 @@ LICENSE file.
     <tarantool.version>1.6.5</tarantool.version>
     <aerospike.version>3.1.2</aerospike.version>
     <solr.version>5.4.0</solr.version>
+    <rados.version>0.3.0-SNAPSHOT</rados.version>
+    <json.version>20160212</json.version>
```
busbey: this should just be set in the pom for the rados client, since it needn't be shared across different datastore bindings..

rados/README.md
```diff
@@ -0,0 +1,64 @@
+<!--
+Copyright (c) 2014 - 2015 YCSB contributors. All rights reserved.
```
busbey: years need to be corrected.

rados/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+    ceph health
+
+### 2. Install Java and Maven
+
+### 3. Set Up YCSB
+
+Git clone YCSB and compile:
+
+    git clone -b dev-ceph-rados https://github.com/jaemyoun/YCSB.git
```
busbey: should probably just tell folks to download. Definitely should refer to the correct git repo if it stays as a clone.

rados/pom.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
busbey: needs a file header with licensing info

rados/src/main/java/com/yahoo/ycsb/db/RadosClient.java
```diff
@@ -0,0 +1,146 @@
+package com.yahoo.ycsb.db;
```
busbey: needs a file header with licensing info

busbey:
The travis failure is because you need to add the new module to the list of dependencies for the distribution module.

jaemyoun:
I fixed all code you mentioned, and an additional error.

rados/src/main/java/com/yahoo/ycsb/db/RadosClient.java
jaemyoun: fixed it

rados/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    </dependency>
+    <dependency>
+	    <groupId>org.json</groupId>
+		<artifactId>json</artifactId>
+		<version>${json.version}</version>
+	</dependency>
+  </dependencies>
+
+  <properties>
+  	<rados.version>0.3.0-SNAPSHOT</rados.version>
```
busbey: We don't rely on SNAPSHOT versions in our releases. Can we rely on a prior version? If not, when is 0.3.0 expected to come out?

jaemyoun:
I changed the version to 0.2.0 and merged from YCSB master for updating to 0.9.0-SNAPSHOT.

rados/src/main/java/com/yahoo/ycsb/db/RadosClient.java
```diff
@@ -0,0 +1,167 @@
...
+    if (id == null) {
+      id = ID_DEFAULT;
+    }
+
+    String pool = props.getProperty(POOL_PROPERTY);
+    if (pool == null) {
+      pool = POOL_DEFAULT;
+    }
+
+    rados = new Rados(id);
```
busbey: This needs to be protected by a lock, sine rados is static and we instantiate a client for each thread.

Presuming the rados and ioctx objects are themselves threadsafe, then you should synchronize and use a threadcount to decide when it's safe to do shutdown in the `cleanup` method.

busbey:
does rados provide any kind of testing utility jar that could be used to include a test?

rados/src/main/java/com/yahoo/ycsb/db/RadosClient.java
jaemyoun: Did you mean these to change Static Strings? The three value will be different depend on user setting.

busbey: No, I mean the member variables `rados` and `ioctx` are static. We instantiate a RadosClient per worker thread in the client, so access to initializing and cleaning up these static members must be protected by a lock.

If the member classes themselves aren't threadsafe, then you'll need to either guard all uses with a lock or switch to making them non-static.

jaemyoun:
I switched `rados` and `ioctx`to non-static, and implemented a test.
The test is success in my cluster, but the Travis CI building was failed because there is no the Ceph cluster.

busbey:
> The test is success in my cluster, but the Travis CI building was failed because there is no the Ceph cluster.

Does Ceph have some kind of artifact for making a local test instance that we could make use of? (see the `hbase10` or `accumulo` bindings for examples) If not,  we should update the test to make use of the JUnit `Assume` API to skip the test when a Ceph instance is not present (see the `mongodb` or `riak` bindings as an example).

jaemyoun:
> If not, we should update the test to make use of the JUnit Assume API to skip the test when a Ceph instance is not present (see the mongodb or riak bindings as an example).

I was not able to find the test instance for Ceph, so I updated to skip the test using `JUnit AssumeNoException`.

PR #673, opened 2016-03-26
[cassandra] properly clear cached containers.

### LINKED ISSUE #127: All subsequent Cassandra inserts fail after one insert fails due to an UnavailableException
This is the first time I am submitting an issue and my explanation may seem convoluted. I apologize before hand for the same. Feel free to edit the contents of this description.

There is an error in the Cassandra driver which causes all inserts to fail after one insert fails due to UnavailableException (the execption that occurs when not enough nodes are up to serve the write request with a consistency level which is not ANY). Please refer to the src file: cassandra/src/main/java/com/yahoo/ycsb/db/CassandraClient10.java

The problem is that the Cassandra client stores all the mutations that it needs to perform on the columns of one key as one operation in List<Mutation> mutations, Map<String, List<Mutation>> mutationMap and Map<ByteBuffer, Map<String, List<Mutation>>> record member variables (Line 89-91).   

After an insert is performed, these variable have to be cleared (Line 479-481). However if an exception is thrown in the client.batch_mutate(record, writeConsistencyLevel) call (Line 477) then these variables don't get cleared as the flow directly jumps to the corresponding catch block. 

Since these variables are not cleared, the next batch_mutate also causes all subsequent writes to fail because though it adds new writes to mutation variables, it still tries to perform the old failed mutation again and again which is not going to succeed as not enough nodes are up.

busbey:
clear containers before we use them rather than after, so that in the case of
exceptions we can recover.

closes #127

### DIFF
--- a/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraClient10.java
+++ b/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraClient10.java
@@ -453,6 +453,10 @@ public class CassandraClient10 extends DB {
     }
 
     for (int i = 0; i < operationRetries; i++) {
+      mutations.clear();
+      mutationMap.clear();
+      record.clear();
+
       if (debug) {
         System.out.println("Inserting key: " + key);
       }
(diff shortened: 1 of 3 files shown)


risdenk:
:+1:

PR #684, opened 2016-04-06
[elasticsearch] Elasticsearch improvements

jasontedor:
This pull request is a major tidying up of the Elasticsearch 
 - upgrade the dependency to Elasticsearch 2.3.1, the latest stable version
 - remove unneeded settings
 - fix/cleanup index creation
 - fix local/remote node settings
 - cleanup comments, docs, code
 - add settings for `number_of_shards`, `number_of_replicas`

These bindings work against an embedded local node and against a remote cluster and work on against *all* released versions of Elasticsearch in the 2.x series.

### DIFF
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -42,16 +42,13 @@ For further configuration see below:
 ### Defaults Configuration
 The default setting for the Elasticsearch node that is created is as follows:
 
-    cluster.name=es.ycsb.cluster
-    node.local=true
-    path.data=$TEMP_DIR/esdata
-    discovery.zen.ping.multicast.enabled=false
-    index.mapping._id.indexed=true
-    index.gateway.type=none
-    gateway.type=none
-    index.number_of_shards=1
-    index.number_of_replicas=0
+    es.cluster.name=es.ycsb.cluster
     es.index.key=es.ycsb
+    es.number_of_shards=1
+    es.number_of_replicas=0
+    es.remote=false
+    es.newdb=false
+    es.hosts.list=localhost:9200 (only applies if es.remote=true)
 
 ### Custom Configuration
 If you wish to customize the settings used to create the Elasticsearch node
@@ -66,25 +63,17 @@ pass it into the Elasticsearch client:
 
     ./bin/ycsb run elasticsearch -P workloads/workloada -P myproperties.data -s
 
-
-If you wish to use a in-memory store type rather than the default disk store add 
-the following properties to your custom properties file. For a large number of 
-insert operations insure that you have sufficient memory on your test system 
-otherwise you will run out of memory.
-
-    index.store.type=memory
-    index.store.fs.memory.enabled=true
-    cache.memory.small_buffer_size=4mb
-    cache.memory.large_cache_size=1024mb
-
 If you wish to change the default index name you can set the following property:
 
     es.index.key=my_index_key
 
-### Troubleshoot
-If you encounter error messages such as :
-"Primary shard is not active or isn't assigned is a known node."
+If you wish to run against a remote cluster you can set the following property:
 
-Try removing /tmp/esdata/ folder. 
-    rm -rf /tmp/esdata
+    es.remote=true
 
+By default this will use localhost:9300 as a seed node to discover the cluster.
+You can also specify
+
+    es.hosts.list=(\w+:\d+)+
+
+(a comma-separated list of host/port pairs) to change this.
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -42,7 +42,7 @@ For further configuration see below:
 ### Defaults Configuration
 The default setting for the Elasticsearch node that is created is as follows:
 
-    es.cluster.name=es.ycsb.cluster
+    cluster.name=es.ycsb.cluster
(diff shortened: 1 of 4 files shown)


jasontedor:
The build failure [here](https://travis-ci.org/brianfrankcooper/YCSB/jobs/121192619) is from Accumulo and has nothing to do with my changes.

busbey:
I relaunched the travis job.

busbey:
when you can, please include the module impacted at the start of your commit messages, e.g. "[elasticsearch] Cleanup code formatting"

elasticsearch/README.md
```diff
@@ -42,16 +42,13 @@ For further configuration see below:
...
-    cluster.name=es.ycsb.cluster
-    node.local=true
-    path.data=$TEMP_DIR/esdata
-    discovery.zen.ping.multicast.enabled=false
-    index.mapping._id.indexed=true
-    index.gateway.type=none
-    gateway.type=none
-    index.number_of_shards=1
-    index.number_of_replicas=0
+    es.cluster.name=es.ycsb.cluster
```
busbey: release manager that covers when this gets published should call this out as an incompatible change.

jasontedor:
> when you can, please include the module impacted at the start of your commit messages, e.g. "[elasticsearch] Cleanup code formatting"

@busbey Sure, sorry about that!

Would you like me to reword the commits and force push over?

elasticsearch/src/main/java/com/yahoo/ycsb/db/ElasticsearchClient.java
```diff
@@ -83,31 +83,25 @@
...
         .put("path.home", System.getProperty("java.io.tmpdir"));
 
     // if properties file contains elasticsearch user defined properties
     // add it to the settings file (will overwrite the defaults).
     settings.put(props);
-    System.out.println(
-        "Elasticsearch starting node = " + settings.get("cluster.name"));
-    System.out
-        .println("Elasticsearch node data path = " + settings.get("path.data"));
+    System.out.println("Elasticsearch starting node = " + settings.get("cluster.name"));
```
busbey: this property name got changed, right?

busbey: could we make these diagnostic outputs go to System.err instead of System.out?

jasontedor: Yes, but note that this is reading from the `settings` object where it is defined as `cluster.name`, not from the properties where it is named `es.cluster.name`.

jasontedor: > could we make these diagnostic outputs go to System.err instead of System.out?

Sure, I'll push a commit that does that later tonight.

busbey:
> > when you can, please include the module impacted at the start of your commit messages, e.g. "[elasticsearch] Cleanup code formatting"
> @busbey Sure, sorry about that!
> 
> Would you like me to reword the commits and force push over?

It'd be nice, but I don't consider it a blocker.

elasticsearch/src/main/java/com/yahoo/ycsb/db/ElasticsearchClient.java
busbey: > > this property name got changed, right?
> Yes, but note that this is reading from the settings object where it is defined as cluster.name, not from the properties > where it is named es.cluster.name.

Right, but right before we overwrote the entries in the settings file from the properties file via `settings.put(props)`. Do we have to worry about not properly overwriting the cluster name that way?

jasontedor:
> It'd be nice, but I don't consider it a blocker.

I'll do it in a few hours.

jasontedor:
> I'll do it in a few hours.

I force pushed commits that reword the commit messages to have the prefix "[elasticsearch]".

elasticsearch/src/main/java/com/yahoo/ycsb/db/ElasticsearchClient.java
jasontedor: > > could we make these diagnostic outputs go to System.err instead of System.out?

> Sure, I'll push a commit that does that later tonight.

I pushed bcf52844c9869145273b0a368c6dd436c8ae31c3.

jasontedor: > Right, but right before we overwrote the entries in the settings file from the properties file via `settings.put(props)`. Do we have to worry about not properly overwriting the cluster name that way?

In this case it's correct because earlier the property `es.cluster.name` was extracted from the `props` object and used to associate the value of this property with the key `cluster.name` in the `settings` object.

That said, I think this can be done in a more straightforward way. I pushed 6eb53b80e212c78fb4df3583263fc8326ab4f45a.

busbey:
Since this became a comment on an old version, reposting for visibility:

> release manager that covers when this gets published should call this out as an incompatible change

It's because configuration parameters changing names. Since we haven't had Elasticsearch tested in a release before, this should have minimal impact.

jasontedor:
Thanks @busbey!

PR #686, opened 2016-04-06
[riak] Added new binding for Riak, based on the one created by Basho Technol…

nygard89:
Added new binding for Riak, based on the one created by Basho Technologies (available at https://github.com/basho-labs/YCSB/tree/master/riak).

Changes made to the original version:

- Made compatible with the latest YCSB version.
- Removed unused Constants.java file.
- Added strong consistency test support.
- Done some refinements to the code.
- Removed many unused function in the RiakUtils.java file.
- Choosed to use executeAsync methods to resolve a bug that didn't allow the benchmark to end whenever one or more nodes of the cluster are killed.
- Added TIME_OUT status return code.

To do:
- Add different LOG errors (i.e. insufficient_vnodes_available, ring_not_ready, etc.) - for references, check page http://docs.basho.com/riak/latest/ops/running/recovery/errors/#Errors-and-Messages
- ...


pom.xml
```diff
@@ -123,9 +123,11 @@ LICENSE file.
     <module>nosqldb</module>
     <module>orientdb</module>
     <module>redis</module>
+    <module>riak</module>
     <module>s3</module>
     <module>solr</module>
     <module>tarantool</module>
+    <module>distribution</module>
```
busbey: Please do not reorder unrelated modules.

riak/src/main/java/com/yahoo/ycsb/db/RiakKVClient.java
```diff
@@ -0,0 +1,493 @@
+/*
+ * Copyright 2016 nygard_89
```
busbey: We try to use a blanket of "YCSB contributors" for the changes that come directly to the project. Generally we then rely on git to provide us with who made a given line.

I agree that we need to leave the Basho line in, similar to how we handle the original Yahoo! Lines. Would you mind changing the bit with your GitHub handle to instead be the contributors reference?

busbey:
You'll need to update the distribution module to list riak as a dependency.

busbey:
Please update the commit message to start with the mobile, i.e. "[riak] add support for riak..."

riak/README.md
```diff
@@ -0,0 +1,51 @@
...
+* `riak.port` - <b>int</b>, the port on which every node is listening. It must match the one specified in the `riak.conf` file at the line `listener.protobuf.internal`.
+* `riak.bucket_type` - <b>string</b>, it must match the value of the bucket type created during setup (see section above).
+* `riak.r_val` - <b>int</b>, the R value represents the number of Riak nodes that must return results for a read before the read is considered successful.
+* `riak.w_val` - <b>int</b>, the W value represents the number of Riak nodes that must report success before an update is considered complete.
+* `riak.read_retry_count` - <b>int</b>, the number of times the client will try to read a key from Riak.
+* `riak.wait_time_before_retry` - <b>int</b>, the time (in milliseconds) before client attempts to perform another read if the previous one failed.
+* `riak.transaction_time_limit` - <b>int</b>, the time (in seconds) the client waits before aborting the current transaction.
+* `riak.strong_consistency` - <b>boolean</b>, indicates whether to use strong consistency (true) or eventual consistency (false).
```
busbey: Please default to strong consistency

riak/src/main/java/com/yahoo/ycsb/db/RiakKVClient.java
```diff
@@ -0,0 +1,493 @@
...
+    if (debugString != null) {
+      debug = Boolean.parseBoolean(debugString);
+    }
+  }
+
+  public void init() throws DBException {
+    loadProperties();
+
+    if (debug) {
+      System.out.println("DEBUG ENABLED. Configuration parameters:");
```
busbey: Please write these to stderr

riak/src/main/java/com/yahoo/ycsb/db/RiakKVClient.java
```diff
@@ -0,0 +1,493 @@
...
+import com.basho.riak.client.core.query.RiakObject;
+import com.basho.riak.client.core.query.indexes.LongIntIndex;
+import com.basho.riak.client.core.util.BinaryValue;
+
+import static com.yahoo.ycsb.db.RiakUtils.getKeyAsLong;
+import static com.yahoo.ycsb.db.RiakUtils.serializeTable;
+
+
+/**
+ * @author nygard_89
```
busbey: Please rely on git for tracking contribution instead of author tags.

riak/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The YCSB binding for <a href="http://basho.com/products/riak-kv/">Riak KV</a>
+ * 2.0+.
+ */
+package com.yahoo.ycsb.db;
```
busbey: Could we put this in a package named `com.yahoo.ycsb.db.riak`? I'm trying to get the datastore bindings more separated.

riak/src/main/resources/riak.properties
```diff
@@ -0,0 +1,39 @@
+# RiakDBClient - Default Properties
```
busbey: Need a copyright / license header

busbey:
Does Riak have any testing utilities we could you to make some tests? Either via mocking or a local test instance?

nygard89:
I wrote a new local test instance (RiakKVClientTest.java), it works fine for me.

riak/src/main/java/com/yahoo/ycsb/db/package-info.java
nygard89: It's okay for me.

riak/README.md
nygard89: Done, but I previously did not because of Riak's design, as it is mainly an eventually consistent db.

busbey:
I don't see RiakKVClientTest in the patch. Maybe a missed `git add`?

nygard89:
Yep, my bad. Now it should be good to go!

nygard89:
Riak module is now ready to be used. I've tested it with different workloads/configuration on a local cluster without any issue.

busbey:
I think the mongodb binding has a test that will work with a local cluster (and skip it if there isn't one). It also has Travis run a mongo instance for doing the test. Looks like we can do that for riak as well: https://docs.travis-ci.com/user/database-setup/#Riak

risdenk:
Should this have checkstyle in the pom.xml?

nygard89:
There actually was an issue with __strong-consistent scan transactions__, which weren't performed at all (they merely returned with `Status.OK` but actually no read was done) due to the lack of support for secondary indexing when using _strong consistency model_. I'm trying to further investigate the matter but, as of the moment, there seems to be no easy solution for implementing such transactions for this consistency model (further info available at http://docs.basho.com/riak/kv/2.1.4/developing/app-guide/strong-consistency/). 
So, for now, I've decided to let the operation immediately return with a `Status.NOT_IMPLEMENTED` whenever one tries to perform a scan transaction using strong consistency model. Moreover, I've decided to set _eventual consistency model_ as default.

Finally, I've configured Travis to start a Riak KV instance to allow the execution of the testing units, which all run correctly.

.travis.yml
```diff
@@ -27,10 +27,15 @@ install: mvn install -q -DskipTests=true
 
 script: mvn test -q
 
+# Bucket-type to activate before running Riak KV tests.
```
risdenk: Can the Riak test class create these if they don't exist? Instead of doing it outside of the tests and depending on changing travis.yml?

nygard89: Actually, the only way to create a bucket-type is via riak-admin commands, so it's not possible. However, I think I could modify the test class to execute those commands from within there, but I would at least keep the `sudo: required` line as it is. This is because the aforementioned commands need to be executed with `sudo`.

risdenk: Does a new bucket type need to be created? Is there a default bucket type that can be used for tests?

.travis.yml
```diff
@@ -27,10 +27,15 @@ install: mvn install -q -DskipTests=true
...
 # Services to start for tests.
 services:
   - mongodb
+  - riak
 
-
-# Use the Container based infrastructure.
-sudo: false
+# Sudo required for Riak KV test.
+sudo: required
```
busbey: ugh. this will put us on the non-container infra. Are we sure the travis-ci user isn't allowed to run riak-admin?

nygard89:
@busbey @risdenk Again, my bad. There's actually no need to let Travis do all that stuff to allow a proper Riak KV testing. There is a default bucket type that can be adequately modified for testing purposes.

busbey:
> @busbey @risdenk Again, my bad. There's actually no need to let Travis do all that stuff to allow a proper Riak KV testing. There is a default bucket type that can be adequately modified for testing purposes.

Oh sweet. Thanks for following up on it!

busbey:
> So, for now, I've decided to let the operation immediately return with a Status.NOT_IMPLEMENTED whenever one tries to perform a scan transaction using strong consistency model. Moreover, I've decided to set eventual consistency model as default.

Can we keep this work-around in place but default to strong consistency? It's already called out well in the README. If you're concerned about folks getting lost because of the NOT_IMPLEMENTED responses happening by default, we could perhaps add a log message on initialization if we see that the workload has scans and we're in strict mode.

nygard89:
Now strong consistency is the default option for the benchmarks!

busbey:
this looks good to me. I'd like to verify that the tests get properly skipped when not run on travis. I should be able to do that this evening.

@risdenk if you have a chance before then and you want to merge, feel free to. (to your earlier checkstyle question, the default now is that all modules get checkstyle run against them. those that have errors rely on a property to tell checkstyle not to fail them. e.g. core/pom.xml has an example.)

busbey:
thanks for taking care of this! If there are any more improvements you'd like to do, please open a new pull request.

We'll probably ping you about testing when we start prep work for our next release.

nygard89:
I think I might have find a good compromise solution for allowing strong-consistent `scan` transactions, but it would be nice to discuss it with you before opening a new pull request. 

So, the idea I had was the following: considering that Riak allows to utilize _secondary indexes_ only for eventually-consistent bucket-types, I've created one of these (called `fakeBucketType`) just to exploit this property. In practice, I've modified the `insert` function to make sure that, beside inserting the actual record to the proper bucket-type (which is the one called `ycsb` by default), a 1-byte record (which is the minimum allowed size) **with the same key as the latter** was stored in the `fakeBucketType` just to keep track of its secondary index. Then, for the scan transactions, I've changed the source of the secondary indexes search operation to `fakeBucketType` to retrieve the key paired with the index. This latter is then used to read the actual object searched (which, again, is stored in the `ycsb` bucket-type). And this is all... I hope I've explained myself!

Ah, just one note: I tried to keep this `fakeBucketType` as less resource-demanding as possible, by disabling all the consistency mechanisms that Riak provides (we don't need them: this "fake object" will be inserted just once, so there's no need to keep much metadata for it).

busbey:
So long as there's a note about the approach and the tradeoff it involves in the Riak module's README, that sounds fine.

cmatser:
I think this is a great feature of measurement to add to the Riak binding.  However, can this be implemented as a configurable option?  I'm concerned about the performance impact on `insert` in order to provide the benefits for __string-consistent scan transactions__.  It would be great if it were configurable so that the performance difference could be measured and also so that people who are not concerned with the `scan` feature are not negatively impacted.  Thanks for the work on this, btw.

nygard89:
@cmatser There are indeed performance drops when using this workaround. This is why I made it configurable and not set it as the default option. I've opened a new pull request a few days ago with these modifications.

PR #689, opened 2016-04-08
[hbase10] Add instructions for testing against Google BigTable's HBase API.

manolama:
Add documentation on how to use the hbase10 client with Google's
Bigtable.

### DIFF
--- a/hbase10/README.md
+++ b/hbase10/README.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2015 YCSB contributors. All rights reserved.
+Copyright (c) 2015-2016 YCSB contributors. All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License"); you
 may not use this file except in compliance with the License. You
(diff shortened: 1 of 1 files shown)


hbase10/README.md
```diff
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2015 YCSB contributors. All rights reserved.
+Copyright (c) 2016 YCSB contributors. All rights reserved.
```
busbey: should be 2015-2016

hbase10/README.md
```diff
@@ -16,8 +16,101 @@ LICENSE file.
...
+
+Make a note of the column family, in this example it's `cf``.
+
+### 4. Fetch the Proper ALPN Boot Jar
+
+The Bigtable protocol uses HTTP/2 which requires an ALPN protocol negotiation implementation. On JVM instantiation the implementation must be loaded before attempting to connect to the cluster. If you're using Java 7 or 8, use this [Jetty Version Table](http://www.eclipse.org/jetty/documentation/current/alpn-chapter.html#alpn-versions) to determine the version appropriate for your JVM. (ALPN is included in JDK 9+). Download the proper jar from [Maven](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.mortbay.jetty.alpn%22%20AND%20a%3A%22alpn-boot%22) somewhere on your system.
+
+### 5. Download the Bigtable Client Jar
+
+Download one of the `bigtable-hbase-1.#` jars from [Maven](http://search.maven.org/#search%7Cga%7C1%7Ccom.google.cloud.bigtable) to your host.
```
busbey: this step makes me think this ought to be an additional binding, e.g. `hbase-bigtable`, but that should probably be evaluated as a follow on after #695

manolama: @cmccoy from Google preferred to use this binding to reduce code duplication. They've been using it that way in house apparently.

busbey: we wouldn't need to duplicate any code, just an additional packaging-only module with the different dependencies. We do the same thing right now for the `hbase094` module, which reuses the implementations in the `hbase098` module.

PR #690, opened 2016-04-08
[googlebigtable] Add a Google Bigtable binding for testing load using

manolama:
the native Protobuf GRPC based API.

### DIFF
--- a/googlebigtable/README.md
+++ b/googlebigtable/README.md
@@ -1,0 +1,80 @@
+<!--
+Copyright (c) 2016 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+# Google Bigtable  Driver for YCSB
+
+This driver provides a YCSB workload binding for Google's hosted Bigtable, the inspiration for a number of key-value stores like HBase and Cassandra. The Bigtable Java client provides both Protobuf based GRPC and HBase client APIs. This binding implements the Protobuf API for testing the native client. To test Bigtable using the HBase API, see the `hbase10` binding.
+
+## Quickstart
+
+### 1. Setup a Bigtable Cluster
+
+Login to the Google Cloud Console and follow the [Creating Cluster](https://cloud.google.com/bigtable/docs/creating-cluster) steps. Make a note of your cluster name, zone and project ID.
+
+### 2. Launch the Bigtable Shell
+
+From the Cloud Console, launch a shell and follow the [Quickstart](https://cloud.google.com/bigtable/docs/quickstart) up to step 4 where you launch the HBase shell.
+
+### 3. Create a Table
+
+For best results, use the pre-splitting strategy recommended in [HBASE-4163](https://issues.apache.org/jira/browse/HBASE-4163):
+
+```
+hbase(main):001:0> n_splits = 200 # HBase recommends (10 * number of regionservers)
+hbase(main):002:0> create 'usertable', 'cf', {SPLITS => (1..n_splits).map {|i| "user#{1000+i*(9999-1000)/n_splits}"}}
+```
+
+Make a note of the column family, in this example it's `cf``.
+
+### 4. Fetch the Proper ALPN Boot Jar
+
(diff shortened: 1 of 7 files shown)


manolama:
May need someone with more Maven experience to help me out. Despite adding the module lexically in the parent pom, it doesn't seem to build the binding package or run clean in lexical order. Anyone have an idea what I'm missing? Thanks!

googlebigtable/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+hbase(main):001:0> n_splits = 200 # HBase recommends (10 * number of regionservers)
+hbase(main):002:0> create 'usertable', 'cf', {SPLITS => (1..n_splits).map {|i| "user#{1000+i*(9999-1000)/n_splits}"}}
+```
+
+Make a note of the column family, in this example it's `cf``.
+
+### 4. Fetch the Proper ALPN Boot Jar
+
+The Bigtable protocol uses HTTP/2 which requires an ALPN protocol negotiation implementation. On JVM instantiation the implementation must be loaded before attempting to connect to the cluster. If you're using Java 7 or 8, use this [Jetty Version Table](http://www.eclipse.org/jetty/documentation/current/alpn-chapter.html#alpn-versions) to determine the version appropriate for your JVM. (ALPN is included in JDK 9+). Download the proper jar from [Maven](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.mortbay.jetty.alpn%22%20AND%20a%3A%22alpn-boot%22) somewhere on your system.
```
busbey: Could we take care of this dependency by using jdk profiles in our Maven pom?

If the needed jar is in Maven Central I can take a stab at it in a follow-on, provided someoen can test for me.

googlebigtable/src/main/java/com/yahoo/ycsb/db/GoogleBigtableClient.java
```diff
@@ -0,0 +1,443 @@
...
+  /** Cache for the last table name/ID to avoid byte conversions. */
+  private String lastTable = "";
+  private byte[] lastTableBytes;
+  
+  /**
+   * If true, buffer mutations on the client. This is the default behavior for
+   * HBaseClient. For measuring insert/update/delete latencies, client side
+   * buffering should be disabled.
+   */
+  private boolean clientSideBuffering = true;
```
busbey: We should be defaulting to no client side buffering. We should correct any hbase clients doing this.

googlebigtable/src/main/java/com/yahoo/ycsb/db/GoogleBigtableClient.java
```diff
@@ -0,0 +1,443 @@
...
+    final Iterator<Entry<Object, Object>> it = props.entrySet().iterator();
+    while (it.hasNext()) {
+      Entry<Object, Object> entry = it.next();
+      CONFIG.set((String)entry.getKey(), (String)entry.getValue());
+    }
+    
+    clientSideBuffering = getProperties().getProperty(CLIENT_SIDE_BUFFERING, "true")
+        .equals("true") ? true : false;
+    
+    System.out.println("Running Google Bigtable with Proto API" +
```
busbey: Can this go to stderr?

googlebigtable/README.md
manolama: Can jdk profiles be range based or do you have to setup a profile for a full major and minor version of the JDK? Another option might be to have Maven run a script to detect the JDK version then inject the proper ALPN version. If you could take a stab I'd be happy to test it.

googlebigtable/src/main/java/com/yahoo/ycsb/db/GoogleBigtableClient.java
manolama: wilco

googlebigtable/README.md
busbey: I think they can be range based. In the past I've just done them for major versions (e.g JDK7, JDK8, etc). I'll filed #695 to remind myself to take a look.

googlebigtable/README.md
```diff
@@ -0,0 +1,80 @@
...
+### 5. Download JSON Credentials
+
+Follow these instructions for [Generating a JSON key](https://cloud.google.com/bigtable/docs/installing-hbase-shell#service-account) and save it to your host.
+
+### 6. Load a Workload
+
+Switch to the root of the YCSB repo and choose the workload you want to run and `load` it first. With the CLI you must provide the column family, cluster properties and the ALPN jar to load.
+
+```
+bin/ycsb load googlebigtable -p columnfamily=cf -p google.bigtable.project.id=<PROJECT_ID> -p google.bigtable.cluster.name=<CLUSTER> -p google.bigtable.zone.name=<ZONE> -p google.bigtable.auth.service.account.enable=true -p google.bigtable.auth.json.keyfile=<PATH_TO_JSON_KEY> -jvm-args='-Xbootclasspath/p:<PATH_TO_ALPN_JAR>' -P workloads/workloada
```
busbey: Interesting. Does the ALPN jar need to be on bootclasspath rather than the normal classpath? (accessible by giving `-cp <PATH_TO_ALPN_JAR>` to the ycsb wrapper)

manolama: Yeah, it has to be in the boot path as it does some funky injection, hence the requirement of different versions of the JAR depending on minor versions of the JDK. Blech.

busbey: well that will make #695 interesting then. :)

Issue #691, opened 2016-04-08
[hbase10] Still too many threads 


saintstack:
#662  and #651 addressed one of causes of too-many-threads by checking if the static Connection instance were null or not. They helped.

For case where many threads starting up at the same time, the concurrency 'protection' around the Connection check is broke so we'll spin up many more connections than were asked for.

Here is total threads when I specify a thread count of 200:

stack@ve0524:~$ jstack 114968 |grep tid=|wc -l
1144

Adding a lock around the Connection check gets me this total on start.

stack@ve0524:~$ jstack 118574|grep tid=|wc -l
476

Instrumenting the ConnectionFactory used by the 1.0.2 hbaseclient10, I can see us making more than a single instance of the cluster connection Connection object:

```
   6 2016-04-08 15:51:07:313 0 sec: 0 operations; est completion in 0 seconds
   7 Maximum execution time specified as: 3600 secs
   8 2016-04-08 15:51:07,435 WARN  [Thread-4] util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
   9 2016-04-08 15:51:07,705 INFO  [Thread-16] client.ConnectionFactory: Created connection with class class org.apache.hadoop.hbase.client.ConnectionManager$HConnectionImplementation, user=stack (auth:SIMPLE), managed=false, pool=null
  10 java.lang.RuntimeException: Exception just to show where we are called form!
  11   at org.apache.hadoop.hbase.client.ConnectionFactory.createConnection(ConnectionFactory.java:241)
  12   at org.apache.hadoop.hbase.client.ConnectionFactory.createConnection(ConnectionFactory.java:221)
  13   at org.apache.hadoop.hbase.client.ConnectionFactory.createConnection(ConnectionFactory.java:122)
  14   at com.yahoo.ycsb.db.HBaseClient10.init(HBaseClient10.java:139)
  15   at com.yahoo.ycsb.DBWrapper.init(DBWrapper.java:77)
  16   at com.yahoo.ycsb.ClientThread.run(Client.java:289)
  17 2016-04-08 15:51:07,717 INFO  [Thread-135] client.ConnectionFactory: Created connection with class class org.apache.hadoop.hbase.client.ConnectionManager$HConnectionImplementation, user=stack (auth:SIMPLE), managed=false, pool=null
  18 java.lang.RuntimeException: Exception just to show where we are called form!
[...]

PR #692, opened 2016-04-08
[hbase10] Still too many threads #691.

saintstack:
Fix broken synchronization meant to guard against over-creation
of cluster connection instances on startup when multiple threads.

### DIFF
--- a/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
+++ b/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
@@ -51,6 +51,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.Vector;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * HBase 1.0 client for YCSB framework.
@@ -63,14 +64,22 @@ import java.util.Vector;
  */
 public class HBaseClient10 extends com.yahoo.ycsb.DB {
   private Configuration config = HBaseConfiguration.create();
-  
-  // Must be an object for synchronization and tracking running thread counts. 
-  private static Integer threadCount = 0;
+
+  private static AtomicInteger threadCount = new AtomicInteger(0);
 
   private boolean debug = false;
 
   private String tableName = "";
+
+  /**
+   * A Cluster Connection instance that is shared by all running ycsb threads.
+   * Needs to be initialized late so we pick up command-line configs if any.
+   * To ensure one instance only in a multi-threaded context, guard access
+   * with a 'lock' object.
+   * @See #CONNECTION_LOCK.
+   */
   private static Connection connection = null;
+  private static final Object CONNECTION_LOCK = new Object();
 
   // Depending on the value of clientSideBuffering, either bufferedMutator
   // (clientSideBuffering) or currentTable (!clientSideBuffering) will be used.
@@ -121,10 +130,10 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB {
       UserGroupInformation.setConfiguration(config);
     }
 
-    if ((getProperties().getProperty("principal")!=null) 
+    if ((getProperties().getProperty("principal")!=null)
         && (getProperties().getProperty("keytab")!=null)) {
       try {
-        UserGroupInformation.loginUserFromKeytab(getProperties().getProperty("principal"), 
+        UserGroupInformation.loginUserFromKeytab(getProperties().getProperty("principal"),
               getProperties().getProperty("keytab"));
       } catch (IOException e) {
         System.err.println("Keytab file is not readable or not found");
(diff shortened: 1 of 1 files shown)


saintstack:
#691 has write up on what this pull request addresses

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -166,7 +176,9 @@ public void init() throws DBException {
     String table = com.yahoo.ycsb.workloads.CoreWorkload.table;
     try {
       final TableName tName = TableName.valueOf(table);
-      connection.getTable(tName).getTableDescriptor();
```
manolama: Ah so ``getTable()`` isn't thread safe in that it could create more connection objects?

busbey: I thought Connection was supposed to be threadsafe? Is this essentially a bug in the implementation? Any idea if it still exists in later HBase client versions?

I'm happy to demonstrate a workaround other applications need to be aware of, but I'd like to make sure we document why we're being more conservative than the HBase docs indicate we need to be.

saintstack:
@manolama The root issue is that when lots of threads starting up, they all see the static 'connection' as null so each creates a new one but only one of which prevails... meantime a bunch of unreferenced connections are just hanging out (along w/ their zk client threads). The synchronization here on connection.getTable is just my being consistent synchronizing all use of connection so threads see the connection probably made by another and don't get a NPE on connection; probably gratuitous (getTable is safe).

@busbey connection is thread safe. This 'fix' is to ycsb HBaseClient trying to have a singleton instance of a connection. It was synchronizing on an Integer that it was then changing inside the synchronize block.... and the synchronization on the threadCount integer was supposed to be 'protecting' connection initialization. Its broken.

busbey:
oh! the autoboxing of the integer changed the reference. of course. /me slaps forehead.

guess next time I won't advocate moving to something "simpler" than the Atomic classes. :/

saintstack:
On Sat, Apr 9, 2016 at 10:40 PM, Sean Busbey <notifications@github.com>
wrote:

> oh! the autoboxing of the integer changed the reference. of course. /me
> slaps forehead.
>
> guess next time I won't advocate moving to something "simpler" than the
> Atomic classes. :/
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/brianfrankcooper/YCSB/pull/692#issuecomment-207922382>
>
Looking at hbase098, it is synchronizing on the AtomicInteger which would
do-the-right-thing. Should I make this patch do same as it?

busbey:
> The synchronization here on connection.getTable is just my being consistent synchronizing all use of connection so threads see the connection probably made by another and don't get a NPE on connection; probably gratuitous (getTable is safe).

This part shouldn't be needed, since we are synchronized within init and cleanup on changing the null/not-null state of connection.

If we leave the synchronized block in here, then we'll contend on the lock across all the threads whenever the table changes. With the current core workloads that will mean the first operation each thread does. Maybe not killer since just at startup, but it will be ugly since it'll be each thread's init + each threads first operation all hitting the same mutex.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -63,14 +64,22 @@
  */
 public class HBaseClient10 extends com.yahoo.ycsb.DB {
   private Configuration config = HBaseConfiguration.create();
-  
-  // Must be an object for synchronization and tracking running thread counts. 
-  private static Integer threadCount = 0;
+
+  private static AtomicInteger threadCount = new AtomicInteger(0);
```
busbey: should be final.

busbey:
> Looking at hbase098, it is synchronizing on the AtomicInteger which would
do-the-right-thing. Should I make this patch do same as it?

That's what this client did before #651. I don't really mind either way. I suppose it's nice for the implementations to be similar?

cmatser:
:+1:

manolama:
@saintstack Ah thank you! That damn autoboxing got me so it makes sense now.

cmatser:
I'm tracking this issue for the release. Can you tell me what's the next step(s) before this can be merged? Thanks!

saintstack:
I'm running the patch locally. Works for me.  It will be different locking from what is in the hbase098 and hbase094 clients but it was different before I showed up. There is one +1 above. Need more?

busbey:
the threadcount variable should be final. but I can clean that up and make things look like hbase098 in a follow-on.

busbey:
shoot. just reread things more carefully and saw my comment about synchronizing around getTable.

probably fine? I'll add it to the follow-on. someone let me know if they think we should revert and fix now.

saintstack:
Fine I'd say @busbey ... just startup. And we're synchronizing anyways.

Thanks for merge.

PR #693, opened 2016-04-09
[elasticsearch] Fix race condition

jasontedor:
This commit fixes a race condition that exists between the shards being
ready to serve requests and the execution of the workload.


jasontedor:
Thanks @busbey!

PR #694, opened 2016-04-09
[core] Add an incrementing printable string generator.

manolama:
This will be used for #653 for databases that require keys or other values to use a limited character set. This generator will create a sequence of strings using a defined list of UTF code points from which to pull.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/generator/IncrementingPrintableStringGenerator.java
+++ b/core/src/main/java/com/yahoo/ycsb/generator/IncrementingPrintableStringGenerator.java
@@ -1,0 +1,389 @@
+/**
+ * Copyright (c) 2016 YCSB contributors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+package com.yahoo.ycsb.generator;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+/**
+ * A generator that produces strings of {@link #length} using a set of code points
+ * from {@link #characterSet}. Each time {@link #nextValue()} is executed, the string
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/yahoo/ycsb/generator/IncrementingPrintableStringGenerator.java
```diff
@@ -0,0 +1,388 @@
...
+    }
+    this.length = length;
+    this.characterSet = characterSet;
+    indices = new int[length];
+  }
+  
+  @Override
+  public String nextValue() {
+    if (hasRolledOver && throwExceptionOnRollover) {
+      throw new RuntimeException("The generator has rolled over to the beginning");
```
busbey: could we make this a subclass of RuntimeException so that if folks want to catch it for whatever reason they needn't parse the message.

alternatively, maybe we should add an accessor for hasRolledOver in case a user need to know if things have rolled over, but doesn't want to consider it an error?

core/src/test/java/com/yahoo/ycsb/generator/TestIncrementingPrintableStringGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+        new IncrementingPrintableStringGenerator(2, ATOC);
+    gen.setThrowExceptionOnRollover(true);
+    
+    int i = 0;
+    try {
+      while(i < 11) {
+        ++i;
+        gen.nextValue();
+      }
+      fail("Expected RuntimeException");
```
busbey: does testng not have an equivalent to junit's `@Test(expected=SomeException.class)` annotation?

core/src/main/java/com/yahoo/ycsb/generator/IncrementingPrintableStringGenerator.java
manolama: Ah good point. I'll change it to the built-in ``NoSuchElementException``.

core/src/test/java/com/yahoo/ycsb/generator/TestIncrementingPrintableStringGenerator.java
manolama: It does have ``expectedException`` but in these tests I wanted to make sure the counter was the proper value after the exception was thrown.

busbey: Ah I see. That makes sense. These should check for the updated NoSuchElementException.

It looks like the checks for IllegalArgumentException below could switch to `expectedException`.

manolama: doh, good catch. Updated.

manolama:
Arg Cass CQL UT failed.

busbey:
restarted the travis job.

busbey:
there we go! thanks @manolama.

PR #696, opened 2016-04-10
[core] Add CoreWorkload.createOperationGenerator() for use in other w…

manolama:
…orkloads.

Eventually it would be nice to move this out into a proper generator with enums to
save space, but as users may have extended CoreWorkload and could be using the
"operationchooser" field, we'll just break it out into a static function so as to
avoid breaking anyone's code.


busbey:
thanks!

PR #698, opened 2016-04-10
[googlebigtable] Fix #697 by synchronizing on the static Config object

### LINKED ISSUE #697: [googlebigtable] race condition in initialization caused by autoboxing
The googlebigtable module has the same problem synchronizing init as hbase10 does in #692

manolama:
who's address does not change.


busbey:
thanks for the fix!

PR #711, opened 2016-04-15
[couchbase2] Allow to log latency & runtime metrics.

daschl:
This changeset adds 2 config params which can be tweaked to enable
and print both the network latency and runtime metrics to the
configured logger.

### DIFF
--- a/couchbase2/README.md
+++ b/couchbase2/README.md
@@ -91,6 +91,26 @@ to just "drive load" and disable the waiting. Note that when the "-p couchbase.s
 used, the measured results by YCSB can basically be thrown away. Still helpful sometimes during load phases to speed
 them up :)
 
+## Debugging Latency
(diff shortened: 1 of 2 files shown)


busbey:
excellent. is this using hdrhistogram under the covers, or something else?

daschl:
@busbey indeed, we utilize HdrHistogram inside the underlying `core-io` library. Note that this functionality is built into the client all the time and we have customers using this to feed it into monitoring systems. I thought it might be super helpful when you are running YCSB and are trying to figure out what the lower level latency stats look like :)

busbey:
yeah, that'll be great. it'd also be nice if you could publish using the hdr historgram export format instead of JSON so that folks could use the combination of the export from YCSB to do analysis at arbitrary time buckets.

PR #714, opened 2016-04-16
[asynchbase] Add AsyncHBase as an alternative to the default HBase cl…

manolama:
…ient.

### DIFF
--- a/asynchbase/README.md
+++ b/asynchbase/README.md
@@ -1,0 +1,59 @@
+<!--
+Copyright (c) 2016 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+# AsyncHBase Driver for YCSB
+
+This driver provides a YCSB workload binding for Apache HBase using an alternative to the included HBase client. AsyncHBase is completely asynchronous for all operations and is particularly useful for write heavy workloads. Note that it supports a subset of the HBase client APIs but supports all public released versions of HBase.
+
+## Quickstart
+
+### 1. Setup Hbase
+
+Follow directions 1 to 3 from ``hbase098``'s readme.
+
+### 2. Load a Workload
+
+Switch to the root of the YCSB repo and choose the workload you want to run and `load` it first. With the CLI you must provide the column family at a minimum if HBase is running on localhost. Otherwise you must provide connection properties via CLI or the path to a config file. Additional configuration parameters are available below.
+
+```
+bin/ycsb load asynchbase -p columnfamily=cf -P workloads/workloada
+
+```
+
+The `load` step only executes inserts into the datastore. After loading data, run the same workload to mix reads with writes.
+
+```
+bin/ycsb run asynchbase -p columnfamily=cf -P workloads/workloada
+
+```
+
+## Configuration Options
+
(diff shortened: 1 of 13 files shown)


busbey:
Could we use one of the hbase minicluster / test utility to add tests, similar to how we do on the hbase client?

asynchbase/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+bin/ycsb run asynchbase -p columnfamily=cf -P workloads/workloada
+
+```
+
+## Configuration Options
+
+The following options can be configured using CLI (using the `-p` parameter) or via a JAVA style properties configuration file.. Check the [AsyncHBase Configuration](http://opentsdb.github.io/asynchbase/docs/build/html/configuration.html) project for additional tuning parameters.
+
+* `columnfamily`: (Required) The Bigtable column family to target.
```
busbey: this is HBase column family, right? Or maybe just say "column family"

asynchbase/src/main/java/com/yahoo/ycsb/db/AsyncHBaseClient.java
```diff
@@ -0,0 +1,410 @@
...
+        Thread.currentThread().interrupt();
+      } catch (Exception e) {
+        System.err.println("Failure reading from row with key " + key + 
+            ": " + e.getMessage());
+        return Status.ERROR;
+      }
+    } else {
+      // hooray! Asynchronous write. But without a callback and an async
+      // YCSB call we don't know whether it succeeded or not
+      client.put(put);
```
busbey: If the operations used are asynchronous then our latency measurements won't mean much, right? Can we call that out in the README?

busbey:
this is excellent, by the way. I've wanted an easy way to switch between asynchbase and the normal client for server load generation.

asynchbase/src/main/java/com/yahoo/ycsb/db/AsyncHBaseClient.java
manolama: I set it up to mimic the HBase client so that if ``clientSideBuffering`` is not set, it behaves in a synchronous manner with the ``.join()`` on each operation. Otherwise if they want buffering, it is async and mimics using the ``BufferedMutator``.

asynchbase/README.md
manolama: Doh, copy-pasta got me ;)

manolama:
Cool, I'll play with the minicluster a bit and get those UTs in.

asynchbase/src/main/java/com/yahoo/ycsb/db/AsyncHBaseClient.java
busbey: Sounds good. I'll try to read more carefully next time. :)

busbey:
Do you prefer waiting on this PR until the unit tests are in place, or adding them as a follow-on?

manolama:
Just added so it should be good to go :)

busbey:
> Just added so it should be good to go :)

I don't see it, maybe a missed `git add`?

manolama:
Yeah, I shouldn't code on a Sunday. I had to dig up some old test cases but with the mini cluster, since HBase requires Guava 12 and AsyncHBase requires 18, I had to copy some classes that were dropped in later versions of Guava in order for the UTs to run. Is there a better way to handle those?

busbey:
hurm. sounds like a decent case for requesting the HBase project to make a shaded version of the testing module, similar to how it already makes one for hbase-client and hbase-server.

busbey:
I think we should include a note somewhere about the fact that the com.google classes are in the test sources to handle bridging the Guava 12 needs of HBaseTestingUtiltiy with the Guava 18 needs of AsyncHBase.

Maybe in the README as "a note about unit tests in this module"?

manolama:
Sounds good, note added and it'd be great to cleanup the HBase test JAR.

busbey:
Okay, I filed [HBASE-15666](https://issues.apache.org/jira/browse/HBASE-15666) to get a version of the testing utility that shades dependencies.

busbey:
the last failure was the solr binding on just one jdk. I restarted it.

PR #716, opened 2016-04-17
[solr] Update Solr binding to Solr 6.0.0

### LINKED ISSUE #715: Add Solr 6 binding
Solr 6.0.0 was released about a week ago. Solr 6 also has some fixes for JDK 9. Relates to #705

risdenk:
Fixes #715

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@ LICENSE file.
     <tarantool.version>1.6.5</tarantool.version>
     <riak.version>2.0.5</riak.version>
     <aerospike.version>3.1.2</aerospike.version>
-    <solr.version>5.4.0</solr.version>
+    <solr.version>6.0.0</solr.version>
   </properties>
 
   <modules>


risdenk:
Downside is that Solr 6.0.0 requires Java 8 as a minimum

busbey:
Can Solr 6.0.0 clients compile with jdk 1.7?

Can Solr 6.0.0 clients run against Solr 5 servers?

busbey:
based on the build output, it looks like the answer to the first question is "yes".

risdenk:
> Can Solr 6.0.0 clients compile with jdk 1.7?

Maybe but not guaranteed. Solr is pretty aggressive when it comes to Java versions and deprecating EOL versions. Java 8 is the minimum required based on the documentation.

> Can Solr 6.0.0 clients run against Solr 5 servers?

Probably since there haven't been too many major changes between last version of Solr 5 and Solr 6.0.0.  Its not guaranteed to work though. Typically should use the same version Solr client as the server.

busbey:
> > Can Solr 6.0.0 clients run against Solr 5 servers?
> Probably since there haven't been too many major changes between last version of Solr 5 and Solr 6.0.0. Its not guaranteed to work though. Typically should use the same version Solr client as the server.

Okay, then we should probably have a `solr5` and a `solr6` module. We can use jdk profiles to make it so the solr5 module isn't built if you have jdk9 and the solr6 isn't built if you have jdk7.

risdenk:
> we should probably have a solr5 and a solr6 module.

Yea thats a good point. I can put that together next week

busbey:
using the same source for both solr5 and solr6 modules will be tricky, since solr6 needs to be able to build under jdk9 when the solr5 module isn't built.

Unless the failures are just in the test code, then maybe that will work if we build solr5 under jdk9 but skip its tests.

busbey:
> using the same source for both solr5 and solr6 modules will be tricky, since solr6 needs to be able to build under jdk9 when the solr5 module isn't built.

Maybe the maven remote-resources plugin could be abused to put source files in place in the generate-sources phase?

risdenk:
haven't gotten a chance to get back to this yet. would be nice to address soon though

risdenk:
This is superseded by #849

Issue #717, opened 2016-04-17
[core] Measure JVM Behavior


manolama:
When comparing the various databases it's critical to determine how the YCSB JVM is behaving regarding thread count, memory usage, garbage collection, etc. Java provides some useful tools out of the box so I'd like to do the following:

* Add a core param like "measurejvm" that enables the JVM tracking
* Track:
  * max/min thread count in the Status thread
  * max/min memory used count in the Status thread
  * System load avg in the Status thread (from the JVM's perspective at least, not going to call any system utilities)
  * gc count in Status thread to track GCs per execution
  * final gc count for each pool (depends on the JVM)
  * Threads and mem just before each client cleanup (to see resource usage before closing down clients in case users don't have the Status thread running)
  * Hopefully track full GCs using the GC MXBeans

So far it looks pretty straightforward. We can slap a GC notification consumer on the GC beans then poll the JVM as needed for the other stats.

nitsanw:
It sounds like an investment in extra monitoring utils, all of which available externally in different forms on different platforms. Not sure it would be worth including the extra code in this project. If there's a clean external module (platform independent, maintained by other people, well established) you could use that would be nice.

manolama:
It would be really nice to get some system stats while running the benchmarks, but that would be completely platform dependent like you said. It would still be good to look at doing that, see if there's a module that would call the proper utilities based on the system it was installed on.

For the JVM metrics though, Java includes all of the calls you need for high-level details (memory, threads, GCs) and I think those are a good enough start for YCSB for now. Anything else you'd like to see measured?

busbey:
anything else going to happen here beyond #719?

nitsanw:
@manolama @busbey  looked at the merge, not clear: why is a secondary reporting thread required? why not use the StatusThread?

risdenk:
Looks like this was addressed via #719 

@nitsanw if you are interested in addressing changes due to #719 can you open a new issue?

PR #719, opened 2016-04-18
[core] Add the "measurement.trackjvm" property to measure JVM statistics

manolama:
when the status thread is enabled. This tracks the GC count, thread count,
memory used and average system load (if implemented) over time.
Related to #717

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Client.java
+++ b/core/src/main/java/com/yahoo/ycsb/Client.java
@@ -47,6 +47,12 @@ class StatusThread extends Thread
   /** Counts down each of the clients completing. */
   private final CountDownLatch _completeLatch;
 
+  /** Stores the measurements for the run. */
+  private final Measurements _measurements;
+  
+  /** Whether or not to track the JVM stats per run */
+  private final boolean _trackJVMStats;
+  
   /** The clients that are running. */
   private final List<ClientThread> _clients;
 
@@ -56,8 +62,17 @@ class StatusThread extends Thread
   /** The interval for reporting status. */
   private long _sleeptimeNs;
 
+  /** JVM max/mins */
+  private int _maxThreads;
+  private int _minThreads = Integer.MAX_VALUE;
+  private long _maxUsedMem;
+  private long _minUsedMem = Long.MAX_VALUE;
+  private double _maxLoadAvg;
+  private double _minLoadAvg = Double.MAX_VALUE;
+  private long lastGCCount = 0;
+
   /**
-   * Creates a new StatusThread.
+   * Creates a new StatusThread without JVM stat tracking.
    *
    * @param completeLatch The latch that each client thread will {@link CountDownLatch#countDown()} as they complete.
    * @param clients The clients to collect metrics from.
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -553,6 +668,20 @@ private static void exportMeasurements(Properties props, int opcount, long runti
       exporter.write("OVERALL", "RunTime(ms)", runtime);
       double throughput = 1000.0 * (opcount) / (runtime);
       exporter.write("OVERALL", "Throughput(ops/sec)", throughput);
+      
+      boolean trackJVMStats = props.getProperty(Measurements.MEASUREMENT_TRACK_JVM_PROPERTY,
```
busbey: Instead of having two places to parse this property, could we just ask the status thread if it's been collecting JVM stats?

core/src/main/java/com/yahoo/ycsb/Utils.java
```diff
@@ -174,4 +178,49 @@ public static double bytesToDouble(final byte[] bytes) {
...
+      
+      /** @return The currently used memory in bytes */
+      public static long getUsedMemoryBytes() {
+        final Runtime runtime = Runtime.getRuntime();
+        return runtime.totalMemory() - runtime.freeMemory();
+      }
+      
+      /** @return The currently used memory in megabytes. */
+      public static int getUsedMemoryMegaBytes() {
+        return (int)getUsedMemoryBytes() / 1024 / 1024;
```
busbey: Should do the division prior to casting to an int.

busbey:
Can you add a descriptive blurb into `workloads/workload_template`?

core/src/main/java/com/yahoo/ycsb/Client.java
manolama: I'd like to dump a few stats at the end (particularly around GC) even if the user doesn't run the status thread (as it's disabled by default). I'll move this into a static (ugh) variable so we just parse it once.

core/src/main/java/com/yahoo/ycsb/Utils.java
manolama: Doh, good catch.

core/src/main/java/com/yahoo/ycsb/Client.java
busbey: why not just dump it all the time then?

manolama: OK, I wasn't sure if ya'll were ok with adding stats like that overall but it would be great to have all the time.

busbey: Unless there's a downside I'm missing, sounds fine to me.

Anyone else?

manolama:
Looks like an Acumulo test failed. ``Failed tests:   workloadd(com.yahoo.ycsb.db.accumulo.AccumuloTest): Didn't get enough total cells. expected:<20000> but was:<19660>``

busbey:
Well that's not good. Maybe race condition caused by a missing flush/close? I know that client library has a bug where it still defaults to client side buffering.

Let me take a look and rerun the test.

manolama:
Had to fix a typo. This time a Cassandra failure:
``[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test (default-test) on project cassandra2-binding: Error occurred in starting fork, check output in log -> [Help 1]``

PR #720, opened 2016-04-19
[kudu] Upgrade client dependency to Kudu 0.8.0

jdcryans:


jdcryans:
I should mention that the 0.8.0 client isn't fully compatible with servers from previous versions.

busbey:
presuming folks might want to compare, could we break things into kudu07 and kudu08 modules?

jdcryans:
Until 1.0 only the latest 0.x version is really supported. We definitely don't want people to settle on a specific version. Anecdotal data suggests that folks have been good at keeping up with the pace. So I'd rather not.

busbey:
So if someone wanted to compare e.g. 0.6.0 to 0.8.0 before deciding to upgrade, what's the plan?

I suppose they can build from source with `-Dkudu.version=0.6.0` and then use those binaries to compare against the version that runs 0.8.0.

That sounds acceptable to me. Can you add a note about this in the kudu module's README?

jdcryans:
Until 1.0 I don't see any reason to do that, but it's so easy to add that I'll do it :)

busbey:
Release manager note: should mark this change as incompatible

busbey:
Thanks JD!

PR #721, opened 2016-04-20
[elasticsearch] Return not found instead of error

jasontedor:
This commit modifies the Elasticsearch client to return `Status.NOT_FOUND`
when an operation fails because the document requested does not
exist. This modifies the previous behavior which would return
`Status.ERROR` in this case.

This commit also moves the statements that would return `Status.ERROR`
into the corresponding catch blocks so that it is clear that
`Status.ERROR` is only being returned on error paths.


busbey:
Thanks!

Future release manager, I would call

jasontedor:
> Thanks!

Thanks @busbey!

> Should we call this out as incompatible since it changes results? Or not since it just moves some measurements out of the generic error case?

That's a good question. I'd lean towards calling it out as incompatible since it does change results.

PR #723, opened 2016-04-20
[riak] Added a workaround to allow strong-consistent scan transactions.

nygard89:
I think I might have find a good compromise solution for allowing strong-consistent `scan` transactions. 

So, the idea I had was the following: considering that Riak allows to utilize _secondary indexes_ only for eventually-consistent bucket-types, I've created one of these (called `fakeBucketType`) just to exploit this property. In practice, I've modified the `insert` function to make sure that, beside inserting the actual record to the proper bucket-type (which is the one called `ycsb` by default), a 1-byte record (which is the minimum allowed size) **with the same key as the latter** was stored in the `fakeBucketType` just to keep track of its secondary index. Then, for the scan transactions, I've changed the source of the secondary indexes search operation to `fakeBucketType` to retrieve the key paired with the index. This latter is then used to read the actual object searched (which, again, is stored in the `ycsb` bucket-type). And this is all... I hope I've explained myself!

Ah, just one note: I tried to keep this `fakeBucketType` as less resource-demanding as possible, by disabling all the consistency mechanisms that Riak provides (we don't need them: this "fake object" will be inserted just once, so there's no need to keep much metadata for it).

### DIFF
--- a/riak/README.md
+++ b/riak/README.md
@@ -21,7 +21,7 @@ Riak KV Client for Yahoo! Cloud System Benchmark (YCSB)
 
 The Riak KV YCSB client is designed to work with the Yahoo! Cloud System Benchmark (YCSB) project (https://github.com/brianfrankcooper/YCSB) to support performance testing for the 2.x.y line of the Riak KV database.
 
-Creating a <i>bucket type</i> to use with YCSB
+Creating a <i>bucket-type</i> to use with YCSB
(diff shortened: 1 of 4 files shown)


busbey:
this sounds fine to me. anyone else have feedback?

cmatser:
:+1:

PR #727, opened 2016-04-26
[core] Add the Utils.getGCTotalTime() method for showing how long

manolama:
the JVM spent in GC. It now prints the time and percent of time in
the final output.
Also fix cases where the garbage collector implementation returns
-1 if the count or time isn't implemented.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Utils.java
+++ b/core/src/main/java/com/yahoo/ycsb/Utils.java
@@ -219,8 +219,25 @@ public class Utils
             ManagementFactory.getGarbageCollectorMXBeans();
         long count = 0;
         for (final GarbageCollectorMXBean bean : gcBeans) {
+          if (bean.getCollectionCount() < 0) {
+            continue;
+          }
           count += bean.getCollectionCount();
         }
         return count;
+      }
+      
+      /** @return The total time, in milliseconds, spent in GC. */ 
+      public static long getGCTotalTime() {
+        final List<GarbageCollectorMXBean> gcBeans = 
+            ManagementFactory.getGarbageCollectorMXBeans();
+        long time = 0;
+        for (final GarbageCollectorMXBean bean : gcBeans) {
+          if (bean.getCollectionTime() < 0) {
+            continue;
+          }
+          time += bean.getCollectionTime();
+        }
+        return time;
       }
 }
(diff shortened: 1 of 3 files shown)


busbey:
+1

manolama:
Does the +1 mean it needs another set of eyes or is it go for merge? ;)

core/src/main/java/com/yahoo/ycsb/Utils.java
```diff
@@ -219,8 +219,25 @@ public static long getGCTotalCollectionCount() {
...
+      /** @return The total time, in milliseconds, spent in GC. */ 
+      public static long getGCTotalTime() {
+        final List<GarbageCollectorMXBean> gcBeans = 
+            ManagementFactory.getGarbageCollectorMXBeans();
+        long time = 0;
+        for (final GarbageCollectorMXBean bean : gcBeans) {
+          if (bean.getCollectionTime() < 0) {
+            continue;
+          }
+          time += bean.getCollectionTime();
```
jasontedor: I wonder if instead of summing across the collectors, each collector should be reported separately? This would enable visibility into, for example, young versus old generation collections.

manolama: I was thinking of that too and even going into the MXbeans with a watcher to get the full cause, memory used/freed etc. I could try passing in the Measurement objects and massaging the collector names into spaceless keys. Should I go that route?

busbey:
> Does the +1 mean it needs another set of eyes or is it go for merge? ;)

Generally a single +1 (from any contributor, not just those with maintainer status) is sufficient to merge a change. The only exception is if you're still getting feedback from someone else.

manolama:
Thanks @busbey. Lemme poke at @jasontedor suggestion and see how it looks.

manolama:
@jasontedor How's this?
```
[OVERALL], RunTime(ms), 25417.0
[OVERALL], Throughput(ops/sec), 3934.3746311523782
[TOTAL_GCS_PS_Scavenge], Count, 55.0
[TOTAL_GC_TIME_PS_Scavenge], Time(ms), 101.0
[TOTAL_GC_TIME_%_PS_Scavenge], Time(%), 0.39737183774639023
[TOTAL_GCS_PS_MarkSweep], Count, 0.0
[TOTAL_GC_TIME_PS_MarkSweep], Time(ms), 0.0
[TOTAL_GC_TIME_%_PS_MarkSweep], Time(%), 0.0
[TOTAL_GCs], Count, 55.0
[TOTAL_GC_TIME], Time(ms), 101.0
[TOTAL_GC_TIME_%], Time(%), 0.39737183774639023
```

jasontedor:
> @jasontedor How's this?

I pulled your patch and tried it out with a bunch of different collector configurations. It looks great! Thanks for doing this.

LGTM.

PR #735, opened 2016-05-02
Shell scripts to run YCSB.

cmatser:
I've been thinking for a while that it'd be better to use a unix shell script or windows batch script to run YCSB. It seems to me that the python script has been mostly sufficient but sometimes problematic. I've seen users run into issues with wrong python versions, not having python installed, and also not understanding the requirement to have python installed.

My hope was that a unix script or windows batch would make things easier for the end user. I had these following additional goals:
-     create a single point of update when new bindings are added
-     the linux script should also run under cygwin on windows
-     work with both release and source checkout
-     model after tomcat

Here are the files I've added:
- binding.properties - This file contains all the binding names and their associated classes. It is used by both the windows batch and unix script files. So, it is the single file to update when new bindings are added to YCSB.
- ycsb.sh - This is the Linux shell script. I've tested it on Ubuntu and on Windows 7 under Cygwin.
- ycsb.bat - This is the Windows batch script. I've tested this on Windows 7.

One caveat: The YCSB installation directory must be in a directory path that does not have spaces. (I tried very hard to make this work but couldn't. The current python script also has this issue.) I did verify that on Windows, the java executable does work in a path with spaces. This was especially important since many people have java.exe under "Program Files".

I hope this can play nicely with the Docker work that @risdenk is doing.


cmatser:
I just created this pull request, but I notice some extra commits get pulled in with it.  I'm going to close and create a new pull that is cleaner.

PR #709, opened 2016-04-13
[scripts] Fix bin/ycsb for Python 2.6 and bin/ycsb check_output

risdenk:
Fix checking of argparse module and custom check_output to not return None for err.output.

Fixes #708

PR #737, opened 2016-05-02
Shell scripts to run YCSB.

cmatser:
I've been thinking for a while that it'd be better to use a unix shell script or windows batch script to run YCSB. It seems to me that the python script has been mostly sufficient but sometimes problematic. I've seen users run into issues with wrong python versions, not having python installed, and also not understanding the requirement to have python installed.

My hope was that a unix script or windows batch would make things easier for the end user. I had these following additional goals:
-     create a single point of update when new bindings are added
-     the linux script should also run under cygwin on windows
-     work with both release and source checkout
-     model after tomcat

Here are the files I've added:
- binding.properties - This file contains all the binding names and their associated classes. It is used by both the windows batch and unix script files. So, it is the single file to update when new bindings are added to YCSB.
- ycsb.sh - This is the Linux shell script. I've tested it on Ubuntu and on Windows 7 under Cygwin.
- ycsb.bat - This is the Windows batch script. I've tested this on Windows 7.

One caveat: The YCSB installation directory must be in a directory path that does not have spaces. (I tried very hard to make this work but couldn't. The current python script also has this issue.) I did verify that on Windows, the java executable does work in a path with spaces. This was especially important since many people have java.exe under "Program Files".

I hope this can play nicely with the Docker work that @risdenk is doing.

### DIFF
--- a/bin/ycsb.sh
+++ b/bin/ycsb.sh
@@ -37,33 +37,43 @@
 #   WARNING!!! YCSB home must be located in a directory path that doesn't
 #   contain spaces.
 #
+#        www.shellcheck.net was used to validate this script
 
 # Cygwin support
 CYGWIN=false
-case "`uname`" in
+case "$(uname)" in
 CYGWIN*) CYGWIN=true;;
 esac
 
 # Get script path
-SCRIPT_DIR=`dirname "$0" 2>/dev/null`
+SCRIPT_DIR=$(dirname "$0" 2>/dev/null)
 
 # Only set YCSB_HOME if not already set
-[ -z "$YCSB_HOME" ] && YCSB_HOME=`cd "$SCRIPT_DIR/.." 2>/dev/null; pwd`
+[ -z "$YCSB_HOME" ] && YCSB_HOME=$(cd "$SCRIPT_DIR/.." || exit; pwd)
 
 # Ensure that any extra CLASSPATH variables are set via setenv.sh
 CLASSPATH=
 
 # Pull in customization options
 if [ -r "$YCSB_HOME/bin/setenv.sh" ]; then
+  # Shellcheck wants a source, but this directive only runs if available
+  #   So, tell shellcheck to ignore
+  # shellcheck source=/dev/null
   . "$YCSB_HOME/bin/setenv.sh"
 fi
 
 # Attempt to find the available JAVA, if JAVA_HOME not set
 if [ -z "$JAVA_HOME" ]; then
-  JAVA_PATH=`which java 2>/dev/null`
+  JAVA_PATH=$(which java 2>/dev/null)
   if [ "x$JAVA_PATH" != "x" ]; then
-    JAVA_HOME=`dirname $(dirname "$JAVA_PATH" 2>/dev/null)`
+    JAVA_HOME=$(dirname "$(dirname "$JAVA_PATH" 2>/dev/null)")
   fi
+fi
+
+# If JAVA_HOME still not set, error
+if [ -z "$JAVA_HOME" ]; then
+  echo "[ERROR] Java executable not found. Exiting."
+  exit 1;
 fi
 
 # Determine YCSB command argument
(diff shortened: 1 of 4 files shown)


risdenk:
Not sure I'm for or against this change. The YCSB python scripts were just updated in #709 to give better error messages. Also, the current implementation invokes Maven to build YCSB if it is not built yet which I didn't see in this PR. 

As far as the Docker testing stuff, there shouldn't be any issue to run these vs the Python implementation.

busbey:
I'd like it if we could put something at `bin/ycsb` that maintains compatibility for some OS (probably the linux based ones)

busbey:
please run the bash script through http://www.shellcheck.net/. I don't know of an equivalent for powershell, unfortunately.

bin/ycsb.sh
```diff
@@ -0,0 +1,183 @@
+#!/bin/sh
```
busbey: Are we expressly trying to keep this compatible with arbitrary shells? Personally, I think bash is prevalent enough to rely on it.

bin/ycsb.sh
```diff
@@ -0,0 +1,183 @@
...
+  for f in `ls "$YCSB_HOME"/core/target/*.jar`; do
+    CLASSPATH="$CLASSPATH:$f"
+  done
+
+  # Database conf (need to find because location is not consistent)
+  for f in `find "$YCSB_HOME"/$BINDING_DIR -name "conf"`; do
+    CLASSPATH="$CLASSPATH:$f"
+  done
+
+  # Database libraries
```
busbey: We should ask maven for these classpaths. It'll avoid errors when someone doesn't do a build before they run things.

bin/ycsb.sh
```diff
@@ -0,0 +1,183 @@
...
+if $CYGWIN; then
+  JAVA_HOME=`cygpath --absolute --windows "$JAVA_HOME"`
+  CLASSPATH=`cygpath --path --windows "$CLASSPATH"`
+fi
+
+# Get the rest of the arguments
+YCSB_ARGS=`echo $* | cut -d' ' -f3-`
+
+# Run YCSB
+echo \"$JAVA_HOME/bin/java\" $JAVA_OPTS -classpath \"$CLASSPATH\" $YCSB_CLASS $YCSB_COMMAND -db $BINDING_CLASS $YCSB_ARGS
```
busbey: before this line, we should confirm that JAVA_HOME works, and that CLASSPATH is non-empty.

bin/ycsb.sh
cmatser: I love bash, but when it comes to shell writing, I go old school.  POSIX sh seems to be the choice with most delivered software that I see.  Perhaps it's time for a change because as you said, bash is very prevalent.  I can change it.

busbey: I'd love to stick to POSIX, but I am familiar with a static analizer for bash and I don't know of one for POSIX. I expect that without one we'll drift out of compliance pretty quick.

bin/ycsb.sh
cmatser: I thought about that.  I was really trying to avoid using tools that are not part of a standard linux distro.  However if you have got the source ycsb, then you'll need to have maven to build it.  How about I put in a check for the core lib and print out a message that says something to the effect of "hey, you need to build first"?

I'm hesitant to put mvn in there because it is so much faster to just grab the jars then make a call to maven each time.  As a developer, I'll call mvn when needed.  Otherwise, I just want the script to grab the available jars and run.  That's my preference.

busbey: We get a fair number of folks new to the build process we use, so I'm hesitant to increase the tasks needed to get a successful first run.

Granted, we already state that you should build prior to running and I'd prefer folks use the binary artifacts and not build at all.

What if we check for the various jar paths we expect and if any are missing output a warning similar to what we do now and then run maven?

cmatser: I like that idea.  I can certainly check for the core and binding jars.  If not present, then invoke maven.  That resolves my issue with invoking maven on every run.

bin/ycsb.sh
cmatser: Not sure what you mean?  The shellcheck tool you refer to checks for "bashisms" if #!/bin/sh is specified.

That being said, I'm having some issue with the shellcheck tool.  A few of the recommend changes are causing the script to not find the jars I specify.

busbey: really? oh that's great. jeez I love shellcheck. Okay, POSIX it is.

Let me know if you'd like help getting the feedback to work.

cmatser:
Sorry, it took me a while to get to this as I was out of town.  Shellcheck was really useful, thanks for pointing this out to me.  I added a mvn build for source checkouts when the binding jar isn't there.

busbey:
I'd like to merge this for the upcoming 0.10.0 release. it looks essentially ready to go to me. @risdenk, @cmatser any objections?

cmatser:
Agree, let's merge.

risdenk:
looks good to me

PR #744, opened 2016-05-09
Add Java 8 Redis Cluster Client

jamespedwards42:
Still have changes to make, mainly a readme, but I had some questions before I continue.  

- Are we allowed to add clients which only support java 8+? 
- I kept the functionality identical to the original RedisClient, but is there a global objective that all DB implementations should follow?
- Is there a code formatter I can use?  I don't have any checkstyle issues with the build, but the style is a bit different.
- The original RedisClient didn't make use of the 'table' field.  For Redis Cluster it would make sense for that to be a [hash tag](http://redis.io/topics/cluster-spec#keys-hash-tags) to bind all corresponding keys and indexes to the same node.  Would it be worth adding it to this current RedisClusterClient, or maybe an additional DB client?

Here's a link to the Redis Cluster client [Jedipus](https://github.com/jamespedwards42/jedipus#jedipus-----)


pom.xml
```diff
@@ -85,7 +85,7 @@ LICENSE file.
     <mongodb.version>3.0.3</mongodb.version>
     <mongodb.async.version>2.0.1</mongodb.async.version>
     <orientdb.version>2.1.8</orientdb.version>
-    <redis.version>2.0.0</redis.version>
+    <jedis.version>2.8.1</jedis.version>
```
jamespedwards42: I changed this because it is only controlling the library version of Jedis, not Redis.  Went ahead and updated it to the latest version.

busbey:
java 8 only clients should ensure they only build when a jdk8 profile is active. besides that caveat they're fine.

redis/pom.xml
```diff
@@ -42,4 +54,18 @@ LICENSE file.
...
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.5.1</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
```
busbey: so the expectation here is that the non-cluster redis client will also not work with jdk7? can we break this into a different module so that the prior support for jdk7 remains?

pom.xml
busbey: does the jedis library version still work with the same range of redis versions? It's fine if it doesn't, so long as folks are unlikely to want to test the versions that are dropped and we add a release note.

redis/src/main/java/com/yahoo/ycsb/db/RedisClusterClient.java
```diff
@@ -0,0 +1,268 @@
+/**
+ * Copyright (c) 2012 YCSB contributors. All rights reserved.
```
busbey: year is wrong.

redis/src/main/java/com/yahoo/ycsb/db/RedisClusterClient.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    final RedisClientFactory.Builder clientFactory = RedisClientFactory.startBuilding();
+    final String password = props.getProperty(PASSWORD_PROP);
+    if (password != null) {
+      clientFactory.withAuth(password);
+    }
+
+    final Function<Node, ClientPool<RedisClient>> masterPoolFactory =
+        node -> poolBuilder.create(clientFactory.createPooled(node));
```
busbey: are we jdk8 only due to the jedis dependency or due to the use of jdk8 features here in the binding?

redis/src/main/java/com/yahoo/ycsb/db/RedisClusterClient.java
```diff
@@ -0,0 +1,268 @@
...
+
+    final boolean readOnly = true;
+    final Function<Node, ClientPool<RedisClient>> slavePoolFactory =
+        node -> poolBuilder.create(clientFactory.createPooled(node, readOnly));
+
+    rce = RedisClusterExecutor.startBuilding(Node.create(host, port)).withReadMode(ReadMode.MIXED)
```
busbey: is this operation relatively expensive? is the rce object threadsafe? I ask because currently we're making one per YCSB client thread, which in many cases can mean ~hundred of them.

jamespedwards42:
@busbey 
* I created a separate 'redis-cluster' module and left the original 'redis' module untouched other than the 'redis' to 'jedis' config naming.
* I moved the Jedis version back to its original value.  I'm not sure how their support for different Redis versions has changed over time.
*  Jedipus only has a build target for jdk8.
*  Thank you for pointing out the init() issue.  I moved that out as a static constant.  The problem now is that I'm not sure if I am accessing user defined properties correctly.  

I can run the following mvn command just fine:
`mvn -pl com.yahoo.ycsb:redis-cluster-binding -am clean package`

But `mvn clean package` on the entire project always SKIPS the redis-cluster module.  I'm not very familiar with maven and I'm struggling trying to figure out why it always skips it.

redis-cluster/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.yahoo.ycsb</groupId>
+    <artifactId>binding-parent</artifactId>
+    <version>0.10.0-SNAPSHOT</version>
```
busbey: please rebase and update to 0.11.0-SNAPSHOT

redis-cluster/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.5.1</version>
+        <configuration>
+          <source>1.8</source>
```
busbey: if this requires Java 1.8, then the module being included needs to be activated by a jdk-specific profile. our default build is Java 1.7 only.

busbey:
> But mvn clean package on the entire project always SKIPS the redis-cluster module. I'm not very familiar with maven and I'm struggling trying to figure out why it always skips it.

You need to add the redis-cluster-binding as a dependency of the distribution module. Without that, the top level build thinks it can build the distribution before it builds this binding.

PR #751, opened 2016-05-23
Fix a typo in Histogram generator

helenchw:


busbey:
Could you describe how the typo impacts users and how they should expect their results to change after this patch?

helenchw:
Currently, the `mean()` of `HistogramGenerator` is not used by other functions, so there is no impact on results. This only affects developers who use this function the as they get an incorrect value. There is no urgent need to fix the typo from the users' perspective.

busbey:
thanks for the description!

Issue #745, opened 2016-05-09
New YCSB Measurement type


tiboratAS:
In the process of running for various NoSQL databases we have added a new measurement type to the Aerospike branch of YCSB.  The measurement outputs a histogram of data periodically at a specified interval.   

The periodic histogram can be used by setting the measurement type:

	measurementtype=periodichistogram

The periodic histogram bucket size  or interval can be set by the following  setting:

	periodichistogram.bucket.interval=0.1

The number of buckets can be set with:

	periodichistogram.buckets=1000

The periodic histogram can be use with the hdrhistogram.

	measurementtype=hdrhistogram+periodichistogram

The output of the periodic measurement is in the JSON format.

JSON object format:

{ start: “ xxx.x “, end: “ yyy.y“,"opsec":zzz.zz,scale: 0.001, linear: { “0”: x, “1”:y, “5”:z, “7”:m } }
{ start: “ xxx.x “, end: “ yyy.y“,"opsec":zzz.zz,scale: 0.001, log2: { “0”: x, “1”:y, “4”:z, “8”:m } }
{ start: “ xxx.x “, end: “ yyy.y“,"opsec":zzz.zz, range: { “0.001:0.002”: x, “0.005:0.008”:y } }

Where start and end are seconds since epoch, with any granularity ( such as milliseconds ) specified as a floating point number. This allows arbitrary granularity.

Where scale is necessary for some data formats. In the linear format, it is the number of seconds in each bucket. Specified as a double ( so 1 millisecond is 0.001 ).

Where the ‘linear’ format has a tag for each bucket, and the buckets are “scale” in size, and buckets with value 0 are not represented.

Where the ‘log2’  defines the starting point, and each element in the subobject is a power of two.

Where the ‘range’ format allows arbitrary ranges to be specified. The values are seconds from the “start” identifier, so there is no “scale”.

Currently the linear format has been tested.  In the future we will be adding a log2 and range format for the histograms.

We will be open-sourcing tools for graphing the output.   

Having the data periodically output provides a powerful tool for evaluating the YCSB results.  

We would like to see the changes merged into the YCSB master.

nitsanw:
2c:
"The measurement outputs a histogram of data periodically at a specified interval." - This is already supported via the HDR histogram log file. Since the logging is loss less you can post process the data to represent precisely periods that are larger than the logging resolution. You can control the hdr logging interval with the status logging interval.

[...]

PR #753, opened 2016-05-23
Changes for New YCSB Measurement type #745

tiboratAS:
Adding the ability to create a  histogram for every measurement   interval.


busbey:
can't we already do this with the hdrhistogram measurements? @nitsanw?

nitsanw:
It's a logarithmic histogram, so:
pro: uses less memory
con: very low resolution of data

If the histogram memory use was a concern then it's a valid option, but as things stand I don't expect that to be an issue for anyone.

nitsanw:
Also, I suggest renaming this PR to something more meaningful (e.g. new logarithmic histogram measurement type).
And a further thought, if people are keen on having their own metrics, perhaps a pluggable config mechanism is in order (e.g. load the metric from a class name). This would allow extending the reporting as long as we can agree on the interface.

busbey:
this will close #745 when it lands.

busbey:
what do y'all think @tiboratAS and @BrianNichols? would having the measurements stuff setup as extensible in the first place have made getting this work off the ground easier?

core/src/main/java/com/yahoo/ycsb/measurements/Measurements.java
```diff
@@ -149,6 +159,12 @@ OneMeasurement constructOneMeasurement(String name)
       return new TwoInOneMeasurement(name,
           new OneMeasurementHdrHistogram("Hdr"+name, _props),
           new OneMeasurementHistogram("Raw"+name, _props));
+	case HDRHISTOGRAM_AND_PERIODICHISTOGRAM:
+      return new TwoInOneMeasurement(name,
+          new OneMeasurementHdrHistogram("Hdr"+name, _props),
+          new PeriodicHistogram(name, _props));
```
busbey: should this second bit be "Periodic"+name?

core/src/main/java/com/yahoo/ycsb/measurements/PeriodicHistogram.java
```diff
@@ -0,0 +1,352 @@
+/**
+ * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
```
busbey: copyright line is wrong

core/src/main/java/com/yahoo/ycsb/measurements/PeriodicHistogram.java
```diff
@@ -0,0 +1,352 @@
...
+import java.text.DecimalFormat;
+import java.util.Properties;
+
+import com.yahoo.ycsb.measurements.exporter.MeasurementsExporter;
+
+/**
+ * Take measurements and maintain a histogram of a given metric, such as READ LATENCY.
+ * Print out this histogram at periodic intervals.
+ * 
+ * @author cooperb
```
busbey: please rely on git authorship information instead of author tags. if you need to refer to multiple authors, just add them in the commit message like

```
Amending-Author: Some Name <sname@example.com>
```

core/src/main/java/com/yahoo/ycsb/measurements/PeriodicHistogram.java
```diff
@@ -0,0 +1,352 @@
...
+    prevhistogram = new int[buckets];
+    histogramoverflow = 0;
+    prevhistogramoverflow = 0;
+    operations = 0;
+    prevoperations = 0;
+    totallatency = 0;
+    totalsquaredlatency = 0;
+    min = -1;
+    max = -1;
+    windowstart = windowend = System.currentTimeMillis();
```
busbey: Sure about using System.cTM instead of System.nanoTime? the former will get moved around with the system clock

core/src/main/java/com/yahoo/ycsb/measurements/PeriodicHistogram.java
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  @Override
+  public String getSummary() 
+  {
+    writejson();    
+    return "";
+  }
+  
+  private void writejson()
```
busbey: why not rely on a json library?

busbey: in particular, the core module already has a jackson dependency

dotyjim-work:
Thanks folks for taking a look at this PR. It has languished on our side for long enough that we are suggesting that it be closed.

For folks looking for the original code referenced in this PR, please see: https://github.com/aerospike/YCSB/tree/logarithmic_histogram.

If we come back to this work, we will start with a more recent version of YCSB.

PR #755, opened 2016-05-26
Add batching in insertion in JDBC client bindings

enis:


busbey:
please prefix the commit message with the module, i.e. "[jdbc] add batching in insertion in JDBC..."

enis:
Thanks @busbey . Done.

busbey:
sorry, missed one other thing. the new parameter should be documented in jdbc/README.

enis:
@busbey no worries. Updated the README.

jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
```diff
@@ -474,7 +486,21 @@ public Status insert(String tableName, String key, HashMap<String, ByteIterator>
       for (String value: fieldInfo.getFieldValues()) {
         insertStatement.setString(index++, value);
       }
-      int result = insertStatement.executeUpdate();
+      int result;
+      if (batchSize > 0) {
+        insertStatement.addBatch();
+        if (++numRowsInBatch % batchSize == 0) {
+          int[] results = insertStatement.executeBatch();
```
busbey: we should also do this in cleanup, if there are outstanding batched inserts. otherwise we have to rely on the record count for insert being an even multiple of the batch size, I think?

kruthar:
I tested this on the edge case @busbey mentioned and it is in fact an issue. Running with `db.batchsize=18`:

```
Command line: -db com.yahoo.ycsb.db.JdbcDBClient -P workloads/workloada -P testing/mysql.properties -p recordcount=100 -load
Loading workload...
Starting test.
Adding shard node URL: jdbc:mysql://mysql.vagrant:3306/ycsb
Using shards: 1, batchSize:18, fetchSize: -1
[OVERALL], RunTime(ms), 10698.0
[OVERALL], Throughput(ops/sec), 9.347541596560104
[TOTAL_GCS_PS_Scavenge], Count, 0.0
[TOTAL_GC_TIME_PS_Scavenge], Time(ms), 0.0
[TOTAL_GC_TIME_%_PS_Scavenge], Time(%), 0.0
[TOTAL_GCS_PS_MarkSweep], Count, 0.0
[TOTAL_GC_TIME_PS_MarkSweep], Time(ms), 0.0
[TOTAL_GC_TIME_%_PS_MarkSweep], Time(%), 0.0
[TOTAL_GCs], Count, 0.0
[TOTAL_GC_TIME], Time(ms), 0.0
[TOTAL_GC_TIME_%], Time(%), 0.0
[CLEANUP], Operations, 1.0
[CLEANUP], AverageLatency(us), 1327.0
[CLEANUP], MinLatency(us), 1327.0
[CLEANUP], MaxLatency(us), 1327.0
[CLEANUP], 95thPercentileLatency(us), 1327.0
[CLEANUP], 99thPercentileLatency(us), 1327.0
[INSERT], Operations, 100.0
[INSERT], AverageLatency(us), 3246.48
[INSERT], MinLatency(us), 174.0
[INSERT], MaxLatency(us), 176767.0
[INSERT], 95thPercentileLatency(us), 10679.0
[INSERT], 99thPercentileLatency(us), 38207.0
[INSERT], Return=OK, 100
```

YCSB reports that it did 100 inserts, but when you check mysql:

```
mysql> select count(*) from usertable;
+----------+
| count(*) |
+----------+
|       90 |
+----------+
1 row in set (0.00 sec)
```

This is because some inserts are left in the buffer when you use an odd batchsize.

kruthar:
nit: ```numRowsInBatch``` is never reset to 0. You can see this with a few print statements inside of the commit block you added:

```
Commiting batch with numRows: 18
Number of rows returned: 18
Commiting batch with numRows: 36
Number of rows returned: 18
Commiting batch with numRows: 54
Number of rows returned: 18
Commiting batch with numRows: 72
Number of rows returned: 18
Commiting batch with numRows: 90
Number of rows returned: 18
```

Although this "probably" won't cause huge issues (unless doing modulo on huge numbers becomes less then negligible), it might be good to reset ```numRowsInBatch``` to 0 when you actually submit a batch.

jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
```diff
@@ -474,7 +486,21 @@ public Status insert(String tableName, String key, HashMap<String, ByteIterator>
...
+          for (int r : results) {
+            if (r != 1) {
+              return Status.ERROR;
+            }
+          }
+        }
+        return Status.OK;
```
kruthar: You might consider taking a look at the Status codes for Batch inserts in Mongo: https://github.com/brianfrankcooper/YCSB/blob/master/mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java.

Essentially, when an operation is added to the batch you return Status.BATCHED_OK, when you submit the whole batch then you return Status.OK. This way users can differentiate between the two actions when looking at the results.

busbey:
I'm planning to cut the 0.10.0 branch this evening. @enis any chance of at least the cleanup addition in time?

enis:
Yes, I'll try to do that today.

jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
enis: Thanks for testing. 
Added batch execution for partial batches in cleanup. I think in almost all of the cases, we will have num rows to be a multiple of batch size, but agreed that it is better to be safe. 
numRowsInBatch is a long. No way we are inserting 2^63 rows from a single thread. 
I've added the BATCHED_OK to be a generic Status. Seems useful across all bindings.

enis:
@busbey updated the PR.

core/src/main/java/com/yahoo/ycsb/Status.java
```diff
@@ -87,6 +87,6 @@ public boolean equals(Object obj) {
   public static final Status BAD_REQUEST = new Status("BAD_REQUEST", "The request was not valid.");
   public static final Status FORBIDDEN = new Status("FORBIDDEN", "The operation is forbidden.");
   public static final Status SERVICE_UNAVAILABLE = new Status("SERVICE_UNAVAILABLE", "Dependant service for the current binding is not available.");
-  
+  public static final Status BATCHED_OK = new Status("BATCHED_OK", "The operation has been batched by the binding to be executed later.");
```
busbey: if you're going to expose this here, please keep it in an isolated commit that also moves the mongo uses of BATCHED_OK to use the newly generalized status result.

busbey: and that commit should note that it impacts "[core,mongodb]" in its commit message.

enis: done.

busbey:
thanks @enis!

PR #756, opened 2016-05-28
Added Rest Client Module to benchmark Restful Web Services

shivam-maharshi:
Please review. Touched only the absolutely necessary files required. CheckStyle formatting is followed, unit tests created, build passes.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -81,6 +81,7 @@ DATABASES = {
     "orientdb"     : "com.yahoo.ycsb.db.OrientDBClient",
     "rados"        : "com.yahoo.ycsb.db.RadosClient",
     "redis"        : "com.yahoo.ycsb.db.RedisClient",
+    "rest"         : "com.yahoo.ycsb.webservice.rest.RestClient",
     "riak"         : "com.yahoo.ycsb.db.riak.RiakKVClient",
     "s3"           : "com.yahoo.ycsb.db.S3Client",
     "solr"         : "com.yahoo.ycsb.db.solr.SolrClient",
--- a/core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
+++ b/core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
@@ -1,0 +1,304 @@
+/**
+ * Copyright (c) 2016 YCSB contributors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+package com.yahoo.ycsb.workloads;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+import com.yahoo.ycsb.ByteIterator;
+import com.yahoo.ycsb.DB;
+import com.yahoo.ycsb.RandomByteIterator;
+import com.yahoo.ycsb.WorkloadException;
+import com.yahoo.ycsb.generator.DiscreteGenerator;
+import com.yahoo.ycsb.generator.ExponentialGenerator;
+import com.yahoo.ycsb.generator.HotspotIntegerGenerator;
+import com.yahoo.ycsb.generator.NumberGenerator;
+import com.yahoo.ycsb.generator.UniformIntegerGenerator;
+import com.yahoo.ycsb.generator.ZipfianGenerator;
+
+/**
+ * Typical RESTFul services benchmarking scenario. Represents a set of client
+ * calling REST operations like HTTP DELETE, GET, POST, PUT on a web service.
+ * This scenario is completely different from CoreWorkload which is mainly
+ * designed for databases benchmarking. However due to some reusable
+ * functionality this class extends {@link CoreWorkload} and overrides necessary
+ * methods like init, doTransaction etc.
+ */
+public class RestWorkload extends CoreWorkload {
+
+  /**
+   * The name of the property for the proportion of transactions that are
+   * delete.
+   */
+  public static final String DELETE_PROPORTION_PROPERTY = "deleteproportion";
+
+  /**
+   * The default proportion of transactions that are delete.
+   */
+  public static final String DELETE_PROPORTION_PROPERTY_DEFAULT = "0.00";
+
+  /**
+   * The name of the property for the file that holds the field length size for insert operations.
+   */
+  public static final String FIELD_LENGTH_DISTRIBUTION_FILE_PROPERTY = "fieldlengthdistfile";
+  
+  /**
+   * The default file name that holds the field length size for insert operations.
+   */
+  public static final String FIELD_LENGTH_DISTRIBUTION_FILE_PROPERTY_DEFAULT = "fieldLengthDistFile.txt";
+
+  /**
+   * In web services even though the CRUD operations follow the same request
+   * distribution, they have different traces and distribution parameter
+   * values. Hence configuring the parameters of these operations separately
+   * makes the benchmark more flexible and capable of generating better
+   * realistic workloads.
+   */
+  // Read related properties.
+  private static final String READ_TRACE_FILE = "url.trace.read";
+  private static final String READ_TRACE_FILE_DEFAULT = "readtrace.txt";
+  private static final String READ_ZIPFIAN_CONSTANT = "readzipfconstant";
+  private static final String READ_ZIPFIAN_CONSTANT_DEAFULT = "0.99";
+  private static final String READ_RECORD_COUNT_PROPERTY = "readrecordcount";
+  // Insert related properties.
+  private static final String INSERT_TRACE_FILE = "url.trace.insert";
+  private static final String INSERT_TRACE_FILE_DEFAULT = "inserttrace.txt";
+  private static final String INSERT_ZIPFIAN_CONSTANT = "insertzipfconstant";
+  private static final String INSERT_ZIPFIAN_CONSTANT_DEAFULT = "0.99";
+  private static final String INSERT_SIZE_ZIPFIAN_CONSTANT = "insertsizezipfconstant";
+  private static final String INSERT_SIZE_ZIPFIAN_CONSTANT_DEAFULT = "0.99";
+  private static final String INSERT_RECORD_COUNT_PROPERTY = "insertrecordcount";
+  // Delete related properties.
+  private static final String DELETE_TRACE_FILE = "url.trace.delete";
+  private static final String DELETE_TRACE_FILE_DEFAULT = "deletetrace.txt";
+  private static final String DELETE_ZIPFIAN_CONSTANT = "deletezipfconstant";
+  private static final String DELETE_ZIPFIAN_CONSTANT_DEAFULT = "0.99";
+  private static final String DELETE_RECORD_COUNT_PROPERTY = "deleterecordcount";
+  // Delete related properties.
+  private static final String UPDATE_TRACE_FILE = "url.trace.update";
+  private static final String UPDATE_TRACE_FILE_DEFAULT = "updatetrace.txt";
+  private static final String UPDATE_ZIPFIAN_CONSTANT = "updatezipfconstant";
+  private static final String UPDATE_ZIPFIAN_CONSTANT_DEAFULT = "0.99";
+  private static final String UPDATE_RECORD_COUNT_PROPERTY = "updaterecordcount";
+
+  private Map<Integer, String> readUrlMap;
+  private Map<Integer, String> insertUrlMap;
+  private Map<Integer, String> deleteUrlMap;
+  private Map<Integer, String> updateUrlMap;
+  private int readRecordCount;
+  private int insertRecordCount;
+  private int deleteRecordCount;
+  private int updateRecordCount;
+  private NumberGenerator readKeyChooser;
+  private NumberGenerator insertKeyChooser;
+  private NumberGenerator deleteKeyChooser;
+  private NumberGenerator updateKeyChooser;
+
+  @Override
+  public void init(Properties p) throws WorkloadException {
+
+    readRecordCount = Integer.parseInt(p.getProperty(READ_RECORD_COUNT_PROPERTY, String.valueOf(Integer.MAX_VALUE)));
+    insertRecordCount = Integer
+        .parseInt(p.getProperty(INSERT_RECORD_COUNT_PROPERTY, String.valueOf(Integer.MAX_VALUE)));
+    deleteRecordCount = Integer
+        .parseInt(p.getProperty(DELETE_RECORD_COUNT_PROPERTY, String.valueOf(Integer.MAX_VALUE)));
+    updateRecordCount = Integer
+        .parseInt(p.getProperty(UPDATE_RECORD_COUNT_PROPERTY, String.valueOf(Integer.MAX_VALUE)));
+      
+      readUrlMap = getTrace(p.getProperty(READ_TRACE_FILE, READ_TRACE_FILE_DEFAULT), readRecordCount);
+      insertUrlMap = getTrace(p.getProperty(INSERT_TRACE_FILE, INSERT_TRACE_FILE_DEFAULT), insertRecordCount);
+      deleteUrlMap = getTrace(p.getProperty(DELETE_TRACE_FILE, DELETE_TRACE_FILE_DEFAULT), deleteRecordCount);
+      updateUrlMap = getTrace(p.getProperty(UPDATE_TRACE_FILE, UPDATE_TRACE_FILE_DEFAULT), updateRecordCount);
+    
+    operationchooser = createOperationGenerator(p);
+
+    // Common distribution for all operations.
+    String requestDistrib = p.getProperty(REQUEST_DISTRIBUTION_PROPERTY, REQUEST_DISTRIBUTION_PROPERTY_DEFAULT);
+
+    double readZipfconstant = Double
+        .parseDouble(p.getProperty(READ_ZIPFIAN_CONSTANT, READ_ZIPFIAN_CONSTANT_DEAFULT));
+    readKeyChooser = getKeyChooser(requestDistrib, readUrlMap.size(), readZipfconstant, p);
+    double updateZipfconstant = Double
+        .parseDouble(p.getProperty(UPDATE_ZIPFIAN_CONSTANT, UPDATE_ZIPFIAN_CONSTANT_DEAFULT));
+    updateKeyChooser = getKeyChooser(requestDistrib, updateUrlMap.size(), updateZipfconstant, p);
+    double insertZipfconstant = Double
+        .parseDouble(p.getProperty(INSERT_ZIPFIAN_CONSTANT, INSERT_ZIPFIAN_CONSTANT_DEAFULT));
+    insertKeyChooser = getKeyChooser(requestDistrib, insertUrlMap.size(), insertZipfconstant, p);
+    double deleteZipfconstant = Double
+        .parseDouble(p.getProperty(DELETE_ZIPFIAN_CONSTANT, DELETE_ZIPFIAN_CONSTANT_DEAFULT));
+    deleteKeyChooser = getKeyChooser(requestDistrib, deleteUrlMap.size(), deleteZipfconstant, p);
+
+    fieldlengthgenerator = getFieldLengthGenerator(p);
+  }
+  
+  public static DiscreteGenerator createOperationGenerator(final Properties p) {
+     // Re-using CoreWorkload method. 
+    final DiscreteGenerator operationChooser = CoreWorkload.createOperationGenerator(p);
+    // Needs special handling for delete operations not supported in CoreWorkload.
+    double deleteproportion = Double
+		        .parseDouble(p.getProperty(DELETE_PROPORTION_PROPERTY, DELETE_PROPORTION_PROPERTY_DEFAULT));
+    if (deleteproportion > 0)
+      operationChooser.addValue(deleteproportion, "DELETE");
+    return operationChooser;
+  }
+
+  private static NumberGenerator getKeyChooser(String requestDistrib, int recordCount, double zipfContant,
+      Properties p) throws WorkloadException {
+    NumberGenerator keychooser = null;
+    
+    switch(requestDistrib) {
+      case "exponential":
+        double percentile = Double.parseDouble(p.getProperty(ExponentialGenerator.EXPONENTIAL_PERCENTILE_PROPERTY,
+            ExponentialGenerator.EXPONENTIAL_PERCENTILE_DEFAULT));
+        double frac = Double.parseDouble(p.getProperty(ExponentialGenerator.EXPONENTIAL_FRAC_PROPERTY,
+            ExponentialGenerator.EXPONENTIAL_FRAC_DEFAULT));
+        keychooser = new ExponentialGenerator(percentile, recordCount * frac);
+        break;
+      case "uniform":
+        keychooser = new UniformIntegerGenerator(0, recordCount - 1);
+        break;
+      case "zipfian":
+        keychooser = new ZipfianGenerator(recordCount, zipfContant);
+        break;
+      case "latest":
+        throw new WorkloadException("Latest request distribution is not supported for RestWorkload.");
+      case "hotspot":
+        double hotsetfraction = Double
+          .parseDouble(p.getProperty(HOTSPOT_DATA_FRACTION, HOTSPOT_DATA_FRACTION_DEFAULT));
+        double hotopnfraction = Double
+          .parseDouble(p.getProperty(HOTSPOT_OPN_FRACTION, HOTSPOT_OPN_FRACTION_DEFAULT));
+        keychooser = new HotspotIntegerGenerator(0, recordCount - 1, hotsetfraction, hotopnfraction);
+        break;
+      default:
+        throw new WorkloadException("Unknown request distribution \"" + requestDistrib + "\"");
+    }
+    return keychooser;
+  }
+
+  protected static NumberGenerator getFieldLengthGenerator(Properties p) throws WorkloadException {
+    // Re-using CoreWorkload method. 
+    NumberGenerator fieldLengthGenerator = CoreWorkload.getFieldLengthGenerator(p);
+    String fieldlengthdistribution = p.getProperty(
+            FIELD_LENGTH_DISTRIBUTION_PROPERTY, FIELD_LENGTH_DISTRIBUTION_PROPERTY_DEFAULT);
+    // Needs special handling for Zipfian distribution for variable Zipf Constant.
+    if (fieldlengthdistribution.compareTo("zipfian") == 0) {
+    	  int fieldlength =
+    		        Integer.parseInt(p.getProperty(FIELD_LENGTH_PROPERTY, FIELD_LENGTH_PROPERTY_DEFAULT));
+        double insertsizezipfconstant = Double
+            .parseDouble(p.getProperty(INSERT_SIZE_ZIPFIAN_CONSTANT, INSERT_SIZE_ZIPFIAN_CONSTANT_DEAFULT));
+        fieldLengthGenerator = new ZipfianGenerator(1, fieldlength, insertsizezipfconstant);
+      }
+    return fieldLengthGenerator;
+  }
+
+  /**
+   *  Reads the trace file and returns a URL map.
+   */
+  private static Map<Integer, String> getTrace(String filePath, int recordCount)
+      throws WorkloadException {
+    Map<Integer, String> urlMap = new HashMap<Integer, String>();
+    int count = 0;
+    String line;
+    try {
+      FileReader inputFile = new FileReader(filePath);
+      BufferedReader bufferReader = new BufferedReader(inputFile);
+      while ((line = bufferReader.readLine()) != null) {
+        urlMap.put(count++, line.trim());
+        if (count >= recordCount)
+          break;
+      }
+      bufferReader.close();
+    } catch (IOException e) {
+      throw new WorkloadException(
+          "Error while reading the trace. Please make sure the trace file path is correct. "
+              + e.getLocalizedMessage());
+    }
+    return urlMap;
+  }
+
+  /**
+   * Not required for Rest Clients as data population is service specific.
+   */
+  @Override
+  public boolean doInsert(DB db, Object threadstate) {
+    return false;
+  }
+
+  @Override
+  public boolean doTransaction(DB db, Object threadstate) {
+    switch (operationchooser.nextString()) {
+    case "UPDATE":
+      doTransactionUpdate(db);
+      break;
+    case "INSERT":
+      doTransactionInsert(db);
+      break;
+    case "DELETE":
+      doTransactionDelete(db);
+      break;
+    default:
+      doTransactionRead(db);
+    }
+    return true;
+  }
+  
+  /**
+   *  Returns next URL to be called.
+   */
+  private String getNextURL(int opType) {
+    if (opType == 1)
+      return readUrlMap.get(readKeyChooser.nextValue().intValue());
+    else if (opType == 2)
+      return insertUrlMap.get(insertKeyChooser.nextValue().intValue());
+    else if (opType == 3)
+      return deleteUrlMap.get(deleteKeyChooser.nextValue().intValue());
+    else
+      return updateUrlMap.get(updateKeyChooser.nextValue().intValue());
+  }
+
+  @Override
+  public void doTransactionRead(DB db) {
+    HashMap<String, ByteIterator> result = new HashMap<String, ByteIterator>();
+    db.read(null, getNextURL(1), null, result);
+  } 
+  
+  @Override
+  public void doTransactionInsert(DB db) {
+    HashMap<String, ByteIterator> value = new HashMap<String, ByteIterator>();
+    // Create random bytes of insert data with a specific size.
+    value.put("data", new RandomByteIterator(fieldlengthgenerator.nextValue().longValue()));
+    db.insert(null, getNextURL(2), value);
+  }
+
+  public void doTransactionDelete(DB db) {
+    db.delete(null, getNextURL(3));
+  }
+
+  @Override
+  public void doTransactionUpdate(DB db) {
+    HashMap<String, ByteIterator> value = new HashMap<String, ByteIterator>();
+    // Create random bytes of update data with a specific size.
+    value.put("data", new RandomByteIterator(fieldlengthgenerator.nextValue().longValue()));
+    db.update(null, getNextURL(4), value);
+  }
+  
+}
--- a/rest/README.md
+++ b/rest/README.md
@@ -1,0 +1,181 @@
+<!--
+Copyright (c) 2016 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+## Quick Start
+
+This section describes how to run YCSB to benchmark HTTP RESTful
+webservices. The aim of the rest binding is to benchmark the 
+performance of any sepecific HTTP RESTful webservices with real
+life (production) dataset. This must not be confused with benchmarking
+various webservers (like Apache Tomcat, Nginx, Jetty) using a dummy 
+dataset.
+
+### 1. Set Up YCSB
+
+Clone the YCSB git repository and compile:
+
+    git clone git://github.com/brianfrankcooper/YCSB.git
+    cd YCSB
+    mvn -pl com.yahoo.ycsb:rest-binding -am clean package
+
+### 2. Set Up an HTTP Web Service
+
+There must be a running HTTP RESTful webservice accesible from 
+the instance on which YCSB is running. If the webservice is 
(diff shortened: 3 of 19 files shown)


shivam-maharshi:
@busbey Please give feedback when you get some time. I have incorporated all your old comments.

shivam-maharshi:
@risdenk @kruthar In case you get sometime to take a look at this. Please do and provide feedback.

kruthar:
Hi @shivam-maharshi - I have been looking at this PR on and off. For me it is hard to review for a couple reasons:

1. The PR description doesn't give any detail into what this PR is trying to accomplish, I'm hesitant to approve this because I don't know really what it is for or what exactly it does, or any of the motivation behind it, why it was necessary to create a new Workload
2. Following the documentation theme there isn't a README.md which would tell me a little more about the Rest driver, what it is for, how to use it, the properties, etc.

I started looking into the RestWorkload.java, it is very different from CoreWorkload.java and so is not easy to view differences between the two with a diff. This reinforces the need for more documentation around what the motivations for this PR are and the benefits it brings.

Some specific things that are confusing to me:

1. I am not exactly sure what this trace stuff is for. You have separated out trace properties for each operation, but then in the workload file examples you have just given the same file name for each operation. what is the point there?
2. It seems strange to me that at the RestWorkload.java level you have removed the scan operation.
3. It looks like the key and value generating paradigms have changed, especially since buildDeterministicValue doesn't exist in RestWorkload.java

Sorry for the negative feedback but some things I would really like to see are more documentation and conversation about the work you've done, at this point there is too much change to really tell everything that is going on. Along with that it would be great to discuss if the changes you have made in RestWorkload.java could actually be integrated through additions to CoreWorkload.java and the bulk of the changes in the `rest` binding you created.

shivam-maharshi:
Hi @kruthar, thanks a lot for providing a detailed explanation. My aim is to eventually get this PR accepted. I am open to making any changes that are required to make it possible. Any feedbacks that will help me move in this direction are more than welcome! Here are the answers to some of your concerns.

1. I have added a detailed README.md for this driver. It includes description about what this driver is, what is the use case, how to use it and the description of the properties to be used with examples. This was an important point, thanks for pointing it out.

2. This main aim of this PR is to add the support of benchmarking RESTFul webservices using YCSB. Right now YCSB can only benchmark databases, however it has the potential to be an excellent tool for benchmarking webservices mainly because of three reasons. Firstly, it is highly configurable and customizable. Secondly, it is capable of producing huge amount of workload without boging the client machine down, something needed for cloud services. Thirdly, since it is written on java, it is platform independent. I have performed a detailed study into the available free opensource tools most suited for benchmarking cloud webservices, and found that there is a lack of such a tool (My MS thesis is on a related topic). Hence I finally decided to add this module to YCSB since it catered to my need and I feel that many developers would be thrilled to use YCSB it. They can also easily tweak it to fit thier own usecase.

3. Regarding RestWorkload, below are the explanations:
a. Trace is a file that holds the path of URL resources that are to be used for making HTTP requests to benchmark your web service. For example, if your webservice exists at http://{host}:{port}/{endpoint}, and you want to benchmark the performance of READS on this webservice then the url.trace.read file might look like this:

http://{host}:{port}/{endpoint}/resource_1
http://{host}:{port}/{endpoint}/resource_2
http://{host}:{port}/{endpoint}/resource_3

The RestWorkload will then pickup these URL resources and put them in a map. According to the requestdistribution type, the URLs will be picked from this map and passed to RestClient binding to make HTTP GET requests on these URLs. Similar logic will also apply to INSERT, DELETE and UPDATE methods.

b. In real life the URL access pattern is different for different operations. For example, HTTP GET will have a different URL access pattern than HTTP POST for a webservice. Hence there must be different trace files for each such operation. However for simplicity, we might want to use similar trace file. This is the case here. 

c. The support for SCAN operations are important to benchmark from a database perspective where sometimes a collection of entries are fetched or updated for some application layer logic. This is a real life use case for databases. However for webservices this is highly unlikely use case. Hence the support for this operation is not needed. However it can always be added in the future if deemed worth the time and efforts.

d. In RestWorkload, the keys generated are simply used to pick a URL from the trace maps, loaded at initialization of the class. Using this technique we are able to decide which URL should the next request be made so as to produce a URL access pattern similar to real life. The buildDeterministicValue is hence not required in this use case.

I totally understand your concerns and I am more than happy to discuss with you the reasoning behind this binding and the design opted. We can have a conversation on phone/email/skype or any other convenient option as you prefer. If you or anyone else have better suggestions regarding design and we would need to restructure the code, I would be more than happy to do that myself as well. Thanks again for your response. Hoping to hear back soon again.

kruthar:
@shivam-maharshi - thanks for the background info and context, it makes a lot more sense now why this type of workload needs to be so different from the traditional CoreWorkload. I think you have also worked in enough flexibility to work with many potential REST layouts by using the trace files, a little more work for the user up front, but it means they can be sure it will hit all the endpoints they want it to and they dont seem to have to heterogenous.

@busbey - there are several things that need to be updated in this PR before being merged in, but do you have any problems with trying to get this one in for 0.10.0 as an Untested binding? I would put it in as untested to start with because of the fact that it introduces a new Workload type, and I don't have a REST services on hand to test this thing.

rest/README.md
```diff
@@ -0,0 +1,174 @@
...
+http://127.0.0.1:8080/my_service/resource_1
+http://127.0.0.1:8080/my_service/resource_2
+http://127.0.0.1:8080/my_service/resource_3
+http://127.0.0.1:8080/my_service/resource_4
+http://127.0.0.1:8080/my_service/resource_5
+
+The rest module will pick up URLs from the above file according to
+the `requestdistribution` property (default is zipfian) mentioned in
+the rest_workload. In the example above we assume that the property `url.prefix` 
+(see below for property description) is set to ``. If url.prefix
```
kruthar: nit - there are two backticks here that are throwing off formatting.

rest/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+	OR CONDITIONS OF ANY KIND, either express or implied. See the License for 
+	the specific language governing permissions and limitations under the License. 
+	See accompanying LICENSE file. -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.yahoo.ycsb</groupId>
+    <artifactId>binding-parent</artifactId>
+    <version>0.10.0-SNAPSHOT</version>
```
kruthar: A release has just been started, update the snapshot version to 0.11.0-SNAPSHOT

kruthar:
Can you rebase this PR to pick up the new ycsb shell scripts, then add a line for this rest binding in bin/bindings.properties. The python tool at bin/ycsb will be deprecated out at some point, but you might as well update that also with a line for your binding.

kruthar:
Would also be greatly appreciated if you could rebase your PR done to just a handful of meaningful commits. Please prefix each commit with `[rest]` as that is the subject matter.

rest/README.md
shivam-maharshi: Fixed it by using empty keyword instead of ``.

rest/pom.xml
shivam-maharshi: Updated snapshot to 0.11.0-SNAPSHOT

shivam-maharshi:
@kruthar Rebased the PR and updated bin/binding.properties and python bin/ycsb to include rest module. Also squashed all the different commits into a single commit and used [rest] in the message log. However I see all the commits by other developers (made in the last month) into this PR as well. I think I made a mistake somewhere rebasing my branch. Not sure how to fix this. Would this PR work or should I close this and create a new PR? Please let me know and I'll do the same.

kruthar:
@busbey, @risdenk - looks like this is failing with the jdk8 test due to the sigar issue addressed in PR #776. I don't see any other recent PR's failing so thinking this is kinda weird. Thoughts? Is #776 required to do jdk8 builds now?

risdenk:
@kruthar The sigar thing is just a warning. It doesn't actually fail anything. The real issue in the test is `Killed` at the end after

```
Connected to cluster: Test Cluster
Datacenter: datacenter1; Host: localhost/127.0.0.1; Rack: rack1
Connected to cluster: Test Cluster
Datacenter: datacenter1; Host: localhost/127.0.0.1; Rack: rack1
Connected to cluster: Test Cluster
Datacenter: datacenter1; Host: localhost/127.0.0.1; Rack: rack1
```

The real issue is the JVM crashed or hung or something. Maybe out of memory or something from Travis?
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test (default-test) on project cassandra2-binding: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test failed: The forked VM terminated without saying properly goodbye. VM crash or System.exit called ? -> [Help 1]
```

busbey:
> @busbey - there are several things that need to be updated in this PR before being merged in, but do you have any problems with trying to get this one in for 0.10.0 as an Untested binding? I would put it in as untested to start with because of the fact that it introduces a new Workload type, and I don't have a REST services on hand to test this thing.

I'd rather take the time to get it right before merging. 0.11.0 will come out in short order after 0.10.0.

jixuan1989:
Sorry, I am not sure that sigar is only suitable for jdk8. And if we do not have sigar files, there will be exception information in the console: no libsigar-universal64-macosx.dylib in java.library.path. 

But as the above said, maybe Cassandra can still startup (I am not sure).

shivam-maharshi:
@kruthar @busbey Please share your feedbacks about what all needs to be updated in the PR. I will be more than happy to work with you guys on this.

bin/ycsb
```diff
@@ -83,6 +83,7 @@ DATABASES = {
     "orientdb"     : "com.yahoo.ycsb.db.OrientDBClient",
     "rados"        : "com.yahoo.ycsb.db.RadosClient",
     "redis"        : "com.yahoo.ycsb.db.RedisClient",
+	"rest"         : "com.yahoo.ycsb.webservice.rest.RestClient",
```
kruthar: nit - line this up

rest/pom.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2011 YCSB project, 2014 - 2016 YCSB contributors. All
```
kruthar: This is a new file it can just have 2016 YCSB Contributors

rest/src/test/resources/workload_rest
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2016 Yahoo! Inc. All rights reserved.
```
kruthar: Is it necessary to have this copy of the file? The copy in workloads/ will be distributed.

kruthar:
license edit in 
- bin/ycsb
- distribution/pom.xml

Does anyone else have a REST service just laying around to test this with?

rest/src/test/resources/workload_rest
shivam-maharshi: Its not necessary but good to have. Since we've already mentioned a detailed README.md, this can be removed without implications.

shivam-maharshi:
@kruthar Incorporated feedbacks. For starters you can just test read operations using 5 Google URLs in the trace file like this:

url.prefix=https://www.google.com/#q=

**TraceFile:**
study
education
aws
cloud+computing
climate+change

shivam-maharshi:
@kruthar Please let me know if you need some help setting a dummy Rest service. We can work together on verifying this anytime you are free.

busbey:
as an alternative for unblocking testing, what if we add an integration test that stands up an embedded REST service and then runs against it?

shivam-maharshi:
@busbey We are using Wiremock for imitating Rest services (here: https://github.com/brianfrankcooper/YCSB/pull/756/files#diff-237d7ac48706b45c30b7c2ce9cec005eR97) in the RestClientTests, which take care of the HTTP handling part. It can also be used for integration test cases. On a side note I can add unit tests for RestWorkload to increase test coverage, thereby making sure that this module works as integration test cases might take sometime.

busbey:
I saw the use of Wiremock for the unit tests, I was hoping for something more thorough for an integration test. Mocks are great for verifying well-encapsulated assumptions about the client, but tend to miss out on the sorts of edge cases that come from talking to an actual system. I was hoping for something like an embedded tomcat demo or the like.

NBD if there isn't something on hand. We'll just need to wait for somebody to have time to test things out. (that needn't be @kruthar btw, or even any of the existing project moderators. we accept reviews from anybody.)

shivam-maharshi:
@busbey @kruthar I've added integration test cases using embedded Tomcat 8. These tests start the embedded server (once at class loading) which exposes a mock RESTFul service used for benchmarking. Benchmarking is invoked using Client class from the core module to generate load using the RestWorkload. Everything works as expected. Please review the changes and let me know.

shivam-maharshi:
@channel I've confirmed the working of the rest-binding both manually as well as using integration test cases. Please let me know if it needs more changes to get accepted.

shivam-maharshi:
@risdenk @jasontedor Can you please take a look at this pull request as the other moderators are busy.

risdenk:
@shivam-maharshi I'll try to look on Friday.

shivam-maharshi:
@risdenk thanks. Even though integration tests runs fine, please let me know if you need some help testing this on a real RESTFul service.

rest/README.md
```diff
@@ -0,0 +1,181 @@
...
+various webservers (like Apache Tomcat, Nginx, Jetty) using a dummy 
+dataset.
+
+### 1. Set Up YCSB
+
+Clone the YCSB git repository and compile:
+
+    git clone git://github.com/brianfrankcooper/YCSB.git
+    cd YCSB
+    mvn -pl com.yahoo.ycsb:solr-binding -am clean package
```
risdenk: This refers to solr-binding instead of rest?

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
```diff
@@ -0,0 +1,245 @@
...
+ * been included in success and failure test cases.
+ */
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class IntegrationTest {
+
+  @Rule
+  public final ExpectedSystemExit exit = ExpectedSystemExit.none();
+
+  private static int port = 8080;
+  private static Tomcat tomcat = new Tomcat();
```
risdenk: Should this be in the init()?

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
```diff
@@ -0,0 +1,245 @@
...
+  private static final String BASE_DIR = System.getProperty("user.dir");
```
risdenk: You should be able to get the paths to these files with Java resources. Some psuedo code:

```
this.class.getClassLoader().getResource()
```

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
```diff
@@ -0,0 +1,245 @@
...
+  private static final String WORKLOAD_FILEPATH = new File(BASE_DIR + "/src/test/resources/workload_rest").getAbsolutePath();
+  private static final String TRACE_FILEPATH = new File(BASE_DIR + "/src/test/resources/trace.txt").getAbsolutePath();
+  private static final String ERROR_TRACE_FILEPATH = new File(BASE_DIR + "/src/test/resources/error_trace.txt").getAbsolutePath();
+  private static final String RESULTS_FILEPATH = new File(BASE_DIR + "/src/test/resources/results.txt").getAbsolutePath();
+
+  @BeforeClass
+  public static void init() throws ServletException, LifecycleException, FileNotFoundException, IOException,
+      DBException, InterruptedException {
+    String webappDirLocation = "WebContent/";
```
risdenk: Can WebContent be moved to test/resources?

rest/src/test/java/com/yahoo/ycsb/webservice/rest/RestClientTest.java
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Mocking REST services with the easiest and the most straight forward usage
+ * using stubbing. Performs these steps in order. 1. Mocks the web server part
+ * of the rest services. 2. Invokes the rest client to perform HTTP operations
+ * on the server. 3. Compares the response received with the expected response.
+ * Various cases like - success, failure, timeout, non-implemented etc have been
+ * demonstrated.
+ */
+ // Test cases pass for JDK8 but fail for JDK7 due to WireMock.
+ @Ignore
```
risdenk: This whole test class is ignored?

rest/src/test/java/com/yahoo/ycsb/webservice/rest/RestClientTest.java
```diff
@@ -0,0 +1,215 @@
...
+  private static byte[] GZIP_SUCCESS_REPONSE = null;
+  private static final String NOT_FOUND_RESPONSE = "<response>not found</response>";
+  private static final String INPUT_DATA = "<field1>one</field1><field2>two</field2>";
+  private static final int RESPONSE_DELAY = 2000; // 2 seconds.
+  private static final String resource = "1";
+
+  @BeforeClass
+  public static void init() throws FileNotFoundException, IOException, DBException {
+    Properties props = new Properties();
+    props.load(new FileReader("src/test/resources/workload_rest"));
```
risdenk: Again should be able to use resources loading instead of a full path?

rest/src/test/resources/workload_rest
risdenk: Looks like you said this would be removed and it hasn't been?

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
```diff
@@ -0,0 +1,300 @@
+/**
```
risdenk: I think more time needs to be spent removing this RestWorkload file. I think it overlaps with CoreWorkload.

The format of CoreWorkload and RestWorkload of how it documents the properties and defines variables is different I think. Maybe try to make them more similar?

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
```diff
@@ -0,0 +1,300 @@
...
+  private static final String DELETE_ZIPFIAN_CONSTANT_DEAFULT = "0.99";
+  private static final String DELETE_RECORD_COUNT_PROPERTY = "deleterecordcount";
+  // Delete related properties.
+  private static final String UPDATE_TRACE_FILE = "url.trace.update";
+  private static final String UPDATE_TRACE_FILE_DEFAULT = "updatetrace.txt";
+  private static final String UPDATE_ZIPFIAN_CONSTANT = "updatezipfconstant";
+  private static final String UPDATE_ZIPFIAN_CONSTANT_DEAFULT = "0.99";
+  private static final String UPDATE_RECORD_COUNT_PROPERTY = "updaterecordcount";
+
+  private static Map<Integer, String> readUrlMap;
```
risdenk: Should these all be static? Can these be accessed by different threads?

rest/pom.xml
```diff
@@ -0,0 +1,144 @@
...
+    <dependency>
+      <groupId>com.yahoo.ycsb</groupId>
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.1</version>
```
risdenk: For consistency, can you pull all the versions into the properties above like you did for tomcat and jersey?

rest/README.md
shivam-maharshi: Corrected

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
shivam-maharshi: Moved.

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
shivam-maharshi: Thanks! Code Snippet - `IntegrationTest.class.getClassLoader().getResource("error_trace.txt").getPath();`

rest/src/test/java/com/yahoo/ycsb/webservice/rest/RestClientTest.java
shivam-maharshi: Wiremock (The framework used to mock rest services for RestClientTest) requires JDK8 support. Hence this class is ignored as the tests fail on JDK7 but pass on JDK8.

rest/src/test/java/com/yahoo/ycsb/webservice/rest/RestClientTest.java
shivam-maharshi: Done. Code Snippet - 
`props.load(new FileReader(RestClientTest.class.getClassLoader().getResource("workload_rest").getPath()));`

rest/src/test/java/com/yahoo/ycsb/webservice/rest/RestClientTest.java
risdenk: Should be able to ignore tests for a specific JDK. See https://github.com/brianfrankcooper/YCSB/blob/master/cassandra/pom.xml#L67

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
shivam-maharshi: Good point! They will be accessed by different threads but only for reading. But yes it makes sense to not keep them static as a single YCSB instance can be used for multiple benchmarking with different traces. Change incorporated.

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
shivam-maharshi: Moved

rest/src/test/resources/workload_rest
shivam-maharshi: This is needed now for the Integration Test cases. Hence hasn't been removed.

risdenk:
@shivam-maharshi - Can you ping this thread after you update? Thanks!

shivam-maharshi:
@risdenk I have incorporated all your feedbacks except for one (Merging RestWorkload into CoreWorkload.). Please take a look.

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
shivam-maharshi: I agree with you that it would be nice to have RestWorkload merged into CoreWorkload but there are some concerns/points.
1. In order to benchmark REST-full web services we must already have URL endpoints to benchmark. This is why we have URL maps in RestWorkload. However the way core workload is written currently, it generates keys to benchmark instead of reading them from a collection of predefined keys. Which definitely makes sense for DBs but not for web services.
2. Considering the point above, if we still merge RestWorkload into CoreWorkload, it would require us to distinguish the key generation logic on the basis of some config value which would make the CoreWorkload code ugly and couple it with logic specific with a particular binding.
3. Delete Operations are not supported in CoreWorload, they need special handling for WebServices to benchmark.
4. Since CoreWorkload is the central component, updating it will require an exhaustive testing for all the other bindings. This is precisely what I wanted to avoid.
5. To minimize code duplication RestWorkload extends CoreWorkload. Hence most future updates in CoreWorkload will be propagated to RestWorkload automatically thereby reducing extra code efforts.

I understand the urge to have a single workload but given the different nature of WebServices there are some inherent differences from databases which makes this task a bit non-trivial. I am open to further improve this design in the future but for now, if we can get this pull request incorporated in the master then I will create another pull request with the improved design later.

shivam-maharshi:
@risdenk @busbey Please let me know if we need further modifications in this PR. Thanks!

rest/pom.xml
```diff
@@ -25,6 +25,12 @@
   <properties>
     <tomcat.version>8.0.28</tomcat.version>
     <jersey.version>2.6</jersey.version>
+    <httpclient.version>4.5.1</httpclient.version>
+    <httpcore.version>4.4.4</httpcore.version>
+    <junit.version>4.12</junit.version>
+    <system-rules.version>1.16.0</system-rules.version>
+    <wiremock.version>1.57</wiremock.version>
+    <skipTests>true</skipTests>
```
risdenk: Any reason skipTests is here?

rest/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+      <version>${tomcat.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.tomcat.embed</groupId>
+      <artifactId>tomcat-embed-websocket</artifactId>
+      <version>${tomcat.version}</version>
+    </dependency>
+  </dependencies>
+  
+  <build>
```
risdenk: Does this plugin need to be here?

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
risdenk: Fair enough we can always improve it later.

rest/src/test/resources/workload_rest
risdenk: ok thanks for the explanation.

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
```diff
@@ -0,0 +1,300 @@
...
+      double frac = Double.parseDouble(p.getProperty(ExponentialGenerator.EXPONENTIAL_FRAC_PROPERTY,
+          ExponentialGenerator.EXPONENTIAL_FRAC_DEFAULT));
+      keychooser = new ExponentialGenerator(percentile, recordCount * frac);
+    } else if (requestDistrib.compareTo("uniform") == 0) {
+      keychooser = new UniformIntegerGenerator(0, recordCount - 1);
+    } else if (requestDistrib.compareTo("zipfian") == 0) {
+      keychooser = new ZipfianGenerator(recordCount, zipfContant);
+    } else if (requestDistrib.compareTo("latest") == 0) {
+      // TODO: To be implemented.
+      System.err.println("Latest request distribution is not supported");
```
risdenk: This should exit instead of returning null?

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
```diff
@@ -0,0 +1,300 @@
...
+		        .parseDouble(p.getProperty(DELETE_PROPORTION_PROPERTY, DELETE_PROPORTION_PROPERTY_DEFAULT));
+    if (deleteproportion > 0)
+      operationChooser.addValue(deleteproportion, "DELETE");
+    return operationChooser;
+  }
+
+  private static NumberGenerator getKeyChooser(String requestDistrib, int recordCount, double zipfContant,
+      Properties p) throws WorkloadException {
+    NumberGenerator keychooser = null;
+    if (requestDistrib.compareTo("exponential") == 0) {
```
risdenk: each of these `.compareTo` should just be `.equals()`?

risdenk: May even be cleaner as a switch case

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
risdenk: The error is printed without throwing a WorkloadException

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
```diff
@@ -0,0 +1,300 @@
...
+      }
+    return fieldLengthGenerator;
+  }
+
+  /**
+   *  Reads the trace file and returns a URL map.
+   */
+  private static synchronized Map<Integer, String> getTrace(String filePath, int recordCount)
+      throws WorkloadException {
+    Map<Integer, String> urlMap = new ConcurrentHashMap<Integer, String>();
```
risdenk: Since this is a synchronized method does this need to be ConcurrentHashMap?

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+    return getStatus(responseCode);
+  }
+
+  @Override
+  public Status update(String table, String endpoint, HashMap<String, ByteIterator> values) {
+    int responseCode;
+    try {
+      responseCode = httpExecute(new HttpPut(urlPrefix + endpoint), values.get("data").toString());
+    } catch (Exception e) {
```
risdenk: This is catching all exceptions and not actually printing what happened? Would be nice to log the actual exception?

risdenk: Might be able to just add the logging to the handleExceptions method

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    @Override
+    public void run() {
+      try {
+        Thread.sleep(timeout);
+        this.timedout.setIsSatisfied(true);
+      } catch (InterruptedException e) { 
+        // Do nothing. Kept just to remove Check-style build error.
+        byte a = 0;
```
risdenk: Hmmm wonder if there is a better way to do this.

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+  @Rule
+  public final ExpectedSystemExit exit = ExpectedSystemExit.none();
+
+  private static int port = 8080;
+  private static Tomcat tomcat;
+  private static final String WORKLOAD_FILEPATH =  IntegrationTest.class.getClassLoader().getResource("workload_rest").getPath();
+  private static final String TRACE_FILEPATH = IntegrationTest.class.getClassLoader().getResource("trace.txt").getPath();
+  private static final String ERROR_TRACE_FILEPATH = IntegrationTest.class.getClassLoader().getResource("error_trace.txt").getPath();
+  private static final String RESULTS_FILEPATH = IntegrationTest.class.getClassLoader().getResource("").getPath() + "results.txt";
```
risdenk: the `getResource("")` calls shouldn't be to `""` they should be to the proper file or directory.

rest/pom.xml
shivam-maharshi: Removed. An error during commit. Build was passing.

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
shivam-maharshi: I had the same opinion but kept it consistent with the older CoreWorkload code. Updated to switch statements now.

core/src/main/java/com/yahoo/ycsb/workloads/RestWorkload.java
shivam-maharshi: Handled - `throw new WorkloadException("Latest request distribution is not supported for RestWorkload.");`

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
shivam-maharshi: Done

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
shivam-maharshi: By updating checkstyle.xml with the below snippet a block now can have some text as opposed to just code to be considered valid:
`<module name="EmptyBlock">
     <property name="option" value="text"/>
 </module>`

rest/src/test/java/com/yahoo/ycsb/webservice/rest/IntegrationTest.java
shivam-maharshi: The file results doesn't exist hence we can't call `getResource("results.txt");`. Since we need the folder path to /src/test/resources even `getResource("resources")` doesn't work. Hence I've used a period - '.' signifying the current target directory.

shivam-maharshi:
@risdenk I have incorporated all your feedbacks. The build is failing due to some Cassandra test cases. Can you please take a look. 

Additionally, someone reached out to me asking for REST Web Services support in YCSB (https://github.com/shivam-maharshi/YCSB4WebServices/issues/1). It would be great if we can get this PR accepted soon so people can make use of it.

risdenk:
@shivam-maharshi - kicked off the test again and looks like it failed related to the rest piece:

```
Failed tests:   insert_200(com.yahoo.ycsb.webservice.rest.RestClientTest): expected:<Status [name=OK, description=The operation completed successfully.]> but was:<Status [name=ERROR, description=The operation failed.]>

Tests run: 26, Failures: 1, Errors: 0, Skipped: 0
```

shivam-maharshi:
@risdenk This error occurs sometime in JDK8 and sometime in JDK7 intermittently. It because some request in the RestClientTest is taking more than 10 seconds to execute and hence timer thread kills it and throws an Error Status. Right now the timeout is 10 seconds which I thought would be enough. What do you propose we do to resolve this? I am going to increase the timeout to 20 seconds and see if that passes. However in long run this is bad to build runtime. If you have any suggestions to handle this elegantly, please let me know.

shivam-maharshi:
@risdenk All checks pass now. Please take a look and lemme know.

risdenk:
@shivam-maharshi Travis CI for whatever reason tends to test really slowly. This seems to cause a lot of issues with Accumulo as well. Raising the timeout shouldn't hurt right? If it the test finishes then the timeout has no effect.

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
```diff
@@ -0,0 +1,367 @@
...
+      return Status.SERVICE_UNAVAILABLE;
+    } else if (rc == 5) {
+      return Status.ERROR;
+    }
+    return Status.OK;
+  }
+
+  private int handleExceptions(Exception e, String url, String method) {
+    if (logEnabled) {
+      System.out.println(new StringBuilder(method).append(" Request: ").append(url).append(" | ")
```
risdenk: The `System.out` in a few places should be `System.err` to make sure that messages go to stderr. I think there are a few examples of this in other bindings.

shivam-maharshi: @risdenk Let me quickly fix the logging and squash all the commit with [rest] today itself. Will notify you with the next commit. Thanks!

shivam-maharshi:
@risdenk Incorporated feedbacks and squashed all the commits into one. Please take a look and lemme know.

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
```diff
@@ -0,0 +1,370 @@
...
+  @Override
+  public Status insert(String table, String endpoint, HashMap<String, ByteIterator> values) {
+    int responseCode;
+    try {
+      responseCode = httpExecute(new HttpPost(urlPrefix + endpoint), values.get("data").toString());
+    } catch (Exception e) {
+      responseCode = handleExceptions(e, urlPrefix + endpoint, HttpMethod.POST);
+    }
+    if (logEnabled) {
+      System.out.println(new StringBuilder("POST Request: ").append(urlPrefix).append(endpoint)
```
risdenk: `System.out` instead of `System.err`

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
```diff
@@ -0,0 +1,370 @@
...
+  @Override
+  public Status delete(String table, String endpoint) {
+    int responseCode;
+    try {
+      responseCode = httpDelete(urlPrefix + endpoint);
+    } catch (Exception e) {
+      responseCode = handleExceptions(e, urlPrefix + endpoint, HttpMethod.DELETE);
+    }
+    if (logEnabled) {
+      System.out.println(new StringBuilder("DELETE Request: ").append(urlPrefix).append(endpoint)
```
risdenk: `System.out` instead of `System.err`

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
```diff
@@ -0,0 +1,370 @@
...
+  @Override
+  public Status read(String table, String endpoint, Set<String> fields, HashMap<String, ByteIterator> result) {
+    int responseCode;
+    try {
+      responseCode = httpGet(urlPrefix + endpoint, result);
+    } catch (Exception e) {
+      responseCode = handleExceptions(e, urlPrefix + endpoint, HttpMethod.GET);
+    }
+    if (logEnabled) {
+      System.out.println(new StringBuilder("GET Request: ").append(urlPrefix).append(endpoint)
```
risdenk: `System.out` instead of `System.err`

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
```diff
@@ -0,0 +1,370 @@
...
+  @Override
+  public Status update(String table, String endpoint, HashMap<String, ByteIterator> values) {
+    int responseCode;
+    try {
+      responseCode = httpExecute(new HttpPut(urlPrefix + endpoint), values.get("data").toString());
+    } catch (Exception e) {
+      responseCode = handleExceptions(e, urlPrefix + endpoint, HttpMethod.PUT);
+    }
+    if (logEnabled) {
+      System.out.println(new StringBuilder("PUT Request: ").append(urlPrefix).append(endpoint)
```
risdenk: `System.out` instead of `System.err`

rest/src/main/java/com/yahoo/ycsb/webservice/rest/RestClient.java
shivam-maharshi: @risdenk these are not error messages but just information logs. That's why I didn't update them since they will carry the messages of all the HTTP codes including success like 200, 204 etc.

shivam-maharshi: @risdenk if that's the norm we follow in YCSB then lemme make this change real quick and notify you back.

shivam-maharshi:
@risdenk I have updated all System.out to System.err. On of the three builds failed due to "build termination". Please take a look and lemme know. Thanks!

risdenk:
@shivam-maharshi looks good to me. Thanks!

Issue #746, opened 2016-05-10
[UPDATE-FAILED], When using YCSB for MongoDB Benachmarking


tanveermadan:
Hi All,

I am getting [UPDATE-FAILED], when trying to do updates on MongoDB using YCSB tool

I am running below command
./bin/ycsb run mongodb-async -s -P workloads/workloada -p mongodb.url=mongodb://poc03:27091,opoc04:27091,poc05:27091/ycsb2?w=0\&readConcernLevel=local\&readPreference=primary

The results are as below
[OVERALL], RunTime(ms), 6780041.0
[OVERALL], Throughput(ops/sec), 147.49173345706907
[CLEANUP], Operations, 1.0
[CLEANUP], AverageLatency(us), 1305.0
[CLEANUP], MinLatency(us), 1305.0
[CLEANUP], MaxLatency(us), 1305.0
[CLEANUP], 95thPercentileLatency(us), 1305.0
[CLEANUP], 99thPercentileLatency(us), 1305.0
[UPDATE-FAILED], Operations, 499272.0
[UPDATE-FAILED], AverageLatency(us), 36.04590283452707
[UPDATE-FAILED], MinLatency(us), 4.0
[UPDATE-FAILED], MaxLatency(us), 9783.0
[UPDATE-FAILED], 95thPercentileLatency(us), 77.0
[UPDATE-FAILED], 99thPercentileLatency(us), 150.0
[READ], Operations, 500728.0
[READ], AverageLatency(us), 13490.998673930757
[READ], MinLatency(us), 237.0
[READ], MaxLatency(us), 1.3918207E7
[READ], 95thPercentileLatency(us), 44575.0
[READ], 99thPercentileLatency(us), 76735.0
[READ], Return=OK, 500728
[UPDATE], Operations, 0.0
[UPDATE], AverageLatency(us), NaN
[UPDATE], MinLatency(us), 9.223372036854776E18
[UPDATE], MaxLatency(us), 0.0
[UPDATE], 95thPercentileLatency(us), 0.0
[UPDATE], 99thPercentileLatency(us), 0.0
[UPDATE], Return=NOT_FOUND, 499272

I have checked the mongodb log file and find entries for the update but dont see an issue

[...]

PR #759, opened 2016-06-03
[mongodb] return OK on updates with w=0

kruthar:
in #746 it was found that when using the `mongodb-async` binding with a write concern of `w=0` then update operations will always return `-1` which in turn reports an operation status of `NOT_FOUND`.

Since you cannot tell what the result of the update operation is whne you fire and forget, it is reasonable to just return success for all update operations in the `mongodb-async` binding when using a write concern of `w=0`.


busbey:
would BATCHED_OK be more appropriate? @allanbank, have an opinion?

I'm +1 with the change as is, or with BATCHED_OK instead.

allanbank:
@busbey - I think the OK status is a closer fit than BATCHED_OK.  The operation is successful to the level of durability that the user requested.

busbey:
Works for me!

PR #773, opened 2016-06-21
[couchbase2] Update Java SDK version, query for workloade

ingenthr:
This PR has a dependency update.

Atop that, there's a change to the query to use RAW results.  More in the commit message there.


ingenthr:
I looked at the build check, it seems to be a missing dependency from the cassandra binding?  I don't think it's related to my change.

daschl:
@busbey what do you think? I'm 👍  on the change :)

busbey:
Let's retry the build failures and make sure it's just flakeys

ingenthr:
Woo, pass that time @busbey !

ingenthr:
Thanks @busbey !

PR #774, opened 2016-06-22
[kudu] Update Kudu binding to 0.9.0

danburkert:
This updates the Kudu client version to the latest published (0.9.0), and fixes a scan performance regression introduced in f66bb1484b0. cc @jdcryans @toddlipcon


toddlipcon:
lgtm (though I'm not a YCSB maintainer)

busbey:
reviews from non-committers are binding in YCSB.

thanks for the contrib and the review!

busbey:
in the future, please start commit messages with an indicator of the modules impacted. In this case, ideally the message would start with "[kudu] Update kudu..."

Issue #495, opened 2015-11-12
Cassandra 2.1+ Binding fails on windows due to timeout


busbey:
Could just be that my build box for windows is underpowered. It's a eee pc with ~1GB memory and an Atom N550@1.5GHz.

[ycsb-0.5.0-rc2-failure-cassandra21.txt](https://github.com/brianfrankcooper/YCSB/files/32527/ycsb-0.5.0-rc2-failure-cassandra21.txt)

busbey:
I don't know if the timeout is adjustable. If it isn't, maybe we could add a profile for resource intensive tests?

risdenk:
Does it make sense to move tests that require a DB/cluster to integration tests and use something like https://maven.apache.org/surefire/maven-failsafe-plugin/? This could allow certain tests to be categorized and run separately.

busbey:
Categorizing the tests would be nice. But I think this particular issue can be closed out; I just finished successfully building with this test in place on a Windows 10 machine with a Celeron N3050 and 2GB of memory. That's still low end enough for me to be comfortable.

busbey:
started seeing this on my windows build laptop again.

risdenk:
This seems to happen on Travis CI quite often. Not sure it is only Windows based.

risdenk:
Found that the timeout is adjustable and will open a PR in a few minutes. This has been failing a lot on recent travis builds.

PR #776, opened 2016-06-23
add sigar-dist dependency for maven test on Cassandra 2.2+ and tracing function

jixuan1989:
* Firstly, I notice that when using `mvn -pl com.yahoo.ycsb:cassandra2-binding -am   test` on current YCSB version, there is an exception: `no libsigar-universal64-macosx.dylib in java.library.path`. 
It is because Cassandra uses Sigar for collecting system information and Sigar requires some native libs. 
So, I modify the pom.xml to automatically download sigar files, unzip them and add them into java classpath.

* A new switch for enable/disable Cassandra tracing function. when `cassandra.tracing=true`, Cassandra will trace the traces of read/write operations. It is helpful for debuging and optimizing Cassandra.  (when `cassandra.tracing=false` (default value), nothing changed.)


jixuan1989:
very strange. It seems all the ERROR msgs are not from my modifications...
And I have pass the test locally.

busbey:
I restarted the failed jdk8 build let's see how it goes.

risdenk:
@jixuan1989 Are there any issues with the sigar-dist changes when compiling on Windows? (sorry don't have a Windows machine to test :/)

jixuan1989:
@risdenk , I have tested it on Windows 10. The result is:

1. When I first ran `mvn -pl com.yahoo.ycsb:cassandra2-binding -am test`, the maven  downloaded sigar-dist and unzipped them. The test failed and said: `no sigar-amd64-winnt.dll in java.library.path`.  However, when I checked the target folder, I could find the dll file. I think the reason is that Windows does not support dynamically loads dll files (at least for Maven).

2.  Fortunately, when I ran `mvn -pl com.yahoo.ycsb:cassandra2-binding -am test` for the second time,  the  test program passed! See the screenshot.
I think the reason is that maven has unzipped the file into the target folder, so that Windows does not think that the dll file is a new one, so Windows can load it successfully.
default

jixuan1989:
It means that when running test on Windows, you have to run two times. Currently I do not know how to repair it by maven.

cassandra2/pom.xml
```diff
@@ -62,6 +62,13 @@ LICENSE file.
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+    <!-- only for Cassandra test (Cassandra 2.2+ uses Sigar for collecting system information, and Sigar requires some native lib files) -->
+	<dependency>
+	    <groupId>org.hyperic</groupId>
+	    <artifactId>sigar-dist</artifactId>
+	    <version>1.6.4.129</version>
+	    <type>zip</type>
```
busbey: if this is for test, should this be at test scope?

cassandra2/pom.xml
```diff
@@ -78,4 +85,60 @@ LICENSE file.
...
+	 -->
+ 	<build>
+	 	<plugins>
+	 		<plugin>
+			    <groupId>org.apache.maven.plugins</groupId>
+			    <artifactId>maven-dependency-plugin</artifactId>
+			    <executions>
+			      <execution>
+			        <id>unpack-sigar</id>
+			        <phase>test<!-- or any other valid maven phase --></phase>
```
busbey: if this is needed in test, then the unpacking should happen before test. probably process-test-resources

busbey:
please update the commit message to start with a brief description that includes the impacted module.

i.e. "[cassandra2] add tracing and sigar-dist for tests.

busbey:
> When I first ran mvn -pl com.yahoo.ycsb:cassandra2-binding -am test, the maven downloaded sigar-dist and unzipped them. The test failed and said: no sigar-amd64-winnt.dll in java.library.path. However, when I checked the target folder, I could find the dll file. I think the reason is that Windows does not support dynamically loads dll files (at least for Maven).

This sounds like we need to configure surefire for the cassandra module to use fork mode once or every so that it has its own JVM instead of relying on loading into the JVM for the maven run.

cassandra2/pom.xml
jixuan1989: yep, add a scope is better.

jixuan1989:
I have fixed commit logs and add test scope declaration, the new PR is #783

jixuan1989:
Finally, I have fixed commit logs and add test scope declaration in this PR. Thanks for @busbey 's help.

cassandra2/pom.xml
busbey: With this still bound to the test phase does running things work without running Maven twice?

jixuan1989: The details are:
1. On Mac and Linux, everything is ok. You just need to run once.
2. On Windows:
  * if your project is clean, i.e., you never run `mvn test` on your project, then you need to run `mvn test` twice. 
  * if you have ran `mvn test`, you do not need to run `mvn test` twice any longer until you run `mvn clean`...

busbey: can you try updating the phase to "process-test-resources"?

cassandra2/pom.xml
```diff
@@ -78,4 +86,60 @@ LICENSE file.
...
+	 -->
+ 	<build>
+	 	<plugins>
+	 		<plugin>
+			    <groupId>org.apache.maven.plugins</groupId>
+			    <artifactId>maven-dependency-plugin</artifactId>
+			    <executions>
+			      <execution>
+			        <id>unpack-sigar</id>
+			        <phase>process-test-resources<!-- or any other valid maven phase --></phase>
```
jixuan1989: @busbey  That is great!! I have add the "process-test-resources" phase in both the plugins and dependency. Now, even you run it on Windows, it can still be passed once. 

The command is `mvn -pl com.yahoo.ycsb:cassandra2-binding -am  clean process-test-resources test`

jixuan1989: Thanks, please let me know if there are any other issues.

cassandra2/pom.xml
```diff
@@ -62,6 +62,14 @@ LICENSE file.
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+    <!-- only for Cassandra test (Cassandra 2.2+ uses Sigar for collecting system information, and Sigar requires some native lib files) -->
+	<dependency>
+	    <groupId>org.hyperic</groupId>
+	    <artifactId>sigar-dist</artifactId>
+	    <version>1.6.4.129</version>
+	    <type>zip</type>
+	    <scope>process-test-resources,test</scope>
```
busbey: I didn't think this is a valid scope (ref [maven docs on dependency scopes](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope)). I believe this should just be `test` scope.

jixuan1989: I think you are right. I confused the "phase" and "lifecycle".

jixuan1989: I have corrected the scope declaration. Now it works well. Thanks for your pointing out the error. :)

cassandra2/pom.xml
```diff
@@ -78,4 +86,60 @@ LICENSE file.
...
+			    </executions>
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>2.8</version>
+				<configuration>
+				<systemProperties>
+				<property>
+					<name>java.library.path</name>
```
busbey: surefire issues a warning that java.library.path can't be set this way and we need to use the argLine configuration value.

busbey:
I applied this change on a windows box and started seeing #495 again. I think I have a change for that, I'm not sure if it's worth getting in place prior to landing this patch.

the current PR is also out of date with master. I have the conflicts resolved locally, but it'd be nice if you could rebase to the current master so that we can merge via the website.

jixuan1989:
Sorry for reply so late. I have merged the pom.xml with the latest Master branch.

By the way, I can still pass the test both on Windows  and Mac OS (but Windows will spend more time to connect to the cassandra clsuter), 
though  the WARN message `"java.library.path cannot be set as system property, use <argLine>-Djava.library.path=...<argLine> instead"` still exists. As a result, I can not reproduce #495.

risdenk:
@jixuan1989 sorry for the delay on this. The `java.library.path cannot be set as a system property...` can probably be fixed my using argLine instead of systemProperty in the surefire plugin. http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#argLine Have you tried that?

jixuan1989:
@risdenk   great, it works! 
badly I messed the git logs.. I will repair it...

jixuan1989:
um.... I have merged it with the master branch,  and I can pass the test process by command:`mvn -pl com.yahoo.ycsb:cassandra-binding -am test`. I do not know why the auto-test by github failed. (maybe because start cassandra-server test mode is slow, so junit thinks it failed because timeout)

risdenk:
@jixuan1989 I kicked off the travis build again to see if that helps.

risdenk:
@jixuan1989 - The test passed when I restarted it.

@busbey - Any more comments on this?

busbey:
can we leave out the whitespace changes to `bin/ycsb.bat`?

jixuan1989:
@busbey fixed! The previous  problem about the whitespace is caused by the Windows OS. Now I switched to Mac OS and have fixed it.

risdenk:
@jixuan1989 took another look at this and looks like only the sigar-dist piece is in here. Did the tracing get removed?

jixuan1989:
@risdenk  yep, I think sigar-dist is more important because it impacts the test. Tracing is not an important function and it requires to modify CassandraClient.java. If you think that the content should match with the title of PR, I can commit the modification.

In fact, tracing is more important for Cassandra research (serve for log based methods, which are my research works..), rather than performance benchmark.

jixuan1989:
Now, I have added the tracing function. Then, I ask for approving this PR  quickly if possible, because the tracing function has to modify CassandraCQLClient.java. If the PR is delayed, I have to merge it with the master branch once again and again ( though it is not a hard work :D).

Besides, I modify the README.md for introducing how to enable tracing and what the tracing function is (from datastax document).

jixuan1989:
By the way, in my observation, Cassandra test is time-consuming and resource-consuming, we can find that sometimes the test failed because of timeout. 

Maybe we can modify the `cassandra.connecttimeoutmillis` property to decrease the probability of timeout. But I am not sure whether it works, because timeout is not a stable phenomenon, and the problem does not belong to the discussion scope of this PR.

risdenk:
@jixuan1989 about the Cassandra test timeout. I think I figured out a fix in #850

risdenk:
All changes look good to me. @busbey any last comments? If not I'll merge later today.

PR #783, opened 2016-06-29
[cassandra2]add sigar dependency in test scope and add tracing function

jixuan1989:
This pull request is a modification from  #776, while in this PR, 
1. I set the sigar-dist dependency as test scope. 
2. I clean the commit logs and simplify them.
(both of them are suggested by @busbey )

### DIFF
--- a/cassandra2/pom.xml
+++ b/cassandra2/pom.xml
@@ -62,6 +62,14 @@ LICENSE file.
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+    <!-- only for Cassandra test (Cassandra 2.2+ uses Sigar for collecting system information, and Sigar requires some native lib files) -->
+	<dependency>
+	    <groupId>org.hyperic</groupId>
+	    <artifactId>sigar-dist</artifactId>
+	    <version>1.6.4.129</version>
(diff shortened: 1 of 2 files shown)


busbey:
please force-push over the branch for #776 instead of creating a new PR.

jixuan1989:
@busbey  Hm... I am sorry I am not an expert of Git. I feel confused about "force-push over the branch for..", what do you mean about "force-push"? Like this? 

1. Modify the pom.xml file.

2. `git add pom.xml`; git commit -m "comments.."

3. `git push --force  myRemoteURL myBranchName`

If so, when I push a modification to the branch (no matter whether I use --force or not), I think  a new commit-id will be generated, while the PR of #776 does not changes...

busbey:
ah. sorry I missed this comment over here.

the branch on your github for PR #776 is named 'cassandra-tracing-function'. even if the commit-ids change, github will associate the PR with the branch-name rather than the commit-ids.

I'm not sure if git force pushing works with a full remote url and branch name, I'm used to using a remote name (very often 'origin') and branch name.

jixuan1989:
That's ok. Thank you for your comments. I made it!!

Issue #772, opened 2016-06-20
Release 0.10.0


busbey:
Hi Folks!

I'm going to spin a release candidate shortly. If there's an in-progress PR that's close that you'd like to see included please try to get it in soon.

I'll comment with the current set of issues later this evening.

busbey:
Changes since 0.9.0:

* #670 [rados] add binding for RADOS of Ceph FS
* #737 [scripts] move to posix and windows specific shell scripts instead of python
* #751 [core] correct calculation of weighted histograms
* #755 [jdbc] add batched inserts to JDBC binding
* #758 [mongodb] added logback.xml to mongodb module
* #759 [mongodb] always return OK status when mongodb-async uses w=0
* #766 [cassandra2] ycsb prebuilt artifact doesn't include cassanda2 module (no commits)


#737 is going to require updates to our docs before final release.

it looks like we'll need to retest a minimum of the following: MySQL (via JDBC), mongo, mongo-async, cassandra2.

I'd like to get rados tested since it's new.

Anything else folks would like to see?

busbey:
I'm not sure how I missed this in reviewing #737, but I now notice that the python script for running ycsb is still in place. Should we leave this in parallel with the dedicated scripts for this release? I wouldn't want to keep it around much longer than that.

busbey:
RC1 is up, if folks want to start testing:

https://github.com/brianfrankcooper/YCSB/releases/tag/0.10.0-RC1

* @jaemyoun could you test out the rados binding?
* @allanbank will you have time to test out the mongodb bindings?
* @kruthar do you think you'll have time to re-check mysql with the recent jdbc changes?
* @cmatser think you could test out the new posix script?

Any volunteers for checking that the cassandra2 module works?

I'll get things tested on windows (both with the python wrapper and the new batch script).

sekikn:
FWIW, I tried jdbc-binding with PostgreSQL 9.5.2 and its JDBC4 Driver. All workloads succeeded with db.batchsize=100 :) https://gist.github.com/sekikn/cc6c7d034bc3343253df7061c75fb1bb

engineerakki:
I am testing the Cassandra2 module in 0.10.0RC1  and will update about it here.

cmatser:
[...]

PR #786, opened 2016-06-30
[couchbase2] logical error in decode()

WindzCUHK:
When dataintegrity=true, no field values in passed to verifyRow() for verify and keep on getting error.


busbey:
this looks correct to me. mind taking a quick look @daschl ?

daschl:
yup makes sense

busbey:
@daschl should I cherry-pick this into the upcoming 0.10.0-staging branch? It'll mean we'll need couchbase testing for #772.

daschl:
yes merge it please, I can do the testing for 0.10

daschl:
@busbey but if we do the testing can you please please also cherry-pick https://github.com/brianfrankcooper/YCSB/commit/5edb1d9e2c9e51ec1ea26e1504305a7090653787 over? Thanks!

busbey:
I think that commit requires the SDK update too? Can we take this discussion over to #772?

daschl:
@busbey this commit introduced a regression, fixing it right now.

PR #790, opened 2016-07-04
[scripts] Rados missing from properties file for new platform-specific scripts.

busbey:
The new Rados binding from 0.10.0 is missing an entry in the properties file used to handle the platform specific scripts.

If someone can review quickly, I'm fine with pulling this into the release for #772.


busbey:
ping @jaemyoun

jaemyoun:
@busbey thanks for your ping and the commit is fine.

busbey:
thanks for the review!

PR #791, opened 2016-07-04
[scripts, couchbase] deprecate the old 'couchbase' client in favor of 'couchbase2'.

busbey:
Point folks towards the new couchbase2 client instead of the old one, so we can eventually swap them out.

ping @daschl

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -249,6 +249,10 @@ def main():
              "cassandra-cql' clients are deprecated. If you are using "
              "Cassandra 2.X try using the 'cassandra2-cql' client instead.")
 
+    if binding == "couchbase":
+        warn("The 'couchbase' client has been deprecated. If you are using "
+             "Couchbase 4.0+ try using the 'couchbase2' client instead.")
+
     if is_distribution():
         db_dir = os.path.join(ycsb_home, binding + "-binding")
         # include top-level conf for when we're a binding-specific artifact.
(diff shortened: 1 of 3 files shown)


daschl:
@busbey I wonder if we can get rid of it completely because `couchbase2` is backwards compat and can do everything `couchbase` can do?

daschl:
Otherwise if we want to keep it for now then 👍 on the change!

busbey:
If someone who was using the old couchbase module can just swap in the couchbase2 module and use the same configs and talk to the same versions of couchbase2, just removing the old one and deprecating the name "couchbase2" would be fine.

daschl:
hm yeah well okay thats a problem since not the exact same options are supported. then let's keep it deprecated for a couple of versions and remove it at some point. I just brought it up since the 1.4 SDK series in couchbase is EOL right now.. 

the other thing I could look at is somehow supporting the "old" options and converting them to the new ones for backwards compat?

busbey:
probably easier to just deprecate it and pick a time to remove it. I think when we deprecated the old cassandra binding we gave it 6 months.

busbey:
ah, looks like #537 made it three months for cassandra (and we're behind schedule on actually removing).

also I forgot to have this one to put the notice in the README. I'll update the PR later today.

daschl:
👍

PR #793, opened 2016-07-05
[jdbc]Fix bug in constructing CREATE TABLE SQL statement

k1xme:
The name of primary key in the original statement conflicts with reserved keyword and is inconsistent with the specified table schema in the documentation.

Fixed.


busbey:
good catch! would you mind updating the commit message to make clear that this is for the JDBC driver by starting it with "[jdbc]" ?

k1xme:
@busbey Done~!

busbey:
thanks!

PR #797, opened 2016-07-10
[jdbc] Use `TEXT` for fields instead of VARCHAR

k1xme:
The original statement specify field type of VARCHAR, which might not be able to store the random-length string produced by YCSB and is also inconsistent with the schema in documentation.


busbey:
you should update the javadocs on the jdbc client as well, it claims that things are stored in VARCHAR still

risdenk:
@busbey @k1xme - Updated this PR to fix JavaDocs. New PR #910

PR #801, opened 2016-07-12
Add additional IntelliJ project file to Git ignore

jasontedor:
This commit adds the IntelliJ ipr extension to .gitignore. These project
files are created when running the Maven goal idea:idea.


busbey:
+1

PR #803, opened 2016-07-19
[couchbase2] Added support for ttl in Couchbase2

kunalsomani:
Time to live (TTL) is the amount of time until a document expires in Couchbase Server. Having this property will help user not to clean up DB manually.

### DIFF
--- a/couchbase2/README.md
+++ b/couchbase2/README.md
@@ -142,4 +142,4 @@ You can set the following properties (with the default settings applied):
    set to the number of physical cores. Setting higher than that will likely degrade performance.
  - couchbase.networkMetricsInterval=0: The interval in seconds when latency metrics will be logged.
  - couchbase.runtimeMetricsInterval=0: The interval in seconds when runtime metrics will be logged.
- - couchbase.objectExpiry=0: Object Expiry is the amount of time(second) until a document expires in Couchbase.
+ - couchbase.documentExpiry=0: Document Expiry is the amount of time(second) until a document expires in Couchbase.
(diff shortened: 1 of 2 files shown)


busbey:
out of all that noise, this is the bit you should care about:

> [ERROR] src/main/java/com/yahoo/ycsb/db/couchbase2/Couchbase2Client.java[96] (sizes) LineLength: Line is longer than 120 characters (found 121).

We do style enforcement on the couchbase2 client, so this will need to get cleaned up.

busbey:
looks reasonable, presuming a TTL of 0 means "never expire" (or whatever the default is when TTL isn't given).

what do you think @daschl?

daschl:
Yes, 0 is the default and means never expire. 

If we could name it `ttl` i'd prefer that, or even `expiry` since thats what the SDK calls it. Other than that I'm good with the change.

daschl:
@kunalsomani can you please just name it "expiry"? I don't like object, since the naming talks about documents, not objects :) .. sorry to be nitpicky but I think its important for usability

kunalsomani:
Name with expiry makes it difficult to understand to user about the system property, "couchbase.expiry" does not say about the document expiry property. We should have meaningful property so that user can understand about it. I chose this property as YCSB is using the similar property in memcached "memcached.objectExpirationTime" but with bigger name :)

Let me know your feedback and will change if required.

daschl:
@kunalsomani okay, then what about `couchbase.documentExpiry` ?

kunalsomani:
ping @daschl

daschl:
looks good to me now, thanks! :)

busbey:
the jdk8 build failed due to cassandra. restarted it.

thanks for the review @daschl!

kunalsomani:
@busbey  Can we merge this change? Jdk8 Build failed because of Cassandra.

busbey:
I'm trying to get a clean CI run. The Cassandra test has historically been finicky, but other PRs are passing it.

PR #805, opened 2016-07-19
Remove deprecated Cassandra clients

### LINKED ISSUE #794: remove deprecated Cassandra client
the old cassandra client was deprecated back in December (#537) with a notice that it would probably get pulled in March.

Now that it's been ~6 months, let's get it removed.

Personally, I think we should delete the current `cassandra` module, rename the `cassandra2` module to `cassandra`, and then make it so using `cassandra2` on the CLI still works but gives a deprecation warning that it's moved to the name `cassandra` now.

busbey:
Removes deprecated clients. Renames the cassandra2 module to cassandra.

closes #794


cmatser:
:+1:  Looks good to me.

PR #812, opened 2016-07-28
Add support for BangDB server

rith-iqlect:
For more info about BangDB, look here:  www.bangdb.com


busbey:
the CI builds appear to be because the bangdb client is not in maven central. is there some other repo the binding could point to?

rith-iqlect:
Hi Busbey,
I've noticed that too :-) Currently, the bangdb client jar is not available in any public repo. I will upload the jar to repo and make modifications in pom file.  Once that it done, I'll inform you.

PR #815, opened 2016-07-31
Optimise fillBytesImpl so it doesn't trigger an Exception on every byte[] generated.

peter-lawrey:
Exceptions are cheap but not that cheap. This will save an exception whenever the byte[] is not a multiple of 6. e.g. 100 bytes

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/BasicDB.java
+++ b/core/src/main/java/com/yahoo/ycsb/BasicDB.java
@@ -96,6 +96,19 @@ public class BasicDB extends DB
 		}
 	}
 
+  static final ThreadLocal<StringBuilder>  TL_STRING_BUILDER = new ThreadLocal<StringBuilder>() {
+    @Override
+    protected StringBuilder initialValue() {
+      return new StringBuilder();
+    }
+  };
+
+  static StringBuilder getStringBuilder() {
+    StringBuilder sb = TL_STRING_BUILDER.get();
+    sb.setLength(0);
+    return sb;
+  }
+
 	/**
 	 * Read a record from the database. Each field/value pair from the result will be stored in a HashMap.
 	 *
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/yahoo/ycsb/BasicDB.java
```diff
@@ -111,21 +124,23 @@ public Status read(String table, String key, Set<String> fields, HashMap<String,
 
 		if (verbose)
 		{
-			System.out.print("READ "+table+" "+key+" [ ");
+      StringBuilder sb = getStringBuilder();
```
busbey: Is this trying to save the string builder allocation that the existing code causes? Is the difference quantifiable?

core/src/main/java/com/yahoo/ycsb/BasicDB.java
```diff
@@ -111,21 +124,23 @@ public Status read(String table, String key, Set<String> fields, HashMap<String,
...
+			sb.append("READ ").append(table).append(" ").append(key).append(" [ ");
 			if (fields!=null)
 			{
 				for (String f : fields)
 				{
-					System.out.print(f+" ");
+          sb.append(f).append(" ");
```
busbey: Ah I see. We're rolling all the string ops into SB appends instead of calls to the output buffer.

core/src/main/java/com/yahoo/ycsb/BasicDB.java
```diff
@@ -80,7 +80,7 @@ public void init()
 		verbose=Boolean.parseBoolean(getProperties().getProperty(VERBOSE, VERBOSE_DEFAULT));
 		todelay=Integer.parseInt(getProperties().getProperty(SIMULATE_DELAY, SIMULATE_DELAY_DEFAULT));
 		randomizedelay=Boolean.parseBoolean(getProperties().getProperty(RANDOMIZE_DELAY, RANDOMIZE_DELAY_DEFAULT));
-		if (verbose)
+		if (verbose) synchronized(System.out)
```
busbey: Please include the braces for the added nesting.

Does this help if another thread is printing something other than. Properties?

core/src/main/java/com/yahoo/ycsb/BasicDB.java
```diff
@@ -111,21 +124,23 @@ public Status read(String table, String key, Set<String> fields, HashMap<String,
...
 			else
 			{
-				System.out.print("<all fields>");
+        sb.append("<all fields>");
 			}
 
-			System.out.println("]");
-		}
+      sb.append("]");
+      System.out.println(sb);
```
busbey: Does anything say that individual calls to S.o.println are threadsafe? Or are we just decreasing the odds of a collision?

busbey:
Great catch on the exception! They are definitely not cheap. :)

core/src/main/java/com/yahoo/ycsb/BasicDB.java
peter-lawrey: The main difference is minimising contention for System.out. We could create a new StringBuilder each time, and while not cheap it is small compared to the cost of writing to System.out

core/src/main/java/com/yahoo/ycsb/BasicDB.java
peter-lawrey: println() holds a lock on System.out while printing.  This means any single operation is thread safe, though multiple operations are not.

PR #826, opened 2016-08-19
Upgrade bigtable version for googlebigtable tests.

sduskis:


risdenk:
@sduskis Sorry for the delay on reviewing this. Looks like the failure was a bad Cassandra run. I kicked off the travis ci job again.

There are quite a few changes in this PR. Does the current bigtable binding work right now? Do v1 and v2 work at the same time?

risdenk:
@sduskis - Any plans to address the questions above? Also looks like this needs to be updated based on master changes.

sduskis:
The v1 and v2 APIs and protobuf objects are in different cloud-bigtable-client versions.  The v1 API will be removed at some point in the future, probably measured in months.  One alternative is to have separate maven projects for the v1 and v2 APIs.

For now I'll figure out how to change this into more manageable chunks starting with changes that can be applied to the v1 version.

PR #828, opened 2016-08-23
[geode] Update to Apache Geode version 1.0.0-incubating.M3

charliemblack:
Update to Apache Geode version 1.0.0-incubating.M3
Change the data storage to use the same technique that Geode uses for JSON which yields less memory in the servers.
Update the Readme to use let the servers know that we are using PDX so queries look better.


metatype:
LGTM

busbey:
thanks or the patch @charliemblack and for the review @metatype!

PR #842, opened 2016-09-20
[jdbc] Assorted cleanup/fixes around batching and supporting different databases

joshelser:
A couple of things included in this PR. Happy to split them out if that is preferred. I just came across all of these in some recent testing.

* Builds on @enis's commit which didn't make it upstream. Implements a "flavor" for Apache Phoenix (which only provides `UPSERT` to write data, no `INSERT`)
* Some basic cleanup of JdbcDBClient (rip out all of the flavors logic into their own files)
* De-couples batching APIs and autocommit. Again, trying to support Phoenix better, but not break JDBC semantics.
* Finally, fixes an issue with the existing batching support where the CoreWorkload treated `Status.BATCHED_OK` as an error (not the same as `Status.OK`)

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/DBWrapper.java
+++ b/core/src/main/java/com/yahoo/ycsb/DBWrapper.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
+ * Copyright (c) 2010 Yahoo! Inc., 2016 YCSB contributors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You
@@ -184,7 +184,7 @@ public class DBWrapper extends DB
   private void measure(String op, Status result, long intendedStartTimeNanos,
       long startTimeNanos, long endTimeNanos) {
     String measurementName = op;
-    if (result != Status.OK) {
+    if (result == null || !result.isOk()) {
       if (this.reportLatencyForEachError ||
           this.latencyTrackedErrors.contains(result.getName())) {
         measurementName = op + "-" + result.getName();
--- a/core/src/main/java/com/yahoo/ycsb/Status.java
+++ b/core/src/main/java/com/yahoo/ycsb/Status.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015 YCSB contributors All rights reserved.
+ * Copyright (c) 2016 YCSB contributors All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You
@@ -77,6 +77,14 @@ public class Status {
     } else if (!name.equals(other.name))
       return false;
     return true;
+  }
+
+  /**
+   * Is {@code this} a passing state for the operation: {@link Status#OK} or {@link Status#BATCHED_OK}.
+   * @return true if the operation is successful, false otherwise
+   */
+  public boolean isOk() {
+    return this == OK || this == BATCHED_OK;
   }
 
   public static final Status OK = new Status("OK", "The operation completed successfully.");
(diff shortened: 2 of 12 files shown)


joshelser:
@busbey this is the changeset I was referring to earlier today in private chat.

joshelser:
Also, I forgot to mention the testing I had done..

1. Verified that both thick and thin JDBC drivers work with expected performance after these changes
2. MySQL works before and after changes with similar perf (workloada)

jdbc/src/main/java/com/yahoo/ycsb/db/flavors/PhoenixDBFlavor.java
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
```
risdenk: This is a new file?

jdbc/src/main/java/com/yahoo/ycsb/db/flavors/DefaultDBFlavor.java
```diff
@@ -0,0 +1,98 @@
+/**
+ * Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
```
risdenk: This is a new file?

jdbc/src/main/java/com/yahoo/ycsb/db/flavors/DBFlavor.java
```diff
@@ -0,0 +1,69 @@
+/**
+ * Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
```
risdenk: New file?

jdbc/src/main/java/com/yahoo/ycsb/db/StatementType.java
```diff
@@ -0,0 +1,110 @@
+/**
+ * Copyright (c) 2010 - 2016 Yahoo! Inc. All rights reserved.
```
risdenk: New file?

core/src/main/java/com/yahoo/ycsb/DBWrapper.java
```diff
@@ -184,7 +184,7 @@ public Status scan(String table, String startkey, int recordcount,
   private void measure(String op, Status result, long intendedStartTimeNanos,
       long startTimeNanos, long endTimeNanos) {
     String measurementName = op;
-    if (result != Status.OK) {
+    if (result == null || !result.isOk()) {
```
risdenk: Update copyright on changed file?

core/src/main/java/com/yahoo/ycsb/Status.java
```diff
@@ -79,6 +79,14 @@ public boolean equals(Object obj) {
     return true;
   }
 
+  /**
```
risdenk: Update copyright on changed file?

joshelser:
Hi @risdenk! Thanks for taking a look -- it's greatly appreciated.

You're entirely right about the license headers: it was not clear to me what the copyright notice should be on new files. I just made a best guess on context. I am happy to fix with proper guidance.

risdenk:
New files are usually:
```
Copyright (c) 2016 YCSB contributors. All rights reserved.
```
An example of one that got modified would be something like this:
```
Copyright (c) 2010 Yahoo! Inc., 2016 YCSB contributors. All rights reserved.
```

I'm just reciting what I've seen previously. Not sure this is set in stone. Maybe @busbey has an opinion?

busbey:
yep, those suggested copyright statements look correct.

joshelser:
> yep, those suggested copyright statements look correct.

Great. Thanks @busbey and @risdenk! I'll fix these changes shortly with those recommendations.

joshelser:
Just updated things with more correct-er license headers. Thanks again for the close look, @risdenk!

risdenk:
@joshelser thanks for the PR!

joshelser:
Thanks for the review and merge, @risdenk !

PR #712, opened 2016-04-16
[build, core, accumulo, couchbase2, hbase10, orientdb] fixes for jdk9

busbey:
* in couchbase2 work around ReactiveX/RxJava#3824
* accumulo test hangs, so skip tests
* hbase10 test crashes with unknown class version, so skip tests
* core jar-with-deps fails, so keep them broken out
* add jdk9 to travis

fixes #704, fixes #705

Issue #705, opened 2016-04-13
JDK 9, 10, and 11 support for YCSB


risdenk:
Since JDK 9 is in early access, it would be good to start testing with it and see what changes may be required to support it.

risdenk:
There is an early report in #704 where the build fails.

busbey:
:shakes_fist: No tests after rebase!!! arg. taking a look.

risdenk:
The fixes from PR #712 worked need to extend for a few other bindings. Just ran on latest 9-ea 138 and got the following:

Solr 5.x - need to exclude from jdk 9
JDBC - redundant throws error from checkstyle

risdenk:
@busbey I'll have a PR in a minute to fix the above two issues.

busbey:
sweet. I just verified it was a flakey build and not that #712 broke things.

risdenk:
Looks like there is an issue with the `rest` binding. The other bindings look good. I ran `mvn clean package` with the following environment:

```bash
# java -version
java version "9-ea"
Java(TM) SE Runtime Environment (build 9-ea+149)
Java HotSpot(TM) 64-Bit Server VM (build 9-ea+149, mixed mode)

# mvn --version
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T11:41:47-05:00)
Maven home: /usr/local/Cellar/maven/3.3.9/libexec
Java version: 9-ea, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk-9.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.12.3", arch: "x86_64", family: "mac"
```

risdenk:
I opened #914 to track fixing the rest binding.

risdenk:
An update here, I just tested master and got no failures with the following:

```
#java -version
java version "9-ea"
Java(TM) SE Runtime Environment (build 9-ea+163)
Java HotSpot(TM) 64-Bit Server VM (build 9-ea+163, mixed mode)

#mvn --version
Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T15:39:06-04:00)
Maven home: /usr/local/Cellar/maven/3.5.0/libexec
Java version: 9-ea, vendor: Oracle Corporation
[...]

PR #849, opened 2016-09-27
[solr,solr6] Add Solr 6 binding

### LINKED ISSUE #715: Add Solr 6 binding

risdenk:
Closes #715

### DIFF
--- a/bin/bindings.properties
+++ b/bin/bindings.properties
@@ -58,7 +58,8 @@ rados:com.yahoo.ycsb.db.RadosClient
 redis:com.yahoo.ycsb.db.RedisClient
 riak:com.yahoo.ycsb.db.riak.RiakKVClient
 s3:com.yahoo.ycsb.db.S3Client
-solr:com.yahoo.ycsb.db.SolrClient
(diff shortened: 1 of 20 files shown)


risdenk:
@busbey can you review? This would help move forward JDK 9 support (#705 and #712)

risdenk:
bump?

risdenk:
bump?

risdenk:
bump?

busbey:
jdk8 was my big question. it looks like compilation with jdk7 is fine, but running (including tests) requires jdk8. Do I have that right?

risdenk:
It should be ok compiling with jdk 7. I haven't tested this locally but it seems to compile in travis with jdk 7. As long as we don't use jdk 8+ things (lambdas,etc) in the client then I think it should be ok. Running (including tests) requires JDK8.

risdenk:
@busbey good catch. I updated the PR with `bin/bindings.properties` updated.

risdenk:
Just checked and YCSB with Solr 6 compiles locally with JDK 7.

PR #850, opened 2016-09-27
[cassandra] Increase Cassandra test timeout and add logging for test failures

### LINKED ISSUE #495: Cassandra 2.1+ Binding fails on windows due to timeout

risdenk:
Added slf4j-simple so we can see what is happening during the Cassandra tests and get an idea why it fails. The timeout was increased from 10 seconds to 120 seconds. I tried 60 seconds but on Docker it still failed. Bumping up to 120 seems to have worked.

Fixes #495

### DIFF
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -57,6 +57,12 @@ LICENSE file.
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>1.7.21</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
--- a/cassandra/src/test/java/com/yahoo/ycsb/db/CassandraCQLClientTest.java
+++ b/cassandra/src/test/java/com/yahoo/ycsb/db/CassandraCQLClientTest.java
@@ -54,6 +54,9 @@ import java.util.Set;
  * Integration tests for the Cassandra client
  */
 public class CassandraCQLClientTest {
+  // Change the default Cassandra timeout from 10s to 120s for slow CI machines
+  private final static long timeout = 120000L;
+
   private final static String TABLE = "usertable";
   private final static String HOST = "localhost";
   private final static int PORT = 9142;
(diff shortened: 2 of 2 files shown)


risdenk:
@busbey can you take a peek?

risdenk:
@jixuan1989 if you could test/review this change it would be greatly appreciated. Reviews and tests can be completed by anyone.

jixuan1989:
@risdenk  Thanks, no problem. I will try it, but maybe a little delay because the holiday of National Day of China is coming.. 
Anyway, I can try it ASAP.

jixuan1989:
I have tested the timeout setting and it really works. 

However, though we can get more information  in the test process by add slf4j-simple, I think loggers print too many log records.. 
 
So do you really want to print these logs all the time? or just add them when you want to debug?

risdenk:
@jixuan1989 Thanks! I agree it prints too much currently. I'll remove the logging since the timeout is the important piece.

risdenk:
Uggghhhh looks like we are going to need the logging: 

```
Datacenter: datacenter1; Host: localhost/127.0.0.1; Rack: rack1
Killed
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.8:test (default-test) on project cassandra-binding: There are test failures.
```

There is no information as to why Cassandra was Killed. Memory maybe?

jixuan1989:
I have tested many times but can not reproduce the upper  error case... I feel a little strange that why Cassandra does not  print anything except for `killed`. 

In general, even if we need the logs for diagnosis, the ERROR level is enough rather than INFO for cassandra (if we just want to know what causes an error, rather than observe the startup progress).  

e.g., 
`[main] ERROR org.cassandraunit.utils.EmbeddedCassandraServerHelper - Cassandra daemon did not start after 120 ms. Consider increasing the timeout`
`Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.143 sec <<< FAILURE!`
this log means the timeout is too small (I change the timeout from 120s to 1.2s for generating this error.)

Because of the upper reason, I think even we use slf4j-simple, we can set the logger level from INFO to ERROR or WARN to reduce log content. 

But if the log of  ERROR level does not help us to find out reasons such as the `killed` case (I cannot reproduce this case on my Windows), I will agree with current modifications.

Or, if setting the log level is not an easy work, I will also agree with  current modifications.

:D

jixuan1989:
Then, as you said,  the timeout is the important piece. So I think if you think deatiled log is more helpful currently, we can approve this modificatins. Anyway, the short timeout causes too many failed tests, and it is troublesome...

risdenk:
Thanks @jixuan1989  going to merge for now to get the timeout in. The logs are already crazy with the code style stuff. We can readdress the log level later.

jixuan1989:
That's OK. 
Let me know for any other issues, Cassandra-related in particular, so I can contribute more..

PR #872, opened 2016-11-08
Update README and sample config file for jdbc

gitanuj:
I was going through jdbc bindings (for integration with cockroachdb) and found some inconsistencies:
1. The README mentions that a workload file can be supplied to com.yahoo.ycsb.db.JdbcDBCreateTable. However there's nothing in the class which makes use of it. I assume it's unnecessary.
2. Update the property key for jdbc driver from "jdbc.driver" to "db.driver" according to JdbcDBClient.DRIVER_CLASS.


risdenk:
@gitanuj Thanks for the PR! I won't have a chance to look at this till Friday but wanted you to know it was seen.

Issue #895, opened 2017-01-18
[core] Enforce checkstyle for core


risdenk:
Checkstyle has been implemented for a bunch of bindings. There are only a few that require checkstyle enforcement. I plan on opening a few PRs to address this. Relates to #267

risdenk:
There are a few areas left:
* measurements
* generators
* CommandLine.java
* Client.java

risdenk:
Once the above 4 PRs are merged, then we can enabled checkstyle enforcing for core!

PR #897, opened 2017-01-18
[core] Fix checkstyle for CoreWorkload

risdenk:
Relates to #895

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
+++ b/core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (c) 2010 Yahoo! Inc., Copyright (c) 2016 YCSB contributors. All rights reserved.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You
  * may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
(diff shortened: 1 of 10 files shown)


risdenk:
I checked this with: `mvn clean validate | grep CoreWorkload` and `mvn clean test`

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -105,14 +83,14 @@
    */
   public static final String FIELD_COUNT_PROPERTY_DEFAULT = "10";
 
-  int fieldcount;
+  private int fieldcount;
```
manolama: All the ``private``s should be ``protected``s for folks overloading the class (like me :)

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -506,14 +483,14 @@ public void init(Properties p) throws WorkloadException {
         INSERTION_RETRY_INTERVAL, INSERTION_RETRY_INTERVAL_DEFAULT));
   }
 
-  public String buildKeyName(long keynum) {
+  private String buildKeyName(long keynum) {
```
manolama: ``protected``

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -523,7 +500,7 @@ public String buildKeyName(long keynum) {
    * Builds a value for a randomly chosen field.
    */
   private HashMap<String, ByteIterator> buildSingleValue(String key) {
-    HashMap<String, ByteIterator> value = new HashMap<String, ByteIterator>();
+    HashMap<String, ByteIterator> value = new HashMap<>();
```
manolama: Are we going to require JDK7+ for YCSB? Or should the core still maintain JDK5 or 6 compatibility?

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
```diff
@@ -653,7 +635,7 @@ public boolean doTransaction(DB db, Object threadstate) {
    * Bucket 1 means incorrect data was returned.
    * Bucket 2 means null data was returned when some data was expected.
    */
-  protected void verifyRow(String key, HashMap<String, ByteIterator> cells) {
+  private void verifyRow(String key, HashMap<String, ByteIterator> cells) {
```
manolama: ``protected`` let overloaders override

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
risdenk: For these it would make more sense to add public getters/setters. It would break compatibility for now but would be better long term. I can add the getters/setters and push.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
risdenk: yup I'll make this protected.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
risdenk: yup I'll make this protected.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
risdenk: The pom.xml has the maven-compiler-plugin set to 1.7 for source and target. I don't think we are maintaining JDK 6 compatibility anymore. We started trying to add support for JDK 9 already.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
risdenk: The other thought here is to actually get the property value yourself instead of relying on CoreWorkload to populate the variable.

```java
String table = p.getProperty(CoreWorkload.TABLENAME_PROPERTY, CoreWorkload.TABLENAME_PROPERTY_DEFAULT);
```

manolama: But for proper inheritance you don't want to use getters/setters, you want to access the variables directly (plus avoid the function call [which would eventually be inlined, but...])

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
manolama: gotcha, sounds good.

core/src/main/java/com/yahoo/ycsb/workloads/CoreWorkload.java
risdenk: ok thats fair. Let me see what I can do about it.

risdenk: I found a way to allow protected variables in checkstyle.xml. It allowed me to set the variables to be protected.

risdenk:
Thanks @manolama!

PR #901, opened 2017-01-18
[core] Fix checkstyle for Core miscellaneous

risdenk:
Relates to #895

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/BasicDB.java
+++ b/core/src/main/java/com/yahoo/ycsb/BasicDB.java
@@ -1,18 +1,18 @@
-/**                                                                                                                                                                                
- * Copyright (c) 2010 Yahoo! Inc. All rights reserved.                                                                                                                             
- *                                                                                                                                                                                 
- * Licensed under the Apache License, Version 2.0 (the "License"); you                                                                                                             
- * may not use this file except in compliance with the License. You                                                                                                                
- * may obtain a copy of the License at                                                                                                                                             
- *                                                                                                                                                                                 
- * http://www.apache.org/licenses/LICENSE-2.0                                                                                                                                      
- *                                                                                                                                                                                 
- * Unless required by applicable law or agreed to in writing, software                                                                                                             
- * distributed under the License is distributed on an "AS IS" BASIS,                                                                                                               
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or                                                                                                                 
- * implied. See the License for the specific language governing                                                                                                                    
- * permissions and limitations under the License. See accompanying                                                                                                                 
- * LICENSE file.                                                                                                                                                                   
+/**
+ * Copyright (c) 2010-2016 Yahoo! Inc., 2017 YCSB contributors All rights reserved.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
  */
 
 package com.yahoo.ycsb;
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/yahoo/ycsb/BasicDB.java
```diff
@@ -21,280 +21,249 @@
...
+
+  private static final ThreadLocal<StringBuilder> TL_STRING_BUILDER = new ThreadLocal<StringBuilder>() {
     @Override
     protected StringBuilder initialValue() {
       return new StringBuilder();
     }
   };
 
-  static StringBuilder getStringBuilder() {
+  private static StringBuilder getStringBuilder() {
```
manolama: Super nit-picky but if anyone had their own code that was calling this from within the package or their own sublass then it would be a breaking change. Probably not a big deal at all.

manolama: Also would increment the copyright :)

core/src/main/java/com/yahoo/ycsb/DB.java
```diff
@@ -41,97 +41,94 @@
...
-{
-	/**
-	 * Properties for configuring this DB.
-	 */
-	Properties _p=new Properties();
+public abstract class DB {
+  /**
+   * Properties for configuring this DB.
+   */
+  private Properties properties = new Properties();
```
manolama: This one needs stay the same ``_p`` name so implementers don't see their code break. And ``protected`` instead of ``private``.

core/src/main/java/com/yahoo/ycsb/DB.java
```diff
@@ -41,97 +41,94 @@
...
+  /**
+   * Read a record from the database. Each field/value pair from the result will be stored in a HashMap.
+   *
+   * @param table The name of the table
+   * @param key The record key of the record to read.
+   * @param fields The list of fields to read, or null for all of them
+   * @param result A HashMap of field/value pairs for the result
+   * @return The result of the operation.
+   */
+  public abstract Status read(String table, String key, Set<String> fields, HashMap<String, ByteIterator> result);
```
manolama: It might be nice to change the ``HashMap`` to a ``Map`` since that shouldn't break the API and be consistent with ``Set``.

core/src/main/java/com/yahoo/ycsb/GoodBadUglyDB.java
```diff
@@ -1,162 +1,159 @@
...
  */
 public class GoodBadUglyDB extends DB {
-    public static final String SIMULATE_DELAY = "gbudb.delays";
-    public static final String SIMULATE_DELAY_DEFAULT = "200,1000,10000,50000,100000";
-    long delays[];
-    static ReadWriteLock DB_ACCESS = new ReentrantReadWriteLock();
+  public static final String SIMULATE_DELAY = "gbudb.delays";
+  public static final String SIMULATE_DELAY_DEFAULT = "200,1000,10000,50000,100000";
+  private long[] delays;
+  private static final ReadWriteLock DB_ACCESS = new ReentrantReadWriteLock();
```
manolama: Nother nit regarding the privatization of these two. Shouldn't be anyone using them but may be good to set them to ``protected`` just to be safe.

core/src/main/java/com/yahoo/ycsb/InputStreamByteIterator.java
```diff
@@ -1,55 +1,60 @@
...
-		off = 0;
-	}
-	
-	@Override
-	public boolean hasNext() {
-		return off < len;
-	}
+  private long len;
+  private InputStream ins;
+  private long off;
```
manolama: Nother nit, maybe ``protected`` just to be safe.

core/src/main/java/com/yahoo/ycsb/StringByteIterator.java
```diff
@@ -1,108 +1,116 @@
...
+import java.util.Map;
 
+/**
+ * A ByteIterator that iterates through a string.
+ */
 public class StringByteIterator extends ByteIterator {
-	String str;
-	int off;
+  private String str;
+  private int off;
```
manolama: nother nit, ``protected``

core/src/main/java/com/yahoo/ycsb/package-info.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2015 - 2016 YCSB contributors. All rights reserved.
```
manolama: ``2015 - 2017`` :)

core/src/main/java/com/yahoo/ycsb/DB.java
risdenk: We can't leave _p since that is not an allowed variable name by checkstyle. I don't see a way around this.

risdenk:
@manolama Thanks for the review! I'll update it tomorrow most likely.

core/src/main/java/com/yahoo/ycsb/DB.java
manolama: We can provide a custom checkstyle config with the project. It's annoying but better to maintain compatibility than worry about styling.

risdenk: Ok thats fair I'll see what I can do. I think there is a Suppress annotation way to handle just this line.

risdenk: So looking at this closer, there already exist get/set property methods so that should encapsulate the _p variable and allow us to change the name. If people were relying on the _p variable they should have been using the get/set methods.

core/src/main/java/com/yahoo/ycsb/DB.java
risdenk: I'm going to leave this to another issue. This requires changing all the implementations to match the method signature of the abstract method.

risdenk:
@manolama Updated the headers and fixed the getStringBuilder method to be protected.

A few of the changes I didn't make since they made checkstyle very unhappy.
* private -> protected variables - this complains that they should be private with getters/setters. I can add getters/setters but figured we could if/when necessary.
* properties -> _p - this complains about variable naming and there are already getters/setters so shouldn't be a problem changing the variable name

The comment about HashMap -> Map
* HashMap -> Map - this requires changing all the implementations of the abstract class to match the signature. This would be better to do in a separate issue.

My thought is that we should break compatibility and enforce checkstyle so it can't happen again. Right now we are not enforcing at all which makes it hard to even guarantee compatibility.

core/src/main/java/com/yahoo/ycsb/DB.java
manolama: Hmm, yeah I guess since none of the existing DB's use it directly and it's a tiny API change it'd be fine. We should just make a big note of it in the release notes.

risdenk:
Thanks @manolama!

PR #484, opened 2015-11-08
[hbase098] Checkstyle updates for HBase 0.98.X binding.

allanbank:

PR #902, opened 2017-01-18
[hbase098] Fix checkstyle for HBase 0.98.x binding

risdenk:
Relates to #267 and supersedes #484

### DIFF
--- a/hbase098/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
+++ b/hbase098/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
@@ -1,12 +1,12 @@
 /**
- * Copyright (c) 2010 Yahoo! Inc. All rights reserved.
- *
+ * Copyright (c) 2010-2016 Yahoo! Inc., 2017 YCSB contributors. All rights reserved.
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License"); you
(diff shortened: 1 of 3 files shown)


risdenk:
I'll rebase this in a few minutes and push the changes. This is the only remaining binding that doesn't enforce checkstyle!

risdenk:
@busbey if you have any spare cycles to review this PR that would be greatly appreciated. It is the last of the checkstyle changes.

risdenk:
I've been using https://github.com/brianfrankcooper/YCSB/pull/902/files?w=1 to check it to ignore the whitespace only changes.

hbase098/src/main/java/com/yahoo/ycsb/db/HBaseClient.java
```diff
@@ -22,576 +22,460 @@
...
 
-/* For customized vim control
- * set autoindent
- * set si
- * set shiftwidth=4
-*/
+    for (Thread t : allthreads) {
+      try {
+        t.join();
+      } catch (InterruptedException ignored) {
```
busbey: This is incorrect behavior. At a minimum the interrupted status for the current thread should be set.

It's a preexisting problem, not something from this fix, so I'd be fine with leaving it so long as there's an issue filed to fix it.

risdenk:
Thanks @busbey!

PR #908, opened 2017-01-19
[core] Fix ycsb.sh and ycsb.bat missing core dependencies

### LINKED ISSUE #877: ycsb.sh and ycsb.bat don't work from source checkout
The core dependencies aren't added to the classpath when bindings are used.

risdenk:
Fixes #877.

I don't know if this is the best way to solve this. Basically the core module's parent was root before. I made it binding parent so that it will pickup the necessary assembly plugin and create the lib directory. I'm not really familiar with Maven module parents and dependencies.

### DIFF
--- a/bin/ycsb.bat
+++ b/bin/ycsb.bat
@@ -144,7 +144,11 @@ GOTO classpathComplete
 
 :gotSource
 @REM Check for some basic libraries to see if the source has been built.
-IF EXIST "%YCSB_HOME%\%BINDING_DIR%\target\*.jar" GOTO gotJars
+IF EXIST "%YCSB_HOME%\core\target\dependency\*.jar" (
+  IF EXIST "%YCSB_HOME%\%BINDING_DIR%\target\*.jar" (
+    GOTO gotJars
+  )
+)
 
 @REM Call mvn to build source checkout.
 IF "%BINDING_NAME%" == "basic" GOTO buildCore
@@ -155,7 +159,7 @@ SET MVN_PROJECT=core
 :gotMvnProject
 
 ECHO [WARN] YCSB libraries not found.  Attempting to build...
-CALL mvn -pl com.yahoo.ycsb:%MVN_PROJECT% -am package -DskipTests
+CALL mvn -Psource-run -pl com.yahoo.ycsb:%MVN_PROJECT% -am package -DskipTests
 IF %ERRORLEVEL% NEQ 0 (
   ECHO [ERROR] Error trying to build project. Exiting.
   GOTO exit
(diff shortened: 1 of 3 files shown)


risdenk:
@busbey - Sorry to keep bugging you with reviews. Not sure who else would know how this should work. If there is someone else to ping let me know.

busbey:
sorry I lost track of this one.

> @busbey - Sorry to keep bugging you with reviews. Not sure who else would know how this should work. If there is someone else to ping let me know.

Might be time to bring in some additional project maintainers.

I don't think adding binding-parent to be the core's parent is the right way to do this. I'm trying to grok the underlying problem reading a few of the referenced problems.

Is it just when using a source checkout and a binding other than the built-in dummy one?

noahdesu:
I'm seeing this issue with a source checkout.

noahdesu:
I added something like this to ycsb.sh where the class path is constructed for the source case and it seems to work. not sure what the best thing is.

```
+  # Core dependencies from Maven
+  CLASSPATH_FILE=$(mktemp)
+  mvn -pl com.yahoo.ycsb:core dependency:build-classpath -Dmdep.outputFile=$CLASSPATH_FILE
+  CLASSPATH_MVN=$(cat $CLASSPATH_FILE)
+  echo XXX $CLASSPATH_MVN
+  CLASSPATH=$CLASSPATH:$CLASSPATH_MVN
 fi
```

risdenk:
@busbey so I had a second look at this. It is simpler to copy the dependencies for `core` and slightly modify the `ycsb.sh` and `yscb.bat` to pick up those dependencies. The parsing of `mvn dependency:build-classpath` isn't easy to do with Windows.

risdenk:
@busbey yea that sounds reasonable. I'll update the PR.

manolama:
@risdenk Still want this in v13?

risdenk:
@manolama up to you. I'm done making changes now unless there are required changes from reviews.

@busbey Can you take a look at the latest changes? I don't have a Windows machine to test on right now. It works on my Mac. I switched from `for` to `ls` since it was easier to read and shouldn't be too many files in the directory that are *.jar.

risdenk:
@busbey that would be great. The tests I ran:

```
mvn clean
bin/ycsb.sh load basic -P workloads/workloada
bin/ycsb.sh load rest -P workloads/workloada
```

This will make sure that basic works and builds correctly. Then check that another binding runs without getting a classpath error.

busbey:
confirmed this didn't get any classpath trouble on Windows 10 Home

```
git clean -xdf
.\bin\ycsb.bat load basic -P .\workloads\workloada
.\bin\ycsb.bat load rest -P .\workloads\workloada
```

PR #910, opened 2017-01-19
[jdbc] Use `TEXT` for fields instead of VARCHAR

risdenk:
Supersedes PR #797. Updated JavaDocs


manolama:
TODO - would be nice to have this as a flag so the executor can choose which type to use.

PR #918, opened 2017-02-02
[core] Fix checkstyle for Client

risdenk:
Relates to #895

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Client.java
+++ b/core/src/main/java/com/yahoo/ycsb/Client.java
@@ -1,12 +1,12 @@
 /**
- * Copyright (c) 2010 - 2016 Yahoo! Inc. All rights reserved.
- *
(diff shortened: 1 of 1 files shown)


risdenk:
Tested with: `mvn -pl core -am validate | grep "ycsb/Client"`

risdenk:
Required by #920

core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -1,12 +1,12 @@
 /**
  * Copyright (c) 2010 - 2016 Yahoo! Inc. All rights reserved.
```
busbey: looks like copyright date needs to change.

risdenk: Ah yup I'll fix that.

Issue #929, opened 2017-02-21
[cassandra-cql] PreparedStatement JUnit Test; key-value iteration order


cricket007:
I'm in the process of merging a piece of #98 (related to #458), and I think I am close. Meaning the Cassandra `PreparedStatements` are in the code, but I am trying to get the `insert` & `update` JUnit tests to pass now. 

The `PreparedStatement` looks like so `INSERT INTO usertable (y_id,field0,field1,field2,field3,field4,field5,field6,field7,field8,field9) VALUES (?,?,?,?,?,?,?,?,?,?,?);`

I was referring to this code on the 98 pull, and I see this.  
Basically, create the Object array from the values Map and simply call `bind()` with that array. 

```java
@Override
public int insert(String table, String key, Map<String, ByteIterator> values)
{
    try
    {
        Object[] vals = new Object[values.size() + 1];
        vals[0] = key;
        int i = 1;
        for (Map.Entry<String, ByteIterator> entry : values.entrySet())
        {
            vals[i++] = ByteBuffer.wrap(entry.getValue().toArray());
        }

        BoundStatement bs = (values.size() == 1 ? updateStatements.get(values.keySet().iterator().next()) : insertStatement).bind(vals);
```

Where the unit tests are failing - [CassandraCQLClientTest.java#L178][1] 

```java
assertThat(row.getString("field0"), is("value1"));

java.lang.AssertionError: 
Expected: is "value1"
     but: was "value2"
```

Which makes sense because `HashMap` ordering is not guaranteed. 

I was looking at the JDBC code and I see this [`private OrderedFieldInfo getFieldInfo`](https://github.com/brianfrankcooper/YCSB/blob/master/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java#L486) method. However, that feels wrong, so I am looking for other suggestions. 

Follow-up question - Is iterating over that map in order something that needs to be done often? If so, why isn't a `LinkedHashMap` used?

  [1]: https://github.com/brianfrankcooper/YCSB/blob/master/cassandra/src/test/java/com/yahoo/ycsb/db/CassandraCQLClientTest.java#L178

risdenk:
Can you share how you are creating the prepared statement? How do you know the order of `(y_id,field0,field1,field2,field3,field4,field5,field6,field7,field8,field9)`? If you know that order, can't you store that order and then loop through that list and grab  the values from the values map by field name? This wouldn't rely on the order of the map then.

[...]

Issue #458, opened 2015-10-30
CassandraCQLClient v2 doesn't use prepared statements


dprotaso:
Loading data is really slow

cmccoy:
v1 doesn't either. Would you be willing to contribute a patch?

dprotaso:
I don't have much free time - plus I see the work being done here: https://github.com/brianfrankcooper/YCSB/pull/98

busbey:
just to set expectations, I've been trying to get #98 into a merge-able state for several months. recent improvements in our code style are probably going to make it even harder.

busbey:
closed by #1051

PR #930, opened 2017-02-21
[cassandra-cql] Cassandra PreparedStatements... again

cricket007:
Requesting to pull changes from #929 which include some commits from #98 and addresses #458. 

Reviews are welcome. 

**Client**

- [X] `insert`
- [X] `update` (Correct me if I am wrong, `insert` is basically an `update` with just one key-value?)
- [X] `delete`
- [X] `scan` (one)
- [x] `read` (one)
- [X] <strike>`scan` (many)</strike>
- [X] <strike>`read` (many)</strike>

**Unit Tests**  
Assuming these should also use PreparedStatements

- [x] `insertRow`
- [x] `testUpdate` 
- [x] `testRead`

### DIFF
--- a/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
+++ b/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
@@ -389,11 +389,7 @@ public class CassandraCQLClient extends DB {
 
       for (ColumnDefinitions.Definition def : cd) {
         ByteBuffer val = row.getBytesUnsafe(def.getName());
-        if (val != null) {
-          result.put(def.getName(), new ByteArrayByteIterator(val.array()));
-        } else {
-          result.put(def.getName(), null);
-        }
+        result.put(def.getName(), val == null ? null : new ByteArrayByteIterator(val.array()));
       }
 
       return Status.OK;
@@ -476,11 +472,7 @@ public class CassandraCQLClient extends DB {
 
         for (ColumnDefinitions.Definition def : cd) {
           ByteBuffer val = row.getBytesUnsafe(def.getName());
-          if (val != null) {
-            tuple.put(def.getName(), new ByteArrayByteIterator(val.array()));
-          } else {
-            tuple.put(def.getName(), null);
-          }
+          tuple.put(def.getName(), val == null ? null : new ByteArrayByteIterator(val.array()));
         }
 
         result.add(tuple);
--- a/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
+++ b/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
@@ -566,20 +566,15 @@ public class CassandraCQLClient extends DB {
   public Status delete(String table, String key) {
 
     try {
-      Statement stmt;
-
-      stmt = QueryBuilder.delete().from(table)
-          .where(QueryBuilder.eq(YCSB_KEY, key));
-      stmt.setConsistencyLevel(writeConsistencyLevel);
 
       if (debug) {
-        System.out.println(stmt.toString());
+        System.out.println(deleteStatement.getQueryString());
       }
       if (trace) {
-        stmt.enableTracing();
+        deleteStatement.enableTracing();
       }
       
-      session.execute(stmt);
+      session.execute(deleteStatement.bind(key));
 
       return Status.OK;
     } catch (Exception e) {
--- a/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
+++ b/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
@@ -249,26 +249,17 @@ public class CassandraCQLClient extends DB {
             .where(QueryBuilder.eq(YCSB_KEY, QueryBuilder.bindMarker()))
             .limit(1);
 
-        if (debug) {
-          System.out.printf("Preparing '%s'\n", selectOne.getQueryString());
-        }
         if (!singleSelectStatements.containsKey(field)) {
-          PreparedStatement preparedSelect = session.prepare(selectOne);
-          preparedSelect.setConsistencyLevel(readConsistencyLevel);
-          singleSelectStatements.put(field, preparedSelect);
+          singleSelectStatements.put(field, prepare(selectOne.getQueryString(), true));
         }
 
         // Scan - single
         String initialScanStmt = QueryBuilder.select(field).from(table).toString();
         String scanOneStmt = getScanQueryString().replaceFirst("_",
             initialScanStmt.substring(0, initialScanStmt.length()-1));
-        if (debug) {
-          System.out.printf("Preparing '%s'\n", scanOneStmt);
-        }
+
         if (!singleScanStatements.containsKey(field)) {
-          PreparedStatement preparedScan = session.prepare(scanOneStmt);
-          preparedScan.setConsistencyLevel(readConsistencyLevel);
-          singleScanStatements.put(field, preparedScan);
+          singleScanStatements.put(field, prepare(scanOneStmt, true));
         }
 
         // Select and Scan many statements
@@ -279,39 +270,32 @@ public class CassandraCQLClient extends DB {
     }
 
     // Prepare insert and update all
-    if (debug) {
-      System.out.printf("Preparing '%s'\n", is.getQueryString());
-    }
-    insertStatement = session.prepare(is);
-    insertStatement.setConsistencyLevel(writeConsistencyLevel);
+    insertStatement = prepare(is.getQueryString(), false);
 
     // Select All
     String ss = QueryBuilder.select().all().from(table)
             .where(QueryBuilder.eq(YCSB_KEY, QueryBuilder.bindMarker())).getQueryString();
-    if (debug) {
-      System.out.printf("Preparing '%s'\n", ss);
-    }
-    selectStatement = session.prepare(ss);
-    selectStatement.setConsistencyLevel(readConsistencyLevel);
+    selectStatement = prepare(ss, true);
 
     // Scan All
     String initialStmt = QueryBuilder.select().all().from(table).toString();
     String scanStmt = getScanQueryString().replaceFirst("_",
         initialStmt.substring(0, initialStmt.length()-1));
-    if (debug) {
-      System.out.printf("Preparing '%s'\n", scanStmt);
-    }
-    scanStatement = session.prepare(scanStmt);
-    scanStatement.setConsistencyLevel(readConsistencyLevel);
+    scanStatement = prepare(scanStmt, true);
 
     // Delete on key statement
     String ds = QueryBuilder.delete().from(table)
         .where(QueryBuilder.eq(YCSB_KEY, QueryBuilder.bindMarker())).toString();
+    deleteStatement = prepare(ds, false);
+  }
+
+  private PreparedStatement prepare(String statement, boolean isRead) {
     if (debug) {
-      System.out.printf("Preparing '%s'\n", ds);
+      System.out.printf("Preparing '%s'\n", statement);
     }
-    deleteStatement = session.prepare(ds);
-    deleteStatement.setConsistencyLevel(writeConsistencyLevel);
+    PreparedStatement prepared = session.prepare(statement);
+    prepared.setConsistencyLevel(isRead ? readConsistencyLevel : writeConsistencyLevel);
+    return prepared;
   }
 
   private String getScanQueryString() {
--- a/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
+++ b/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
@@ -18,7 +18,6 @@
 package com.yahoo.ycsb.db;
 
 import com.datastax.driver.core.*;
-import com.datastax.driver.core.querybuilder.Delete;
 import com.datastax.driver.core.querybuilder.Insert;
 import com.datastax.driver.core.querybuilder.QueryBuilder;
 import com.datastax.driver.core.querybuilder.Select;
@@ -73,7 +72,7 @@ public class CassandraCQLClient extends DB {
       "cassandra.readtimeoutmillis";
 
   public static final String TRACING_PROPERTY = "cassandra.tracing";
-  public static final String TRACING_PROPERTY_DEFAULT = "false";
+  public static final String TRACING_PROPERTY_DEFAULT = Boolean.FALSE.toString();
 
   private static List<String> fieldList;
   private static boolean readallfields;
@@ -89,9 +88,9 @@ public class CassandraCQLClient extends DB {
     one to select all columns, and one for selecting each individual column
   */
   private static PreparedStatement selectStatement = null;
-  private static Map<String, PreparedStatement> selectStatements = null;
+  private static Map<String, PreparedStatement> singleSelectStatements = null;
   private static PreparedStatement scanStatement = null;
-  private static Map<String, PreparedStatement> scanStatements = null;
+  private static Map<String, PreparedStatement> singleScanStatements = null;
 
   /**
    * Count the number of times initialized to teardown on the last
@@ -100,9 +99,11 @@ public class CassandraCQLClient extends DB {
   private static final AtomicInteger INIT_COUNT = new AtomicInteger(0);
 
   private static boolean debug = false;
-
   private static boolean trace = false;
-  
+
+  private static TreeMap<Integer, Map<Integer, PreparedStatement>> selectManyStatements;
+  private static TreeMap<Integer, Map<Integer, PreparedStatement>> scanManyStatements;
+
   /**
    * Initialize any state for this DB. Called once per DB instance; there is one
    * DB instance per client thread.
@@ -127,7 +128,7 @@ public class CassandraCQLClient extends DB {
         final Properties p = getProperties();
 
         debug =
-            Boolean.parseBoolean(p.getProperty("debug", "false"));
+            Boolean.parseBoolean(p.getProperty("debug", Boolean.FALSE.toString()));
         trace = Boolean.valueOf(p.getProperty(TRACING_PROPERTY, TRACING_PROPERTY_DEFAULT));
         
         String host = p.getProperty(HOSTS_PROPERTY);
@@ -179,18 +180,18 @@ public class CassandraCQLClient extends DB {
               Integer.valueOf(coreConnections));
         }
 
-        String connectTimoutMillis = p.getProperty(
+        String connectTimeoutMillis = p.getProperty(
             CONNECT_TIMEOUT_MILLIS_PROPERTY);
-        if (connectTimoutMillis != null) {
+        if (connectTimeoutMillis != null) {
           cluster.getConfiguration().getSocketOptions()
-              .setConnectTimeoutMillis(Integer.valueOf(connectTimoutMillis));
+              .setConnectTimeoutMillis(Integer.valueOf(connectTimeoutMillis));
         }
 
-        String readTimoutMillis = p.getProperty(
+        String readTimeoutMillis = p.getProperty(
             READ_TIMEOUT_MILLIS_PROPERTY);
-        if (readTimoutMillis != null) {
+        if (readTimeoutMillis != null) {
           cluster.getConfiguration().getSocketOptions()
-              .setReadTimeoutMillis(Integer.valueOf(readTimoutMillis));
+              .setReadTimeoutMillis(Integer.valueOf(readTimeoutMillis));
         }
 
         Metadata metadata = cluster.getMetadata();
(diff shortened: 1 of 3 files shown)


cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -85,6 +72,24 @@
   public static final String READ_TIMEOUT_MILLIS_PROPERTY =
       "cassandra.readtimeoutmillis";
 
+  private static List<String> fieldList;
```
cricket007: Looking back over, they do not need to be static at all.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -195,12 +205,84 @@ public void init() throws DBException {
...
+      selectStatement.setConsistencyLevel(readConsistencyLevel);
+
+      // Scan
+      String initialStmt = QueryBuilder.select().all().from(table).toString();
+      String scanStmt = getScanQueryString().replaceFirst("_", initialStmt.substring(0, initialStmt.length()-1));
+      scanStatement = session.prepare(scanStmt);
+      scanStatement.setConsistencyLevel(readConsistencyLevel);
+
+    } else {
+      selectStatements = new ConcurrentHashMap<String, PreparedStatement>(fieldCount);
```
cricket007: The `readallfields` boolean will build a `SELECT *` statement when true, otherwise the for-loop below would be used, and a `fieldCount == 1` would select just one field.  
Are you suggesting it is instead `if (readallfields || fieldCount == 1)`?

risdenk: Well just curious if selecting from two fields is possible. I haven't looked that closely at it.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
cricket007: Actually, maybe they should be. All threads with an instance of this class should share them. These variables are declared in the `synchronized` block.

cricket007:
Not too sure about the support for selecting and scanning more than one field. 

I started writing this before I realized that 

1. The fields to be selected aren't always sequentially ordered. 
2. I would need to generate a large number of prepared statements. Like `nCr(10,1) + nCr(10,2) + nCr(10,3) + ... + nCr(10,10)`, or 1023 statements, [according to WolframAlpha](http://www.wolframalpha.com/input/?i=sum+ncr(10,+x)+for+x%3D1..10)

```java
for (int i = 0; i < fieldCount; i++) {
  ...
  Select.Selection selectManyBuilder = QueryBuilder.select();;
  for (int j = i; j < fieldCount; j++) {
      selectManyBuilder.column(FIELD_PREFIX + j);
  }
  Select selectMany = selectManyBuilder.from(table)
    .where(QueryBuilder.eq(YCSB_KEY, QueryBuilder.bindMarker())).limit(1);
  PreparedStatement preparedSelectManyStatement = session.prepare(selectMany);
}
```

Therefore, I think keeping the QueryBuilder within the `read` and `scan` functions makes sense. Unless someone has a better idea.

risdenk:
Just thinking out loud, can a query be repeated? How much benefit would it be to create the prepared statement on demand (first time) and then try to reuse the prepared statement if the same query comes in again?

That may give the best of both worlds in terms of not generating them all up front, but at least being able to generate them just in time? I don't know the performance of prepared vs nonprepared statements though and how much effort it takes to generate one.

cricket007:
I do think generating thousands of queries ahead of time would be pointless. Similar thoughts for creating those objects without any known future use; maybe lazily preparing them would be the way to go.  

Regarding queries repeating - isn't that the point of YCSB? To hit the database with the same query repeatedly?

cricket007:
@risdenk  Ready for review.

cassandra/src/test/java/com/yahoo/ycsb/db/CassandraCQLClientTest.java
```diff
@@ -65,25 +57,56 @@
...
     Properties p = new Properties();
-    p.setProperty("hosts", HOST);
-    p.setProperty("port", Integer.toString(PORT));
-    p.setProperty("table", TABLE);
-
+    p.setProperty(CassandraCQLClient.HOSTS_PROPERTY, HOST);
+    p.setProperty(CassandraCQLClient.PORT_PROPERTY, Integer.toString(PORT));
+    p.setProperty(CoreWorkload.TABLENAME_PROPERTY, TABLE);
+    p.setProperty(CoreWorkload.READ_ALL_FIELDS_PROPERTY, Boolean.FALSE.toString());
+    p.setProperty("debug", Boolean.TRUE.toString());
```
cricket007: Might want to set `debug = false`?

risdenk: This is just in the test so less worried about it.

cassandra/src/test/java/com/yahoo/ycsb/db/CassandraCQLClientTest.java
```diff
@@ -17,27 +17,16 @@
...
-
-import com.datastax.driver.core.ResultSet;
-import com.datastax.driver.core.Row;
-import com.datastax.driver.core.Session;
-import com.datastax.driver.core.Statement;
+import com.datastax.driver.core.*;
 import com.datastax.driver.core.querybuilder.Insert;
 import com.datastax.driver.core.querybuilder.QueryBuilder;
 import com.datastax.driver.core.querybuilder.Select;
+import com.google.common.collect.Sets;
```
risdenk: Do we need to have a dependency on google Sets here? (I know it was there before but just curious)

cricket007: It appears to be used in `testReadSingleColumn`

    final Set<String> fields = Sets.newHashSet("field1");

risdenk:
@cricket007 any benchmark numbers to share? Just want to confirm this isn't slower than the original implementation?

Another thought - would it make sense to make it an option to use prepared statements or not? A property perhaps?

cricket007:
@risdenk How would you recommend benchmarking it? 

I can setup a few node Cassandra cluster in vagrant, but is one YCSB node enough?

risdenk:
@cricket007 It would be good to try to get at least some performance numbers. A single node is better than nothing.

cricket007:
I was going to work on this over the weekend, but seems my previously working Vagrant / Ansible setup is busted. I'll try to see if I can get Docker working.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -425,30 +526,34 @@ public Status insert(String table, String key,
...
+        String field = fieldList.get(i);
+        if (values.containsKey(field)) {
+          vals[i] = values.get(field).toString();
+        }
       }
 
-      insertStmt.setConsistencyLevel(writeConsistencyLevel);
+      // If there is one value, this is actually an update
+      BoundStatement bs = (numValues == 1 ?
+               updateStatements.get(values.keySet().iterator().next()) :
```
jixuan1989: it seems that `updateStatements` has not been initialized, i.e., a null exception may occur...

cricket007: Hmm. Good catch. Not sure why I didn't I didn't see that in IntelliJ. 

I feel like I did have it at one point before I attempted to re-order the "sections" of statement initializations. 
https://github.com/jbellis/YCSB/blob/217020bd07e556ba2870e2487435d6297e18fb3c/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java#L174-L185

cricket007: I was considering adding more unit tests, which probably would have caught that.  Suggestions welcome.

jixuan1989: Sounds good :)....

By the way, the line `is.value(YCSB_KEY, QueryBuilder.bindMarker());` seems need to  be out of the for-loop.

```java
updateStatements = new ConcurrentHashMap<String,PreparedStatement>(fieldCount);
        for (int i = 0; i < fieldCount; i++)
        {
            is = QueryBuilder.insertInto(table);
            is.value(YCSB_KEY, QueryBuilder.bindMarker());
            is.value(fieldPrefix + i, QueryBuilder.bindMarker());

            PreparedStatement ps = session.prepare(is);
            ps.setConsistencyLevel(writeConsistencyLevel);
            updateStatements.put(fieldPrefix + i, ps);
        }
```

cricket007: I don't know about that since this loop seems to be generating a new `insertInto`, then binding the key and one individual column. 
 
That link is the other pull request mentioned in the description up top upon which my code is based.

jixuan1989: Oops, you are right~

By the way, because you say `// YCSB always inserts a full row, but updates can be either full-row or single-column`, the above codes serve for `single-column` update, so you can add one more updateStatement for add all the fields in the statement.

jixuan1989: Oh I see! you use `insertStatement` rather than `updateStatement` for a full-row update/insertion.

jixuan1989: Ok, Now I have no questions any more~

cricket007: @jixuan1989 Should be good now

cricket007:
@risdenk 

### Preliminary results

Loaded 50000 records    
3 Servers  
1 Client  
35 threads  
Workload E for 20 minutes

Before
```
[OVERALL], Throughput(ops/sec), 470.2543284237526
[INSERT], 95thPercentileLatency(us), 129727.0
[SCAN], 95thPercentileLatency(us), 197375.0
```

After
```
[OVERALL], Throughput(ops/sec), 6894.908356323305
[INSERT], 95thPercentileLatency(us), 6595.0
[SCAN], 95thPercentileLatency(us), 0.0
```

CPU usage ~20% less

risdenk:
@cricket007 that looks good. Only one question, why is [SCAN] after "0.0" for latency in both runs? `[SCAN], 95thPercentileLatency(us), 0.0` That seems really weird to me. Did we lose tracking latency on scans?

cricket007:
Yeah, that surprised me too. Something is cached, so reads are really quick? Probably not since it's zero microseconds, though...  

I only grep'd those values, but I can give the rest of the run output.

cricket007:
@risdenk 

Found the problem. Didn't check the error output

```
com.datastax.driver.core.exceptions.CodecNotFoundException: Codec not found for requested operation: [bigint <-> java.lang.String]
        at com.datastax.driver.core.CodecRegistry.notFound(CodecRegistry.java:679)
        at com.datastax.driver.core.CodecRegistry.createCodec(CodecRegistry.java:540)
        at com.datastax.driver.core.CodecRegistry.findCodec(CodecRegistry.java:520)
        at com.datastax.driver.core.CodecRegistry.codecFor(CodecRegistry.java:470)
        at com.datastax.driver.core.AbstractGettableByIndexData.codecFor(AbstractGettableByIndexData.java:77)
        at com.datastax.driver.core.BoundStatement.bind(BoundStatement.java:196)
        at com.datastax.driver.core.DefaultPreparedStatement.bind(DefaultPreparedStatement.java:126)
        at com.yahoo.ycsb.db.CassandraCQLClient.scan(CassandraCQLClient.java:468)
        at com.yahoo.ycsb.DBWrapper.scan(DBWrapper.java:176)
        at com.yahoo.ycsb.workloads.CoreWorkload.doTransactionScan(CoreWorkload.java:782)
        at com.yahoo.ycsb.workloads.CoreWorkload.doTransaction(CoreWorkload.java:640)
        at com.yahoo.ycsb.ClientThread.run(Client.java:458)
        at java.lang.Thread.run(Thread.java:745)
```

Seems the `bind` parameters should be the correct types, or something about a CodecRegistry 

```
String startkey, int recordcount 
...
final BoundStatement bs = ps.bind(startkey, recordcount);
```

manolama:
@cricket007 Did you have to patch anything for that scan latency issue or is it good to go? Thanks!

cricket007:
Ah, hey @manolama. 

Unfortunately, no. I actually just tried reaching out to the DataStax slack channel a few days ago and I got a "that's strange", while dismissing it as a driver issue. I was going to try upgrading that and see what happens... 

I've had very limited experience with Cassandra, and mostly just going off of the few PreparedStatements pages on the DataStax documentation, so I really don't know what the error is trying to tell me as I would assume a CQL `bigint` type would have a conversion to a String, and I'm also not sure where that `bigint` value is at. 

I'm only trying to bind

    SELECT FROM usertable WHERE token(y_id) >= ? LIMIT ?;

So, the limit must be a number, and I don't think I changed the syntax of the query before these changes, so the while the `startkey` is a `java.lang.String`, I don't know why it would be trying to convert into a CQL `bigint`...  https://github.com/brianfrankcooper/YCSB/blob/master/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java#L333-L347

cricket007:
omg... months later and I finally see the difference... 😲 

The startkey is being tokened and quoted. 

      scanStmt.append("token('");
      scanStmt.append(startkey);
      scanStmt.append("')");

I'll start running this again as soon as I can.

manolama:
Heh, sounds good, thanks!

cricket007:
Oops. Didn't mean to push that commit. 

Well, I believe I  fixed the binding problem in the scan, but now I'm getting NullPointerException in about every usage of getting the `fieldList` or the Hashmaps containing the PreparedStatements. 

The unit tests pass fine, so it's probably a multithreading issue

cricket007:
Definitely a threading issue. Moving around the property initialization blocks out of the `synchronized` blocks has now resolved those issues. 

Now getting errors when trying to log out the `BoundStatement`s, but might just leave out the debug feature for those.

busbey:
anything we can help with @cricket007?

cricket007:
Hey @busbey, I'm not well-versed in Cassandra, so any assistance would be great. 
I can't remember what the current errors are, I can collect those later on my other computer. 

But if you are able to pull my changes and try things out, I would gladly take suggestions.

cricket007:
Thanks @busbey , I do have a question actually, which might help me. 

When I run JUnit in IntelliJ, is that simulating one YCSB instance?  If so, how can I configure IntelliJ so that there would be multiple threads during testing? That way, I can set breakpoints and watch the necessary variables.  

I ask because all my unit tests pass, so I believe the root issue is when there is more than one thread.

jixuan1989:
@cricket007   `NullPointException` is not caused by multi-threads. 
It is because you wrote the following codes for optimizing Cassandra's performance:
```java
        // Select and Scan few statements
        if (!readallfields) {
....
}
```
Only `readallfileds==false`, you initialize `updatestatements`, that is fine.  
However, `CoreWorkload.READ_ALL_FIELDS_PROPERTY_DEFAULT=true` :( 
So, you have to add a parameter in your workload file: `readallfields=false`

I think we need to write it in Readme.md or provide a workload template for Cassandra.

Besides, I have tested the scan operations:
```
[SCAN], Operations, 408
[SCAN], AverageLatency(us), 3389.5906862745096
[SCAN], MinLatency(us), 391
[SCAN], MaxLatency(us), 18239
[SCAN], 95thPercentileLatency(us), 7087
[SCAN], 99thPercentileLatency(us), 9343
[SCAN], Return=OK, 408
```

jixuan1989:
Besides, I think the insert() function should be:
```java
@Override
  public Status insert(String table, String key,
      Map<String, ByteIterator> values) {

    try {
      final int numValues = values.size();
      int i = 0;
      Object[] vals = new Object[numValues + 1];

      // Add key
      vals[i++] = key;
      // Add fieldList
      for (int j=1; j < fieldList.size(); j++) {
        String field = fieldList.get(j);
        
        if (values.containsKey(field)) {
          vals[i++] = values.get(field).toString();
        }
      }

      // If there is one value, this is actually an update
      BoundStatement bs = (numValues == 1 ?
               updateStatements.get(values.keySet().iterator().next()) :
               insertStatement).bind(vals);

      if (trace) {
        bs.enableTracing();
      }
      if (debug) {
        System.out.println(">> " + getBoundQuery(bs));
      }
      
      session.execute(bs);

      return Status.OK;
    } catch (Exception e) {
      e.printStackTrace();
    }

    return Status.ERROR;
  }
```
the original codes are:
```java
      final int numValues = values.size();
      int i = 0;
      Object[] vals = new Object[numValues + 1];
      // Add key
      vals[i++] = key;
      // Add fieldList
      for (; i < fieldList.size(); i++) {
        String field = fieldList.get(i);     
        if (values.containsKey(field)) {
          vals[i] = values.get(field).toString();
        }
      }
```
it is dangerous that ArrayOutOfBound for `vals[]`

jixuan1989:
One more thing, it is better to add a dependency in the `core` project, because `core` uses `snakeyaml` version 1.6 (involved by `testng`) while Cassandra uses version 1.11 which is not compatible with 1.6.  
I can find the conflict in Eclipse IDE (but no conlict in a terminal with `mvn` command).

@cricket007  If you need, I can fork your codes and then submit a new PR. 
Or, I can fork your codes and then submit a PR to your branch first. After you merge my modifications, this PR can be updated automatically.

cricket007:
Thanks for reviewing again @jixuan1989 - If you're able to better pinpoint the issue, feel free to fork mine, or if you are free sometime, we can schedule a call to go over what you think needs to be changed to get it fully working. 

>  I have tested the scan operations:

Yes, I believe I was able to make Scan and Insert (Workload E) work with my latest commits. I started getting busy in April and didn't have time to resolve or throughly test Read and Update (Workload A or B)

I claimed it was a threading issue because moving the `synchronized` blocks around has resolved one issue I was experiencing.

jixuan1989:
@cricket007  I have reviewed the codes. 
I think your codes are correct for multi-threads scenario. 
Cassandra Java Driver supports using one `cluster` instance and `session` for multi-threads. However, you have to initialize the `cluster` synchronically. 
However, the synchronized keyword in `buildStatements`  function is useless. 

(I am not sure that whether YCSB consider the init() time as a part of time cost for CRUD. If so, it is unfair for current implementation, and we have to try to find another implementation. Otherwise, the current codes are fine.)

I have correct other codes and merge the codes with latest master branch.

I submit a PR to your repository. You can merge it so that they can be pushed here automatically.

cricket007:
@jixuan1989 Can you explain when you say 

> Only `readallfileds==false`, you initialize `updateStatements`, that is fine.

That isn't true - see this section isn't inside of any if statements

```java
    // Insert and Update all statement
    final Insert is = QueryBuilder.insertInto(table);
    is.value(YCSB_KEY, QueryBuilder.bindMarker());

    updateStatements = new ConcurrentHashMap<>(fieldCount);
```

That is only used when one value is passed to `insert()`, though. 

```java
  @Override
  public Status insert(String table, String key,
      Map<String, ByteIterator> values) {

    try {
      final int numValues = values.size();
      ...
      // If there is one value, this is actually an update
      BoundStatement bs = (numValues == 1 ?
               updateStatements.get(values.keySet().iterator().next()) :
               insertStatement).bind(vals);
```

It can be moved outside that if statement, if you think that's the solution; I'm not completely sure what field(s) would be passed to the UPDATE query.

jixuan1989:
@cricket007  Current YCSB workload inserts ten fields for each insertion command, and just update one field for updating commands.

But.. see the Excetion Stack:

> java.lang.NullPointerException
	at com.yahoo.ycsb.db.CassandraCQLClient.insert(CassandraCQLClient.java:585)
	at com.yahoo.ycsb.db.CassandraCQLClient.update(CassandraCQLClient.java:545)
	at com.yahoo.ycsb.DBWrapper.update(DBWrapper.java:198)
	at com.yahoo.ycsb.workloads.CoreWorkload.doTransactionUpdate(CoreWorkload.java:775)
	at com.yahoo.ycsb.workloads.CoreWorkload.doTransaction(CoreWorkload.java:608)
	at com.yahoo.ycsb.ClientThread.run(Client.java:454)
	at java.lang.Thread.run(Thread.java:748)

You call `insert()` when YCSB calls `update()`. That's why the exception appears. 

I can try to fix it.

jixuan1989:
@cricket007  I have fixed the update problem as following:

```java
init(){
...
    updateCommands = Double.parseDouble(p.getProperty(CoreWorkload.UPDATE_PROPORTION_PROPERTY,
        CoreWorkload.UPDATE_PROPORTION_PROPERTY_DEFAULT))==0 ? false : true;
}

buildStatements(){
...
if(updateCommands) {
        // Update - single
        Insert updateOneStmt = QueryBuilder.insertInto(table).values(new String[] {YCSB_KEY, field },
            new Object[] {QueryBuilder.bindMarker(), QueryBuilder.bindMarker() });
        if (!updateStatements.containsKey(field)) {
          updateStatements.put(field, prepare(updateOneStmt.getQueryString(), false));
        }
      }
}
```
but ... I notice that you close the PR of my branch. If so, you can digest the above part of codes manually.

cricket007:
@jixuan1989 I closed the PR, but I did merge the code. I added some changes on top of it in a different branch. 

Anyways, I think the latest commit I made was able to run without error with either value of `readallfields`, however now I see that there are `READ-FAILED` events, which I do not think were there before. 

By the way, `==0 ? false : true;` can simply be written as `!=0;`

jixuan1989:
@cricket007 

> however now I see that there are READ-FAILED events, which I do not think were there before.

Can you give me a screenshot or detailed running log? I did not find the event before.


@robertpang  Your codes are more concise. It uses less API of Datastax Java Driver so that it is more easy to be understood.   Do you notice the same problem (READ-FAILED events)? 

No matter which PR is accepted finally, I hope we can accelerate the progress... It is really important that using this better way to benchmark Cassandra.

cricket007:
@robertpang Thanks for the review. 

> that the way statements are pre-prepared during `init()` requires the same logic to construct the query strings in `buildStatements()`

Same logic where? `init()` is calling `buildStatements()`. I'm not seeing where anything is duplicated.  

> in the individual `read`, `scan`, `insert`, `update` and `delete()` methods and I have concern about this duplication of codes

In the individual methods, you have similar `QueryBuilder` creation, then map checking. Is that what you are referring to? 

@jixuan1989 

CLI Opts:  

```
-db com.yahoo.ycsb.db.CassandraCQLClient 
-P /opt/ycsb/workloads/benchmark-workloada 
-threads 35 
-p maxexecutiontime=1200 
-p operationcount=900000000 
-p hosts=cassandra01,cassandra02,cassandra03 
-p cassandra.connecttimeoutmillis=50000 
-p cassandra.writeconsistencylevel=QUORUM 
-p cassandra.readconsistencylevel=QUORUM 
-tYCSB Client 0.14.0-SNAPSHOT
```

Workload A

```
recordcount=150000000
operationcount=150000000
maxexecutiontime=1200

workload=com.yahoo.ycsb.workloads.CoreWorkload

readallfields=true

readproportion=0.5
updateproportion=0.5
scanproportion=0
insertproportion=0

requestdistribution=zipfian
```

A have a shell script that looks like this - The gist here is that I split apart the stdout and stderr of `ycsb run`

```sh
YCSB_CMD="ycsb run $YCSB_OPTS"
${YCSB_HOME}/bin/${YCSB_CMD} > $OUTPUT_DIR/${OUTPUT_FILE}.log 2> $OUTPUT_DIR/${OUTPUT_FILE}.err 
```

And when I do something like `grep FAILED workloada.log`, I see 

```
[READ-FAILED], Operations, 1027229
[READ-FAILED], AverageLatency(us), 7635.317831759034
[READ-FAILED], MinLatency(us), 535
[READ-FAILED], MaxLatency(us), 221823
[READ-FAILED], 95thPercentileLatency(us), 18527
[READ-FAILED], 99thPercentileLatency(us), 29007
[READ-FAILED], 535, 1.0
[READ-FAILED], 539, 1.0
[READ-FAILED], 544, 1.0
[READ-FAILED], 545, 1.0
[READ-FAILED], 570, 1.0
...
```

That actually was a run off of **master**. I think I didn't notice that before because of the histogram (?) output at the end. Seems to be a new feature.... It prints for all the methods. For example, 

```
[CLEANUP], Operations, 35
[CLEANUP], AverageLatency(us), 64046.08571428571
[CLEANUP], MinLatency(us), 1
[CLEANUP], MaxLatency(us), 2242559
[CLEANUP], 95thPercentileLatency(us), 8
[CLEANUP], 99thPercentileLatency(us), 2242559
[CLEANUP], 1, 31.0
[CLEANUP], 2, 1.0
[CLEANUP], 8, 1.0
[CLEANUP], 36, 1.0
[CLEANUP], 2242559, 1.0
```

Anyways, a sample run off this branch, also has READ-FAILED, but the READ numbers are . 

**master**

```
[READ], Operations, 1642900
[READ], AverageLatency(us), 8385.432664191369
[READ], MinLatency(us), 568
[READ], MaxLatency(us), 221311
[READ], 95thPercentileLatency(us), 19759
[READ], 99thPercentileLatency(us), 30863
[READ], Return=OK, 1642900
[READ], Return=NOT_FOUND, 1027227
[READ], Return=ERROR, 2

[READ-FAILED], Operations, 1027229
[READ-FAILED], AverageLatency(us), 7635.317831759034
[READ-FAILED], MinLatency(us), 535
[READ-FAILED], MaxLatency(us), 221823
[READ-FAILED], 95thPercentileLatency(us), 18527
[READ-FAILED], 99thPercentileLatency(us), 29007
```

**cassandra-preparedstatements**

```
[READ], Operations, 2454539
[READ], AverageLatency(us), 6109.3850209754255
[READ], MinLatency(us), 456
[READ], MaxLatency(us), 190207
[READ], 95thPercentileLatency(us), 15463
[READ], 99thPercentileLatency(us), 25567
[READ], Return=OK, 2454539
[READ], Return=NOT_FOUND, 1460773

[READ-FAILED], Operations, 1460773
[READ-FAILED], AverageLatency(us), 4937.38611132599
[READ-FAILED], MinLatency(us), 376
[READ-FAILED], MaxLatency(us), 136575
[READ-FAILED], 95thPercentileLatency(us), 12647
[READ-FAILED], 99thPercentileLatency(us), 20271
```

So, for these runs, 49% increase in operations (likely due to lower latency), but a 42% increase in READ failures.

robertpang:
> @robertpang Your codes are more concise. It uses less API of Datastax Java Driver so that it is more easy to be understood. Do you notice the same problem (READ-FAILED events)?

@jixuan1989 No. I don't encounter READ-FAILED events in my YCSB run with my changes.

> @robertpang Thanks for the review.

>> that the way statements are pre-prepared during init() requires the same logic to construct the query strings in buildStatements()

> Same logic where? init() is calling buildStatements(). I'm not seeing where anything is duplicated.

>> in the individual read, scan, insert, update and delete() methods and I have concern about this duplication of codes

> In the individual methods, you have similar QueryBuilder creation, then map checking. Is that what you are referring to?

@cricket007 Yes. I am referring the duplicative QueryBuilder calls that appear in both buildStatements() and in the individual read, scan, insert, update and delete() methods. It requires them to be in sync with each other.

cricket007:
@robertpang Unless I am mistaken, we both have taken a similar approach for only building the queries when needed. 

In mine, I only build a subset of queries in `buildStatements()`, and depending on the value of `readallfields`. Regarding your comment, I have a map of maps, where the `fieldCount` acts as the size of "buckets" for queries built later in read, scan, insert/update. 

delete is just a one-off query. I don't imagine a significant overhead between preparing in init() or upon delete() invocation.  

Also, [according to the DataStax CQL documentation, UPDATE and INSERT are identical](http://docs.datastax.com/en/cql/3.3/cql/cql_reference/cqlUpdate.html), so I didn't see the need to prepare those separately. 

```java
// Select and Scan many statements
selectManyStatements.put(i, new ConcurrentHashMap<Integer, PreparedStatement>());
scanManyStatements.put(i, new ConcurrentHashMap<Integer, PreparedStatement>());
```

I then build the query, check if it exists, then add, which [is equivalent `putIfAbsent`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html#putIfAbsent-K-V-).  
The only difference I see is that you have a `ConcurrentMap<String, PreparedStatement>`, making the query the key, and I have a `Map<Integer, PreparedStatement>` using the hashcode of the query. Hence my comment saying that I hope those are unique enough to prevent collisions. I couldn't think of any immediate alternative. 

```java
/* read() */
Map<Integer, PreparedStatement> possibleStatements = selectManyStatements.get(fields.size());

Select.Builder selectBuilder = QueryBuilder.select();
for (String col : fields) {
  ((Select.Selection) selectBuilder).column(col);
}
String statement = selectBuilder.from(table)
    .where(QueryBuilder.eq(YCSB_KEY, QueryBuilder.bindMarker())).getQueryString();

int statementHash = statement.hashCode(); // Hoping these are unique enough between statements
if (!possibleStatements.containsKey(statementHash)) {
  possibleStatements.put(statementHash,  prepare(statement, true));
}
ps = possibleStatements.get(statementHash);
```

cricket007:
@robertpang 

> No. I don't encounter READ-FAILED events in my YCSB run with my changes.

Can you provide your property file? Because I pulled your code, ran `mvn -pl com.yahoo.ycsb:cassandra-binding clean package -DskipTests=true`, overwrote my YCSB directory with that package, did another run with the properties mentioned above, and still am seeing `READ-FAILED`. 

Which again, I am seeing the same when I checkout master branch, so either I have done something wrong on that end (running bad code), there is a property that I have not accounted for, or I should recreate the keyspace and load the data again.

jixuan1989:
I reproduce the READ-FAILED error.

> [READ-FAILED], Operations, 150384
[READ-FAILED], AverageLatency(us), 193.32657729545696
[READ-FAILED], MinLatency(us), 117
[READ-FAILED], MaxLatency(us), 48415
[READ-FAILED], 95thPercentileLatency(us), 298
[READ-FAILED], 99thPercentileLatency(us), 396
[READ-FAILED], 117, 1.0
[READ-FAILED], 118, 2.0
[READ-FAILED], 120, 2.0
[READ-FAILED], 121, 6.0
[READ-FAILED], 122, 8.0
[READ-FAILED], 123, 5.0
[READ-FAILED], 124, 11.0
[READ-FAILED], 125, 16.0
[READ-FAILED], 126, 22.0

I need more time to find out why..

robertpang:
@cricket007 
> Also, according to the DataStax CQL documentation, UPDATE and INSERT are identical, so I didn't see the need to prepare those separately.

There are some subtle semantic differences between INSERT and UPDATE not covered in DataStax's documentation. A couple of examples are like when TTL is involved as described [here](
https://ajayaa.github.io/cassandra-difference-between-insert-update/) or when a non-key column is set to null ([here](https://stackoverflow.com/a/23705766)). A properly written Cassandra application will want to use INSERT vs UPDATE properly and so it is better for YCSB Cassandra binding to reflect how those applications use those statements.

robertpang:
> @robertpang Unless I am mistaken, we both have taken a similar approach for only building the queries when needed. In mine, I only build a subset of queries in buildStatements(), and depending on the value of readallfields. Regarding your comment, I have a map of maps, where the fieldCount acts as the size of "buckets" for queries built later in read, scan, insert/update. delete is just a one-off query. I don't imagine a significant overhead between preparing in init() or upon delete() invocation.

@cricket007 Yes, our approaches are similar but I think we can minimize the changes just preparing the statements on-demand and caching them as they are issued from read, scan, ... methods. This way the code changes will be more concise and easy to review and verify for correctness. There may be performance benefits of pre-preparing the statements ahead of time in the init() method but the benefits will be marginal and diminish as YCBS runs for longer and longer period of time.

robertpang:
@cricket007 I don't use any property file but here are the parameters I specify on command-line:

ycsb run cassandra-cql -p hosts=<host> -P workloads/workload... \\
   -p recordcount=1000000 -p operationcount=10000000 \\
   -p cassandra.readconsistencylevel=QUORUM -p cassandra.writeconsistencylevel=QUORUM \\
   -p threadcount=256 -p maxexecutiontime=90

I ran the 6 core workloads.

cricket007:
@robertpang I've ran your branch multiple times now, and even rebuilt my cluster to use the latest DSE version, and I still get READ-FAILED on just workload A. Even using the CLI options you gave. 

That being said, my setup is now a 3 node 5.1.5 cluster in VirtualBox, installed by Vagrant, which I've done [all the recommended tuning](http://docs.datastax.com/en/dse/5.1/dse-admin/datastax_enterprise/config/configRecommendedSettings.html) onto. 

Any ideas?

robertpang:
> @robertpang I've ran your branch multiple times now, and even rebuilt my cluster to use the latest DSE version, and I still get READ-FAILED on just workload A. Even using the CLI options you gave.
>
> That being said, my setup is now a 3 node DSE 5.1.5 (Cassandra 3.11.0.1900) cluster in VirtualBox, installed by Vagrant, which I've done all the recommended tuning onto.
>
> Any ideas?

@cricket007 Since you encounter READ-FAILED with the master branch, I guess you have a general issue with your setup. I am unsure what may have caused this. Can you catch the  exception that causes READ-FAILED and dump further info and see?

robertpang:
>> @cricket007
I'd say your PR is more focused than #930. You only concern yourself with the core workload value for readallfields=true. You therefore aren't reading fieldcount from the properties, nor do you care about checking fields.size() in any of the methods. Only insert & update use values.size().
>>
> Thanks for your review. I think there is a misunderstanding. My PR does not assume any specific workload. That is the reason why it does not depend on any of the core workload properties. On the other hand, your PR #930 depends heavily on those properties and pre-prepares the statements in init() specifically for core workload. For other 3rd-party workloads, say one that updates 2 fields in one UPDATE statement (possible by writing their own Java class instead of using CoreWorkload.java), PR #930 will fall back to preparing the statement on-the-fly. So your PR is essentially special-casing for core workloads.

Here is an example of a non-core [time-series workload](https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/com/yahoo/ycsb/workloads/TimeSeriesWorkload.java#L731) in YCSB that reads a random set of fields (not all and not one) for your reference.

cricket007:
Closing for #1051

PR #934, opened 2017-02-24
Add Google Cloud Spanner binding.

scode:
This adds a (basic) Google Cloud Spanner binding. Caveats:

* Only application default credentials are supported.
* Primary key values are not properly escaped during reads because
  I could not find the necessary information to do so with Spanner's
  SQL dialect (probably my mistake).

### DIFF
--- a/googlespanner/README.md
+++ b/googlespanner/README.md
@@ -21,8 +21,9 @@ LICENSE file.
 
 * Only Application Default Credentials are supported. In other words,
   if running on a Google Cloud instance with permissions granted when
-  creating it, authentication will seamlessly work. But use of
-  explicit keys is not implemented.
+  creating it, authentication will seamlessly work, but you cannot
+  authenticate using a private key file (see https://cloud.google.com/docs/authentication#getting_credentials_for_server-centric_flow
+  and https://developers.google.com/identity/protocols/application-default-credentials).
 * The primary key is not escaped in the implementation, and so the workload
   generator must not generate keys containing characters not safe to put
   within single-quotes in Spanner SQL.
--- a/googlespanner/README.md
+++ b/googlespanner/README.md
@@ -58,6 +58,6 @@ No special process required other than specifying `googlespanner` as
 the name of the binding. Example:
 
 ```
-./bin/ycsb load googlespanner -P wload -threads 20
+./bin/ycsb load googlespanner -P wload
 ./bin/ycsb run googlespanner -P wload -threads 20 -target 200
 ```
--- a/googlespanner/README.md
+++ b/googlespanner/README.md
@@ -1,0 +1,19 @@
+<!--
+Copyright (c) 2016 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+# Google Spanner Driver for YCSB
+
--- a/googlespanner/README.md
+++ b/googlespanner/README.md
@@ -58,6 +58,6 @@ No special process required other than specifying `googlespanner` as
 the name of the binding. Example:
 
 ```
-./bin/ycsb load googlespanner -P wload
+./bin/ycsb load googlespanner -P wload -threads 20
 ./bin/ycsb run googlespanner -P wload -threads 20 -target 200
 ```
--- a/googlespanner/README.md
+++ b/googlespanner/README.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2016 YCSB contributors. All rights reserved.
+Copyright (c) 2017 YCSB contributors. All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License"); you
 may not use this file except in compliance with the License. You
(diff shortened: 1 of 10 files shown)


scode:
Note that I've never worked on the code base before and this was a "minimal until it worked for me" kind of effort. Some inspiration taken from other bindings. Please let me know if I'm not following best practices etc.

googlespanner/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+      <artifactId>google-cloud-spanner</artifactId>
+      <version>${googlespanner.version}</version>
+    </dependency>
+    <!-- A sufficiently new Guava is required by the Spanner client. -->
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>21.0</version>
+    </dependency>
+    <!-- Additionally, avoid a conflicting older -jdk15 variant of Guava ending up on the classpath. -->
```
scode: Note that the `-jdk15` variant is still pulled in for other bindings. However when running with a binding tailored classpath as done by:

```
./bin/ycsb run googlespanner -P wload -threads 20 -target 200
```

This does not cause a problem. However, if uberjars are generated or runs are otherwise done in such a way as to contain a global set of dependencies, it would still fail.

The failures happen at runtime due to a missing method in one of the Guava classes.

googlespanner/pom.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2016 YCSB contributors. All rights reserved.
```
risdenk: Can this be updated to 2017?

googlespanner/README.md
```diff
@@ -0,0 +1,19 @@
+<!--
+Copyright (c) 2016 YCSB contributors. All rights reserved.
```
risdenk: 2017

googlespanner/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+# Google Spanner Driver for YCSB
```
risdenk: Looks like this is empty? It would be great to follow one of the other bindings and talk about how to use.

googlespanner/src/main/java/com/yahoo/ycsb/db/GoogleSpannerClient.java
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+package com.yahoo.ycsb.db;
```
risdenk: This is something new that we are trying to do. Can you make sure that this is in its own package? com.yahoo.ycsb.db.googlespanner or something similar?

This will be a change in most of the files.

googlespanner/src/main/java/com/yahoo/ycsb/db/GoogleSpannerClient.java
```diff
@@ -0,0 +1,185 @@
...
+      this.spanner.closeAsync().get();
+    } catch (InterruptedException | ExecutionException e) {
+      throw new DBException(e);
+    }
+  }
+
+  @Override
+  public Status read(String table, String key, Set<String> fields,
+          HashMap<String, ByteIterator> result) {
+    // XXX: Key is not escaped. Unclear to me at this time how to do this in Spanner's SQL dialect.
```
risdenk: What happens in this case? Errors?

googlespanner/README.md
scode: Will do, sorry about that.

googlespanner/src/main/java/com/yahoo/ycsb/db/GoogleSpannerClient.java
scode: Sure, that's my preference too. I just aped what another binding was doing and wasn't sure what the preference was.

googlespanner/src/main/java/com/yahoo/ycsb/db/GoogleSpannerClient.java
scode: Standard escaping problem. Depending on whether or not we want to rely on the workload generating keys that can be naively put between single quotes, this code is wrong and arbitrary SQL may be executed. Whether or not this is acceptable given the intentions of YCSB workload generation I'm not sure, but I timed out on attempting to find out what escaping is required in the Spanner dialect.

scode:
@risdenk There is an SQL language for *queries*, but Spanner doesn't allow updates through SQL. I haven't looked at https://cloud.google.com/spanner/docs/partners/drivers but unless it's implementing rewriting of SQL itself I don't expect it to provide that. The mechanism used here to talk to Spanner is that which is documented in the Spanner documentation.

risdenk:
@scode Thanks for the SQL explanation.

scode:
@risdenk Addressed feedback.

I do want to call attention to the comment in the code about conflicting dependencies. I don't know if this solution is acceptable since I don't expect it to run in a context where all sub modules dependencies have been put on the class path.

The dependency exclusion I added to the `googlespanner` submodule here can probably be duplicated on other modules, provided that we don't care about JDK 1.5 support (the conflicting dependency being a `-jdk5` version of Guava). Since Guava is extremely high quality and release engineered, I don't expect backwards compatibility concerns being very likely - but they are possible and I'm not set up to test other Guava consumers that may be affected by such a change.

googlespanner/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Limitations and caveats
+
+* Only Application Default Credentials are supported. In other words,
+  if running on a Google Cloud instance with permissions granted when
+  creating it, authentication will seamlessly work. But use of
+  explicit keys is not implemented.
```
risdenk: explicit keys being like access key? This is slightly confusing with the next line talking about primary keys for inserting/reading.

scode: Adjusted to hopefully be clearer, and link to documentation.

Yes, like an AWS access key.

siamaktz:
Thank you for contributing this, we actually have an official YCSB client for Cloud Spanner by Google that we'd like to merge. Would it be okay to drop this in favor of the official one? Thanks.

scode:
Sure

PR #937, opened 2017-03-03
Change the write policy to REPLACE_ONLY

tiboratAS:
The original Aerospike interface layer was created when the REPLACE_ONLY option was not available.  This provides a policy more in line with the policies that other databases have implemented their interface layers.


risdenk:
Thanks @tiboratAS.

Note to release manager: This should be called out in the release notes.

PR #940, opened 2017-03-09
[arangodb] Updated ArangoDB binding (Java driver 3.1.4)

mpv1989:
Updated arangodb-java-driver to 3.1.4. This version is only compatible with ArangoDB 3.1.0 and above.

This allows a better comparison between the http java-driver and the vst (VelocyStream) java-driver which is used within binding `arangodb3`.

### DIFF
--- a/arangodb/src/main/java/com/yahoo/ycsb/db/ArangoDBClient.java
+++ b/arangodb/src/main/java/com/yahoo/ycsb/db/ArangoDBClient.java
@@ -63,7 +63,7 @@ public class ArangoDBClient extends DB {
    * The database name to access.
    */
   private static String databaseName = "ycsb";
-
+  private static String collectionName;
   /**
    * Count the number of times initialized to teardown on the last
    * {@link #cleanup()}.
@@ -92,6 +92,8 @@ public class ArangoDBClient extends DB {
       }
 
       Properties props = getProperties();
+
+      collectionName = props.getProperty("table", "usertable");
 
       // Set the DB address
       String ip = props.getProperty("arangodb.ip", "localhost");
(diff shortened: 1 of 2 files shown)


risdenk:
There is already an `arangodb3` binding in master? This hasn't been released yet in an official release but you should be able to compile and build with it.

Let us know if that doesn't handle the case you need.

mpv1989:
Yes, I implemented binding `arangodb3` in an other pull request. It uses the new arangodb-java-driver (version 4.1.x) which uses the binary transport protocol **VelocyStream**. Which is supported by ArangoDB 3.1.0 and above.

Within this pull request I updated the original `arangodb` binding with the latest version of the **old** arangodb-java-driver (version 3.1.4) which uses **http** as transport protocol and is also compatible with the ArangoDB 3.1.0.

This allows comparison the http driver and the velocystream driver with ArangoDB 3.1 by using both bindings `arangodb` and `arangodb3`.

risdenk:
Ah but changing the existing arangodb binding would mean it is no longer compatible with <3.1.0?

This sounds like it is more of a two different types of drivers for ArangoDB 3.x. Maybe instead of modifying the old binding, it would make sense to add a property to arangodb3 and then in the init choose which implementation driver to use? This is how the Elasticsearch binding works with either the node or transport client.

mpv1989:
I understand your point of view.

But first I see the problem that both drivers has the same groupId:artifactId and only differ in its version which maven shouldn't be able to deal with in one binding.

Beside that the current `arangodb` binding only supports ArangoDB <2.8 which is last updated at May 2016 and shouldn't be interesting anymore.

risdenk:
@mpv1989 The idea to keep the old binding and the new binding across major versions allows for benchmarking the major versions against either other for regression. This means that the `arangodb` binding is for 2.x and then the `arangodb3` binding would be for 3.x. Eventually the old binding can be phased out.

mpv1989:
Ok, I will take a look at it. Thanks

PR #947, opened 2017-03-27
A general "refresh" to the Accumulo binding

joshelser:
* Expand on the README, covering table creation and best-practices for table config
* Avoid unnecessary `Text` object creations (in loops and instead of `byte[]` usage)
* Use a `ConcurrentHashMap` to better match the DB API
* Fix error messages and always call `printStackTrace()` on exceptions
* Use `BATCHED_OK` instead of `OK` in `insert()` (more correct)

Tagging some potentially interested folks.. @busbey @madrob @keith-turner

### DIFF
--- a/accumulo/README.md
+++ b/accumulo/README.md
@@ -36,7 +36,42 @@ Git clone YCSB and compile:
     cd YCSB
     mvn -pl com.yahoo.ycsb:aerospike-binding -am clean package
 
-### 3. Load Data and Run Tests
+### 3. Create the Accumulo table
+
+By default, YCSB uses a table with the name "usertable". Users must create this table before loading
+data into Accumulo. For maximum Accumulo performance, the Accumulo table must be pre-split. A simple
+Ruby script, based on the HBase README, can generate adequate split-point. 10's of Tablets per
+TabletServer is a good starting point. Unless otherwise specified, the following commands should run
+on any version of Accumulo.
+
+    $ echo 'num_splits = 20; puts (1..num_splits).map {|i| "user#{1000+i*(9999-1000)/num_splits}"}' | ruby > /tmp/splits.txt
+    $ accumulo shell -u <user> -p <password> -e "createtable usertable"
+    $ accumulo shell -u <user> -p <password> -e "addsplits -t usertable -sf /tmp/splits.txt"
+    $ accumulo shell -u <user> -p <password> -e "config -t usertable -s table.cache.block.enable=true"
+
+Additionally, there are some other configuration properties which can increase performance. These
+can be set on the Accumulo table via the shell after it is created. Setting the table durability
+to `flush` relaxes the constraints on data durability during hard power-outages (avoids calls
+to fsync). Accumulo defaults table compression to `gzip` which is not particularly fast; `snappy`
+is a faster and similarly-efficient option. The mutation queue property controls how many writes
+that Accumulo will buffer in memory before performing a flush; this property should be set relative
+to the amount of JVM heap the TabletServers are given.
+
+Please note that the `table.durability` and `tserver.total.mutation.queue.max` properties only
+exists for >=Accumulo-1.7. There are no concise replacements for these properties in earlier versions.
+
+    accumulo> config -s table.durability=flush
+    accumulo> config -s tserver.total.mutation.queue.max=256M
+    accumulo> config -t usertable -s table.file.compress.type=snappy
+
+On repeated data loads, the following commands may be helpful to re-set the state of the table quickly.
+
+    accumulo> createtable tmp --copy-splits usertable --copy-config usertable
+    accumulo> deletetable --force usertable
+    accumulo> renametable tmp usertable
+    accumulo> compact --wait -t accumulo.metadata
+
+### 4. Load Data and Run Tests
 
 Load the data:
 
--- a/accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
+++ b/accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
@@ -18,17 +18,25 @@
 
 package com.yahoo.ycsb.db.accumulo;
 
-import com.yahoo.ycsb.ByteArrayByteIterator;
-import com.yahoo.ycsb.ByteIterator;
-import com.yahoo.ycsb.DB;
-import com.yahoo.ycsb.DBException;
-import com.yahoo.ycsb.Status;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.Vector;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.AccumuloSecurityException;
 import org.apache.accumulo.core.client.BatchWriter;
 import org.apache.accumulo.core.client.BatchWriterConfig;
 import org.apache.accumulo.core.client.Connector;
+import org.apache.accumulo.core.client.IteratorSetting;
 import org.apache.accumulo.core.client.MutationsRejectedException;
 import org.apache.accumulo.core.client.Scanner;
 import org.apache.accumulo.core.client.TableNotFoundException;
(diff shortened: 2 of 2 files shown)


accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
```diff
@@ -100,62 +104,72 @@ public void init() throws DBException {
...
    */
-  public void getTable(String t) throws TableNotFoundException {
-    if (bw != null) { // Close the existing writer if necessary.
-      try {
-        bw.close();
-      } catch (MutationsRejectedException e) {
-        // Couldn't spit out the mutations we wanted.
-        // Ignore this for now.
-        System.err.println("MutationsRejectedException: " + e.getMessage());
+  public BatchWriter getWriter(String t) throws TableNotFoundException {
```
phrocker: why not just call it table? Vestigial?

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
```diff
@@ -100,62 +104,72 @@ public void init() throws DBException {
...
-    scanScanner = connector.createScanner(t, Authorizations.EMPTY);
-
-    table = t; // Store the name of the table we have open.
+    final String numThreadsValue = getProperties().getProperty("accumulo.batchWriterThreads");
+    // Try to saturate the client machine.
+    int numThreads = Math.max(1, Runtime.getRuntime().availableProcessors() / 2);
+    if (null != numThreadsValue) {
+      numThreads = Integer.parseInt(numThreadsValue);
+    }
+    System.out.println("Using " + numThreads + " threads to write data");
```
phrocker: Does it matter if this println remains? Given what the harness does I'm inclined not to really care if this is here.

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
```diff
@@ -165,39 +179,41 @@ public void getTable(String t) throws TableNotFoundException {
...
         Value v = entry.getValue();
         byte[] buf = v.get();
-        result.put(entry.getKey().getColumnQualifier().toString(),
+        result.put(cq.toString(),
             new ByteArrayByteIterator(buf));
       }
     } catch (Exception e) {
-      System.err.println("Error trying to reading Accumulo table" + key + e);
+      System.err.println("Error trying to reading Accumulo table " + t + " " + key);
+      e.printStackTrace();
```
phrocker: ^same?

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
```diff
@@ -100,62 +104,72 @@ public void init() throws DBException {
   @Override
   public void cleanup() throws DBException {
     try {
-      if (bw != null) {
-        bw.close();
+      Iterator<BatchWriter> iterator = writers.values().iterator();
+      while (iterator.hasNext()) {
+        BatchWriter writer = iterator.next();
+        writer.close();
```
phrocker: Can't this throw MutationsRejectedException? Should we safeguard this and remove it from the writers anyway? Will it matter at this point in the process?

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
```diff
@@ -100,62 +104,72 @@ public void init() throws DBException {
...
-        // Couldn't spit out the mutations we wanted.
-        // Ignore this for now.
-        System.err.println("MutationsRejectedException: " + e.getMessage());
+  public BatchWriter getWriter(String t) throws TableNotFoundException {
+    // tl;dr We're paying a cost for the ConcurrentHashMap here to deal with the DB api.
+    // We know that YCSB is really only ever going to send us data for one table, so using
+    // a concurrent data structure is overkill (especially in such a hot code path).
+    // However, the impact seems to be relatively negligible in trivial local tests and it's
+    // "more correct" WRT to the API.
+    BatchWriter writer = writers.get(t);
```
keith-turner: Did you look into using [MultiTableBatchWriter.getBatchWriter(String)][1]?  It does what you are doing here, but its performance may be worse.  

I looked at the MTBW implementation, it may do more reads to main memory that what you have in this patch.  The MTBW is trying to handle the case of a table being renamed, which causes more reads to main memory.

[1]: http://accumulo.apache.org/1.8/apidocs/org/apache/accumulo/core/client/MultiTableBatchWriter.html#getBatchWriter(java.lang.String)

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
busbey: This should be on stderr of it's going to be manually printed. Otherwise it'll mess up folks using stdout to read results.

busbey:
Would switching to a logging api still be in scope for the amount of effort you're looking to do here?

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
joshelser: Ahh, good catch. This was originally because of the `table` member (checkstyle rule about a local variable masking a class member). Will fix.

busbey:
Please start the commit message with "[accumulo]".

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
joshelser: At the cleanup point, there isn't much else we can do but throw an exception. The cleanup here is best-effort, but probably unnecessary (I don't have a use-case for how YCSB would reasonably recover from this state)

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
joshelser: > Otherwise it'll mess up folks using stdout to read results.

Will do!

accumulo/README.md
```diff
@@ -36,7 +36,38 @@ Git clone YCSB and compile:
...
 
-### 3. Load Data and Run Tests
+### 3. Create the Accumulo table
+
+By default, YCSB uses a table with the name "usertable". Users must create this table before loading
+data into Accumulo. For maximum Accumulo performance, the Accumulo table must be pre-split. A simple
+Ruby script, based on the HBase README, can generate adequate split-point. 10's of Tablets per
+TabletServer is a good starting point.
+
+    $ echo 'num_splits = 20; puts (1..num_splits).map {|i| "user#{1000+i*(9999-1000)/num_splits}"}' | ruby > /tmp/splits.txt
```
busbey: Excellent!

accumulo/README.md
```diff
@@ -36,7 +36,38 @@ Git clone YCSB and compile:
...
+that Accumulo will buffer in memory before performing a flush; this property should be set relative
+to the amount of JVM heap the TabletServers are given.
+
+    accumulo> config -s table.durability=flush
+    accumulo> config -s tserver.total.mutation.queue.max=256M
+    accumulo> config -t usertable -s table.file.compress.type=snappy
+
+On repeated data loads, the following commands may be helpful to re-set the state of the table quickly.
+
+    accumulo> createtable tmp -cs usertable -cc usertable
```
busbey: I don't remember if these flags have long versions that more clearly self-document setting things up. If they do, those would be preferred.

accumulo/README.md
```diff
@@ -36,7 +36,38 @@ Git clone YCSB and compile:
     cd YCSB
     mvn -pl com.yahoo.ycsb:aerospike-binding -am clean package
 
-### 3. Load Data and Run Tests
+### 3. Create the Accumulo table
```
busbey: Could you spell out what versions these commands work with?

accumulo/src/main/java/com/yahoo/ycsb/db/accumulo/AccumuloClient.java
joshelser: > I looked at the MTBW implementation, it may do more reads to main memory than what you have in this patch. The MTBW is trying to handle the case of a table being renamed, which causes more reads to main memory.

I did forget that MTBW was a thing. It would be nice if YCSB ever actually provided more than one table to write to. It would be nice to explore this as a follow-on.

joshelser:
> Would switching to a logging api still be in scope for the amount of effort you're looking to do here?

Could be. I wasn't sure if there was a reason a logger wasn't used in the first place :)

> Please start the commit message with "[accumulo]".

Will do.

Thanks for the review, folks.

accumulo/README.md
joshelser: Yup, I've switched the non-obvious ones over (everything but `-t` and `-s`).

accumulo/README.md
joshelser: Looks like >=1.7 would work with what's in the readme, with table durability and the server-side mutation buffer lacking prior. I've tried to clarify this.

busbey:
the travis failure is a checkstyle violation for line length, fyi.

joshelser:
> the travis failure is a checkstyle violation for line length, fyi.

Ah, thanks. I was unintentionally ignoring that check :)

I'll fix that up and then squash/rebase these together

joshelser:
@busbey  are you able to merge this for me? :)

risdenk:
@joshelser looks like all comments were addressed. I can merge it.

PS - Guess we keep crossing internet paths today :P

joshelser:
Thanks @risdenk!

> PS - Guess we keep crossing internet paths today :P

I like it :)

PR #965, opened 2017-05-04
[memcached] support binary protocol

solganik:
Adding support for memcached binary protocol as described in
https://github.com/memcached/memcached/blob/master/doc/protocol.txt.

Protocol can be set via memcached.protocol property of YCSB memcached workload. if specified
protocol must be "binary" or "text". If unspecified text version is used.

### DIFF
--- a/memcached/README.md
+++ b/memcached/README.md
@@ -91,6 +91,10 @@ A sample configuration is provided in
   What to do with failures; this is one of `net.spy.memcached.FailureMode` enum
   values, which are currently: `Redistribute`, `Retry`, or `Cancel`.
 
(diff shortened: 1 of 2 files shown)


busbey:
Thanks for the change! Any other improvements or cleanup you think would help the memcached binding?

solganik:
Hi Sean, project is great. currently i cant think of any missing
functionality in memcached plugin. However i think that in general some
things can be slightly improved for example data verification- currently
hashes for kept in memory, this results in  huge memory usage during
validation. are there any plans to address it ? for example doing
verification in "sliding windows" ?

Thanks.

PR #970, opened 2017-05-17
[hbase12] Add HBase 1.2+ specific client that relies on the shaded client artifact provided by those versions.

busbey:
Adds a new module that reduces the footprint of third party jars by relying on the shaded artifact provided by the apache hbase project as of 1.2.0.

A current limitation of the shaded clients is that they can't be tested using the minicluster that ships with HBase. So for now, this patch relies on the `hbase10` module for implementation and testing. The `hbase12` module and client just do pass through to the ycsb client-side logic in that module.

### DIFF
--- a/hbase12/pom.xml
+++ b/hbase12/pom.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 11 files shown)


hbase12/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+    <!-- Tests can't run without a shaded hbase testing util.
+         See HBASE-15666, which blocks us.
+         For now, we rely on the HBase 1.0 binding and manual testing.
+      -->
+    <maven.test.skip>true</maven.test.skip>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>com.yahoo.ycsb</groupId>
+      <artifactId>hbase10-binding</artifactId>
```
saintstack: This is a facade over hbase10?

saintstack: nvm... reread readme

busbey:
thanks for the review!

PR #972, opened 2017-05-26
[IMPROVMENT] Adds maven wrapper

Odravison:
This PR has a mission (and beautiful mission) to add THE MAVEN WRAPPER to this project.
Maybe you guys has one reason to not added this before. But is so easy to run the project with Maven Wrapper that I did it.

Enjoy.


allanbank:
I am not a big fan of binary blobs in git repos and always thought that was one of gradles biggest weaknesses. I know there are those that will disagree. I also don't think we have had that many issues with people not having maven installed. 

Having said that I am not completely opposed to switching to using maven wrapper but if we do that I think all of the project documentation has to also be updated.

Just my 2-cents.

jasontedor:
> I am not a big fan of binary blobs in git repos and always thought that was one of gradles biggest weaknesses.

+1

Odravison:
Ok @allanbank.
Everybody has a opinion, but not all are equals.

[]'s

manolama:
I really don't like checking in a binary if we can help it. It would be better if the scripts could download the jar if necessary.

Also, how big of an issue is it that folks don't have maven installed when they want to run YCSB?

Odravison:
Was just a improvement. An little detail that I want to contribute.

Best regards.

busbey:
This would simplify our spinup for new contributors.

@Odravison could you update things to also download the maven-wrapper jar?

Odravison:
Alright man, 
I realized that only now.

Odravison:
And, seem that ONLY me was able to close this PR :rofl:

Issue #976, opened 2017-06-05
Switching from "histogram" to "hdrhistogram" lost the histogram in ycsb output


liujiongxin:
We are using YCSB 0.9.0, and we try to parse the results in "_str(stderr + stdout)_". 

When using "histogram" for "measurementtype", p50, p75, p90, ... can be obtained from "_str(stderr + stdout)_". When using "hdrhistogram", these percentiles are gone. Maybe there is something we missed?

manolama:
Actually it's odd but with 13-SNAPSHOT I see the percentiles from the hdr histogram:
```
[UPDATE], Operations, 499.0
[UPDATE], AverageLatency(us), 33.88577154308617
[UPDATE], MinLatency(us), 14.0
[UPDATE], MaxLatency(us), 879.0
[UPDATE], 95thPercentileLatency(us), 62.0
[UPDATE], 99thPercentileLatency(us), 86.0
[UPDATE], Return=OK, 499
...
```

but not those from just the regular histogram:
```
[UPDATE], Operations, 499.0
[UPDATE], AverageLatency(us), 33.88577154308617
[UPDATE], MinLatency(us), 14.0
[UPDATE], MaxLatency(us), 879.0
[UPDATE], 95thPercentileLatency(us), 62.0
[UPDATE], 99thPercentileLatency(us), 86.0
[UPDATE], Return=OK, 499
.....
```

busbey:
Can you reproduce this error with the 0.14.0 release @liujiongxin?

liujiongxin:
What's the error? I didn't see any output difference in manolama's comment.

busbey:
Closing as stale. Please reopen if you're still having trouble.

liujiongxin:
Sorry to get back to you this late. I had a chance to test it with version 0.15.0, however the problem still exists. 

The commandline is: /opt/pkb/ycsb/bin/ycsb run hbase10 -cp /opt/pkb/hbase/conf -jvm-args '"'"' -Xmx3699m'"'"' -threads 1 -P /opt/pkb/workloada -p table=ycsb9660d6d8 -p recordcount=100000 -p operationcount=100000000 -p **measurementtype=hdrhistogram** -p **measurement.interval=intended** -p maxexecutiontime=1200 -p columnfamily=cf -p clientbuffering=false

however, the output .json file does not contain the metrics such as update p50 latency, update p95 latency, etc.

liujiongxin:
[...]

PR #983, opened 2017-06-26
Export totalHistogram for HdrHistogram measurement

liujiongxin:
Try to make HdrHistogram on par with Histogram in the exported result. This PR is specifically addressing the issue https://github.com/brianfrankcooper/YCSB/issues/976


manolama:
LGTM though we do need to add a flag so that folks can decide whether or not to dump the buckets out since, particularly with the HDR histo, the output can be pretty big.

manolama:
Committed in 2c66bc652b3e768975a5e8bf08855c40fcf9b2f9, thanks!

PR #1000, opened 2017-08-03
[core] Fix an issue where the threadid and threadCount were not passed

manolama:
to the workload client threads.


manolama:
Merged, thanks!

PR #1002, opened 2017-08-03
[core] Add an operation enum to the Workload class. This can eventually

manolama:
be used to replace the strings.


manolama:
Merged, thanks!

PR #1003, opened 2017-08-06
[core] Add the NumericByteIterator to store byte encoded longs or dou…

manolama:
…bles

for passing to db implementations as an abstraction on the byte array itorator.


manolama:

manolama:
Merged in 3e8cf8693b21d5820da057b269346638c069106f, thanks!

PR #1004, opened 2017-08-06
[core] Add the RandomDiscreteTimestampGenerator to generate a range of

manolama:
timestamps in a non-repeating random order.
Modify UnixEpochTimestampGenerator so that the random generator can
extend it.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
+++ b/core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
@@ -43,17 +43,20 @@ import java.util.concurrent.TimeUnit;
  */
 public class UnixEpochTimestampGenerator extends Generator<Long> {
 
+  /** The base timestamp used as a starting reference. */
+  protected long startTimestamp;
+  
   /** The current timestamp that will be incremented. */
-  private long currentTimestamp;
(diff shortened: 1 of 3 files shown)


manolama:

core/src/main/java/com/yahoo/ycsb/generator/UnixEpochTimestampGenerator.java
```diff
@@ -107,28 +111,35 @@ public void initalizeTimestamp(final long intervalOffset) {
...
       break;
     case HOURS:
       currentTimestamp = (System.currentTimeMillis() / 1000) +
           getOffset(intervalOffset);
+      startTimestamp = currentTimestamp;
       break;
     case DAYS:
       currentTimestamp = (System.currentTimeMillis() / 1000) +
           getOffset(intervalOffset);
+      startTimestamp = currentTimestamp;
```
risdenk: nit: could do this `startTimestamp = currentTimestamp` could go right before the return after the switch/case

manolama: good point, cleaned it up. Thanks!

manolama:
Merged in b945c6604385e9069daeedbda114eefea89eed21, thanks!

PR #1005, opened 2017-08-06
[core] Add 'basicdb.count' as a config option to track and print

manolama:
the number of unique combinations of tables, keys and fields per
database operation. This is useful for tracking distributions.


manolama:

manolama:
Merged in b1d4ac629d53343b0ea05e19e028302e84ed77e6, thanks!

Issue #994, opened 2017-07-20
S3Client could not estimate read perfomance correctly


bosher:
The S3Client does not read the whole object. 

The  problem is in the call  "objectData.read(inputStreamToByte, 0, sizeOfFile)" inside readFromStorage method. This call is returning only the beginning of the object if the object is long enough. We should have something like below:
...
int len;
int offset = 0;
while ((len = objectData.read(inputStreamToByte, offset, sizeOfFile - offset)) > 0) {
      	offset  += len;
}
...

busbey:
good find! could you put up a pull request?

bosher:
I have sent pull request #1010.

PR #1010, opened 2017-08-16
[s3] S3Client could not estimate read perfomance correctly (#994)

bosher:
The S3Client does not read the whole object.

The problem is in the call "objectData.read(inputStreamToByte, 0, sizeOfFile)" inside readFromStorage method. This call is returning only the beginning of the object if the object is long enough.


risdenk:
I think this might be able to be simplified further with IOUtils from the aws dependency already.

```java
...
import com.amazonaws.util.IOUtils;
...
// writing the stream to bytes and to results
result.put(key, new ByteArrayByteIterator(IOUtils.toByteArray(objectData)));
objectData.close();
...
```

This removes the need for manually building the bytearray anyway.

busbey:
@risdenk you okay with that as a follow-on, or want to see it here?

risdenk:
@busbey either way works for me.

Issue #701, opened 2016-04-11
[hbase10] refactor thread counting to match hbase098


busbey:
as a follow on to #692,
- make the AtomicInteger used to count threads final
- update the synchronization check to lock on said final AtomicInteger
- remove the locking around the getTable call

PR #1028, opened 2017-09-19
[hbase10] Address #701 by mimicking the same locks from the HBase 0.9…

manolama:
…8 client in the HBase 10 client.

### DIFF
--- a/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
+++ b/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
@@ -66,10 +66,12 @@ import static com.yahoo.ycsb.workloads.CoreWorkload.TABLENAME_PROPERTY_DEFAULT;
  * durability.
  */
 public class HBaseClient10 extends com.yahoo.ycsb.DB {
+  private static final AtomicInteger THREAD_COUNT = new AtomicInteger(0);
+  
+  private static final Object TABLE_LOCK = new Object();
+  
   private Configuration config = HBaseConfiguration.create();
-
-  private static AtomicInteger threadCount = new AtomicInteger(0);
-
+  
   private boolean debug = false;
 
   private String tableName = "";
@@ -82,7 +84,7 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB {
    * @See #CONNECTION_LOCK.
    */
   private static Connection connection = null;
-  private static final Object CONNECTION_LOCK = new Object();
+  //private static final Object CONNECTION_LOCK = new Object();
 
   // Depending on the value of clientSideBuffering, either bufferedMutator
   // (clientSideBuffering) or currentTable (!clientSideBuffering) will be used.
@@ -145,8 +147,8 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB {
     }
 
     try {
-      threadCount.getAndIncrement();
-      synchronized (CONNECTION_LOCK) {
+      THREAD_COUNT.getAndIncrement();
+      synchronized (THREAD_COUNT) {
         if (connection == null) {
           // Initialize if not set up already.
           connection = ConnectionFactory.createConnection(config);
@@ -179,7 +181,7 @@ public class HBaseClient10 extends com.yahoo.ycsb.DB {
     String table = getProperties().getProperty(TABLENAME_PROPERTY, TABLENAME_PROPERTY_DEFAULT);
     try {
       final TableName tName = TableName.valueOf(table);
-      synchronized (CONNECTION_LOCK) {
+      synchronized (THREAD_COUNT) {
         connection.getTable(tName).getTableDescriptor();
       }
     } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -82,7 +84,7 @@
    * @See #CONNECTION_LOCK.
    */
   private static Connection connection = null;
-  private static final Object CONNECTION_LOCK = new Object();
+  //private static final Object CONNECTION_LOCK = new Object();
```
allanbank: Just delete the line. Don't comment out.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -225,7 +224,7 @@ public void cleanup() throws DBException {
 
   public void getHTable(String table) throws IOException {
     final TableName tName = TableName.valueOf(table);
-    synchronized (CONNECTION_LOCK) {
+    synchronized (TABLE_LOCK) {
```
allanbank: #701 asked for this lock to be removed, if I am reading it correctly.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -208,14 +210,11 @@ public void cleanup() throws DBException {
...
-        synchronized (CONNECTION_LOCK) {
-          if (connection != null) {
-            connection.close();
-            connection = null;
-          }
+      synchronized (THREAD_COUNT) {
+        int threadCount = THREAD_COUNT.decrementAndGet();
+        if (threadCount <= 0 && connection != null) {
+          connection.close();
+          connection = null;
```
allanbank: This synchronized block should not be needed. The closing of the connection is protected by the decrement of the atomic integer.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -145,8 +147,8 @@ public void init() throws DBException {
     }
 
     try {
```
allanbank: Can we merge this try into the try after the "Terminate right now" comment?

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -179,7 +181,7 @@ public void init() throws DBException {
     String table = getProperties().getProperty(TABLENAME_PROPERTY, TABLENAME_PROPERTY_DEFAULT);
     try {
```
allanbank: If you merge the try above into this try you would have something like:

    if( THREAD_COUNT.getAndIncrement() == 0) {
        String table = getProperties().getProperty(TABLENAME_PROPERTY, TABLENAME_PROPERTY_DEFAULT);
        try {
            if (connection == null) {
                // Initialize if not set up already.
                connection = ConnectionFactory.createConnection(config);

                // Validate the table name.
                final TableName tName = TableName.valueOf(table);
                connection.getTable(tName).getTableDescriptor();
            }
        } catch (java.io.IOException e) {
            throw new DBException(e);
       }
    }

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
manolama: doh, thanks.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
manolama: Good point, merged em.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
manolama: Good point, yanked the sync.

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
manolama: He did mention it but I need confirmation from an HBase dev that the table fetch is thread safe in 1.X. I think it wasn't in the past.

busbey: Connection objects in hbase 1.y are threadsafe, so getTable is fine. The returned Table object is not threadsafe, but making them is supposed to be lightweight and done per-thread when needed.

manolama:
Yanked the lock so we should be good now @busbey, thanks!

hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
```diff
@@ -208,15 +201,10 @@ public void cleanup() throws DBException {
...
-      if (threadCount.get() <= 0) {
-        // Means we are done so ok to shut down the Connection.
-        synchronized (CONNECTION_LOCK) {
-          if (connection != null) {
-            connection.close();
-            connection = null;
-          }
-        }
+      int threadCount = THREAD_COUNT.decrementAndGet();
+      if (threadCount <= 0 && connection != null) {
```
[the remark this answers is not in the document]
manolama: But Alan was right in that this should only fire when the atomic int returns a 0 at which point only one thread will perform the null check and close + null the connection.

Or are you thinking of the case where there may be some initialization issue wherein a run initializes multiple threads but the initializations fail and possibly cleanup is executed on a thread before initialization begins in another thread?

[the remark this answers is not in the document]
busbey: Right exactly.

PR #1031, opened 2017-09-21
[scripts, core] cleanup bindings

### LINKED ISSUE #328: CLI help lists several currently removed datastores
our `ycsb --help` shows several datastore that we currently can't run. remove them.

busbey:
* remove mapkeeper and voldemort from command help, since we don't build them
* fix ordering of modules in root pom

closes #328


busbey:
sorry, got overzealous on waiting for the build to finish. had to push an update because the azure table binding was missing from the distribution's dependencies (so it was a coincidence of pom ordering that it worked before).

busbey:
Thanks Jason!

PR #1034, opened 2017-09-22
Switch Accumulo into multiple modules, one per minor release

busbey:


busbey:
@madrob, @joshelser  can I get a review?

accumulo1.7/src/main/conf/accumulo.properties
```diff
@@ -0,0 +1,44 @@
+# Copyright 2014 Cloudera, Inc. or its affiliates. All Rights Reserved.
```
joshelser: Intentional?

accumulo1.8/README.md
```diff
@@ -0,0 +1,118 @@
...
+LICENSE file.
+-->
+
+## Quick Start
+
+This section describes how to run YCSB on [Accumulo](https://accumulo.apache.org/). 
+
+### 1. Start Accumulo
+
+See the [Accumulo Documentation](https://accumulo.apache.org/1.7/accumulo_user_manual.html#_installation)
```
joshelser: Update for the 1.8 user manual?

accumulo1.7/src/main/conf/accumulo.properties
busbey: must be from before we switched to the "YCSB contributors" language.  not sure if copying it into a new directory is a creative act worthy of updating though.

accumulo1.8/README.md
busbey: ✅

busbey:
yeah, future proofing. I figure this will make it easier to compare perf using the 1.6 - 1.8 clients, but I have no specific reason to think they'll be different.

busbey:
I pushed an update that switched the 1.8 and 1.6 specific bindings to both refer to their respective version's user manual.

busbey:
planning to push this based on @joshelser's previous +1 later today. speak up if anyone would like more time to review further.

joshelser:
+1 from me. Thanks Sean. Nice work here.

PR #1053, opened 2017-10-22
[filestore] Database binding for storing writes to the database in files

ChristianNavolskyi:
The idea for this binding is, to use the files for multiple benchmarks runs with different databases.
Using the exact same data should be good for consistency and makes the results more repeatable and reproducible.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/workloads/TimeSeriesWorkload.java
+++ b/core/src/main/java/com/yahoo/ycsb/workloads/TimeSeriesWorkload.java
@@ -1252,7 +1252,7 @@ public class TimeSeriesWorkload extends Workload {
         boolean tagRollover = false;
         for (int i = tagCardinality.length - 1; i >= 0; --i) {
           if (tagCardinality[i] <= 1) {
-            // nothing to increment here
+            tagRollover = true; // Only one tag so needs roll over.
             continue;
           }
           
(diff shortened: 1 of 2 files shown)


filestore/src/main/java/com/yahoo/ycsb/db/FileStoreClient.java
```diff
@@ -0,0 +1,233 @@
...
+
+    if (!properties.getProperty(ENABLE_PRETTY_PRINTING, ENABLE_PRETTY_PRINTING_DEFAULT).equals("false")) {
+      gsonBuilder.setPrettyPrinting();
+    }
+
+    gson = gsonBuilder.create();
+  }
+
+  /**
+   * Reads the file with the name {@code table}_{@code key}.json.
```
busbey: This layout is going to create directories that folks have a hard time managing with even modest datasets. if we must keep with one-file-per-key, then I'd suggest making directories within the data directory. Something like 3 or 4 layers based on a prefix of the key.

filestore/src/main/java/com/yahoo/ycsb/db/FileStoreClient.java
```diff
@@ -0,0 +1,233 @@
...
+      }
+    } else {
+      result.putAll(map);
+    }
+
+    return result;
+  }
+
+  private String getDatabaseFileName(String table, String key) {
+    return outputDirectory + table + "_" + key + ".json";
```
busbey: How do I ensure that when I load these files into a datastore I'm doing it in the same order that YCSB generated them?

PR #1056, opened 2017-10-25
[accumulo] README was incorrectly referencing aerospike-binding

hinchliff:
Not sure this is really worthy of a pull-request, but the accumulo README(s) were (presumably) incorrectly referencing the aerospike-binding.


busbey:
Thanks for the catch!

PR #1060, opened 2017-11-10
Remove unnecessary getObjectMetadata() from getS3ObjectAndMetadata()

xinglin:
ObjectMetadata is available from S3Object.getObjectMetadata().
We don't need to send a separate GetObjectMetadata Request (an
HTTP HEAD request), in order to get the metadata. With this patch, 
read performance will be doubled.

Signed-off-by: Xing Lin <Xing.Lin@netapp.com>


busbey:
Thanks for the contribution!

xinglin:
Thanks for merging the change! Glad to help.

PR #1063, opened 2017-11-14
Added clients for MapR-DB and MapR-JSONDB

rohanjayaraj:
Added clients for MapR-DB (Binary) and MapR-JSONDB


rohanjayaraj:
Thank you @busbey. 

Have moved all the classes to separate package.

Test jars are not shipped by MapR. jdk9 skip is not required.

PR #1069, opened 2017-12-19
[geode] Ensure geode client will work with v1.3 by removing references to internal classes

upthewaterspout:
Geode 1.3 changed some of it's internals. The geode binding was using an
internal Geode class which means it won't compile against later versions
of geode.

### DIFF
--- a/geode/README.md
+++ b/geode/README.md
@@ -17,11 +17,11 @@ LICENSE file.
 
 ## Quick Start
 
-This section describes how to run YCSB on Apache Geode (incubating).
+This section describes how to run YCSB on Apache Geode.
 
 ### Get Apache Geode
 
-You can download Geode from http://geode.incubator.apache.org/releases/
+You can download Geode from https://geode.apache.org/releases/
 
 #### Start Geode Cluster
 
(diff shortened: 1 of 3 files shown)


busbey:
Can a Geode 1.3 client talk to a Geode 1.2 service? If not, could we do this in a way that lets folks still make use of Geode 1.2 directly? For example, the HBase and Accumulo support is broken into several modules to allow the same YCSB version to be used against different versions of those storage systems.

upthewaterspout:
@busbey - No, but the other way around works - a 1.2 client can talk to a 1.3 service. 

So maybe we should just leave the version at 1.2 but keep the rest of this PR. With the other changes in this PR to remove references to internal classes, this binding should be usable with geode clients of all versions just by changing the dependency. I can fix the PR to stick with 1.2 if that sounds good to you.

busbey:
yeah that sounds good. Worth a note in the binding's README that expressly calls that out.

I dunno how often the client side of the library matters for Geode. If it does, you could do similar to what the HBase bindings do for hbase 1.y support: provide an implementation against 1.0 and then reuse it when packaging up client bits for later release versions. That way there's only one implementation to maintain but folks can easily see the results of optimizations done on the client side if they run the binding with the newer jars.

upthewaterspout:
Sounds good! I pushed a change to stick with 1.2. What the hbase binding is doing is neat, but since geode is aiming for monthly releases we probably don't want to add a binding for each version anyway.

busbey:
thanks @upthewaterspout! If you don't mind, I'll ping you when we start the release process that includes this change to verify that Geode works as expected.

upthewaterspout:
Sure, no problem. Thanks @busbey!

PR #1072, opened 2017-12-22
Redis cluster support

j3ns:
I suggest to integrate support for Redis clusters with this pull request. A new boolean parameter `redis.cluster` is introduced to control if Redis is running in cluster mode or not.


j3ns:
The Jedis client in the proposed latest version 2.9.0 is fully compatible with Redis 2.8.x and 3.x.x (see https://github.com/xetorthio/jedis). Cluster mode has been introduced with Redis 3.x.x in 2015. The change should still work with older Redis versions if the new parameter `redis.cluster` remains `false` (default).

PR #1068, opened 2017-12-16
Proof-Of-Concept for folding the YCSB-TS fork back into YCSB after #1008 [WIP]

Vogel612:
This PullRequest is a continued development of the work performed by Andreas Bader in https://github.com/TSDBBench/YCSB-TS.

For authorship reasons the commits are currently split like this. In addition: since the language-binding for influxdb requires java 8, this pull-request also drops JDK 7 support.

There are some uncertainties for me in regards to how the binding should behave in certain situations. These are marked in InfluxDBClient.java with FIXME comments.
Guidance is very much appreciated.

Do also note that the DELETE operation was not supported (or required) in the benchmarking performed for YCSB-TS. I will amend the PullRequest to include an implementation for that operation

PR #1095, opened 2018-02-14
[core] changes to enable folding YCSB-TS back into YCSB

Vogel612:
This is the first pull-request in a series that intend to fold back the YCSB-TS fork into the mainline. 

See #1068 for an "overarching" pull request, as well as TSDBBench/YCSB-TS/issues/8


Vogel612:
Ping @manolama

Vogel612:
For now the follow-up pull-requests are suspended, until this one is merged in, because any subsequent work will require the changes from this PR. Given that merging apparently happens through squashes, I'd prefer to keep rebases and cherry-picks to a minimum. Accordingly I'll just continue working in the overarching PR, which should approach a diff of +/- 0 when follow-up PRs are merged

Vogel612:
Hey @manolama @busbey this has been sitting for a month now. Can I pretty please get a review? Thanks

manolama:
Ah got it, will complete the CR by Sunday, thanks!

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+    Map<String, List<String>> tagQueries = new HashMap<>();
+    Long timestamp = null;
+    for (String field : fields) {
+      if (field.startsWith(timestampKey)) {
+        String[] timestampParts = field.split(tagPairDelimiter);
+        if (timestampParts[1].contains(queryTimeSpanDelimiter)) {
+          // seems like this should be a more elaborate query.
+          // for now we don't support querying ranges
+          // TODO: Support Timestamp range queries
+          return Status.NOT_IMPLEMENTED;
```
manolama: ~~We should have both start and stop timestamps for the `read()` method so I'd go ahead and pass it. There are definitely much more complex queries dealing with timestamps but we really just want to focus on common features in TSDs and querying over a time-range is the most common use case.~~

Ah never mind, you're right if we are doing a read, we're looking for a data point. In that case you may want to throw a DBException here saying that range queries aren't allowed for a read.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+
+  /**
+   * Read a record from the database. Each value from the result will be stored in a HashMap
+   *
+   * @param metric    The name of the metric
+   * @param timestamp The timestamp of the record to read.
+   * @param tags      actual tags that were want to receive (can be empty)
+   * @return Zero on success, a non-zero error code on error or "not found".
+   */
+  protected abstract Status read(String metric, Long timestamp, Map<String, List<String>> tags);
```
manolama: Still need to pass through `table` though a lot of TSDs will ignore it, and add the end timestamp and convert `Long` to `long` to use the primitive.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+  /**
+   * @inheritDoc
+   * @implNote this method parses the information passed to it and subsequently passes it to the modified
+   * interface at {@link #scan(String, Long, Long, Map, AggregationOperation, int, TimeUnit)}
+   */
+  @Override
+  public final Status scan(String table, String startkey, int recordcount, Set<String> fields,
+                           Vector<HashMap<String, ByteIterator>> result) {
+    Map<String, List<String>> tagQueries = new HashMap<>();
+    Long start = null;
```
manolama: Primitives please :)

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+          // for now we don't support scanning single timestamps
+          // TODO: Support Timestamp range queries
+          return Status.NOT_IMPLEMENTED;
+        }
+        String[] rangeParts = timestampParts[1].split(queryTimeSpanDelimiter);
+        start = Long.valueOf(rangeParts[0]);
+        end = Long.valueOf(rangeParts[1]);
+      } else if (field.startsWith(groupByKey)) {
+        String groupBySpecifier = field.split(tagPairDelimiter)[1];
+        aggregationOperation = TimeseriesDB.AggregationOperation.valueOf(groupBySpecifier);
```
manolama: Might be nice to wrap this up in a DBException or WorkloadException saying that the group by wasn't in the enum.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+      } else if (field.startsWith(downsamplingKey)) {
+        String downsamplingSpec = field.split(tagPairDelimiter)[1];
+        // apparently that needs to always hold true:
+        if (!downsamplingSpec.equals(downsamplingFunction.toString() + downsamplingInterval.toString())) {
+          // FIXME instead return BAD_REQUEST?
+          System.err.print("Downsampling specification for Scan did not match configured downsampling");
```
manolama: Can throw the DBException here.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+   * @param metric    The name of the metric
+   * @param startTs   The timestamp of the first record to read.
+   * @param endTs     The timestamp of the last record to read.
+   * @param tags      actual tags that were want to receive (can be empty).
+   * @param aggreg    The aggregation operation to perform.
+   * @param timeValue value for timeUnit for aggregation
+   * @param timeUnit  timeUnit for aggregation
+   * @return A {@link Status} detailing the outcome of the scan operation.
+   */
+  protected abstract Status scan(String metric, Long startTs, Long endTs, Map<String, List<String>> tags,
```
manolama: `Long` -> `long`

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+   * Insert a record in the database. Any tags/tagvalue pairs in the specified tags HashMap and the given value
+   * will be written into the record with the specified timestamp
+   *
+   * @param metric    The name of the metric
+   * @param timestamp The timestamp of the record to insert.
+   * @param value     actual value to insert
+   * @param tags      A HashMap of tag/tagvalue pairs to insert as tagsmv c
+   * @return A {@link Status} detailing the outcome of the insert
+   */
+  protected abstract Status insert(String metric, Long timestamp, double value, Map<String, ByteIterator> tags);
```
manolama: `Long` -> `long`. And I know a lot of TSDs only support doubles (grrr) but we should have an override for signed ints too. Implementations can return a `Status.NOT_IMPLEMENTED` if they don't support storing ints.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+
+  /**
+   * NOTE: This operation is usually <b>not</b> supported for Time-Series databases.
+   * Deletion of data is often instead regulated through automatic cleanup and "retention policies" or similar.
+   *
```
manolama: Just a note for devs who see this comment, it will be more important for TSDs to support arbitrary deletion in the future due to laws like GDPR.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+   * @return Status.NOT_IMPLEMENTED or a {@link Status} specifying the outcome of deletion
+   * in case the operation is supported.
+   */
+  public Status delete(String table, String key) {
+    return Status.NOT_IMPLEMENTED;
+  }
+
+  protected final String buildTagFilter(Map<String, List<String>> tags) {
```
manolama: Add some docs about what this helper is used for. It seems SQLish so may only apply to certain TSDs and may actually be better in a DB implementation class.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+    return tagNames;
+  }
+
+
+  /**
+   * An enum containing the possible aggregation operations.
+   */
+  public enum AggregationOperation {
+    NONE, SUM, AVERAGE, COUNT;
```
manolama: Would be good to have `MAX` and `MIN` as well, those are fairly common. And javadocs would be nice :)

pom.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
```
manolama: The sorting is nice but we may want to break it out into another PR.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
Vogel612: Can not throw DBException here, because `DB#read` does not declare a throw for that. I'll be changing this to return `Status.BAD_REQUEST` though.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
Vogel612: `metric` is the equivalent to `table`. See also the callsite (`return read(table, timestamp, tagQueries;`)

As established above, range queries for reads make no sense. Not sure what you mean by "add the end timestamp"

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
Vogel612: Same problem as above. DB#scan does not declare an Exception in it's throws clauses

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
Vogel612: dropped the helper, since it's as of now only used in InfluxDB

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
```diff
@@ -0,0 +1,282 @@
+package com.yahoo.ycsb;
```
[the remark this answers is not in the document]
Vogel612: Copyright header has been added

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
Vogel612: The same problem as mentioned above applies. There is no exception declared in the super method, as such it's not possible to throw an exception here.

core/src/main/java/com/yahoo/ycsb/TimeseriesDB.java
manolama: Hmm, I hate checked exceptions, maybe we should add them later. That's fine though.

PR #1098, opened 2018-02-24
Kudu update

toddlipcon:
A few updates for the Kudu binding to improve performance and update to the latest client library. cc @danburkert


busbey:
Thanks for the update!

There are a couple of new client configurable properties with this change. Do we need to call them out in the README or are they esoteric?

PR #1109, opened 2018-03-17
fixes for tarantool

isopov:
- fix scan query from using ">" to ">="
- Add @Override annotation
- Replace logging with std.err since it is used in other DBs
- fix single read not writing to result
- change default port to 3303 since it is 3303 in default conf

### DIFF
--- a/tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
+++ b/tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
@@ -19,28 +19,25 @@ package com.yahoo.ycsb.db;
 import com.yahoo.ycsb.*;
 import org.tarantool.TarantoolConnection16;
 import org.tarantool.TarantoolConnection16Impl;
-import org.tarantool.TarantoolException;
 
+import java.io.IOException;
 import java.util.*;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
  * YCSB binding for <a href="http://tarantool.org/">Tarantool</a>.
  */
 public class TarantoolClient extends DB {
-  private static final Logger LOGGER = Logger.getLogger(TarantoolClient.class.getName());
-
   private static final String HOST_PROPERTY = "tarantool.host";
   private static final String PORT_PROPERTY = "tarantool.port";
   private static final String SPACE_PROPERTY = "tarantool.space";
   private static final String DEFAULT_HOST = "localhost";
-  private static final String DEFAULT_PORT = "3301";
+  private static final String DEFAULT_PORT = "3303";
   private static final String DEFAULT_SPACE = "1024";
 
   private TarantoolConnection16 connection;
   private int spaceNo;
 
+  @Override
   public void init() throws DBException {
     Properties props = getProperties();
 
@@ -50,11 +47,12 @@ public class TarantoolClient extends DB {
 
     try {
       this.connection = new TarantoolConnection16Impl(host, port);
-    } catch (Exception exc) {
+    } catch (IOException exc) {
       throw new DBException("Can't initialize Tarantool connection", exc);
     }
   }
 
+  @Override
   public void cleanup() throws DBException {
     this.connection.close();
   }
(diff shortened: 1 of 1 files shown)


tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -64,12 +62,12 @@ public Status insert(String table, String key, Map<String, ByteIterator> values)
     return replace(key, values, "Can't insert element");
   }
 
-  private HashMap<String, ByteIterator> tupleConvertFilter(List<String> input, Set<String> fields) {
-    HashMap<String, ByteIterator> result = new HashMap<>();
+  private static <T extends Map<String, ByteIterator>>
+      T tupleConvertFilter(List<String> input, Set<String> fields, T result) {
```
Vogel612: why introduce an out parameter here?

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -81,12 +79,11 @@ public Status insert(String table, String key, Map<String, ByteIterator> values)
...
     try {
       List<String> response = this.connection.select(this.spaceNo, 0, Arrays.asList(key), 0, 1, 0);
-      result = tupleConvertFilter(response, fields);
+      tupleConvertFilter(response, fields, result);
       return Status.OK;
-    } catch (TarantoolException exc) {
-      LOGGER.log(Level.SEVERE, "Can't select element", exc);
-      return Status.ERROR;
-    } catch (NullPointerException exc) {
+    } catch (Exception exc) {
```
Vogel612: Should be `TarantoolException | NullPointerException exc`, because we want to catch the most specific one according to Java conventions and recommendations

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -97,17 +94,16 @@ public Status scan(String table, String startkey,
                      Vector<HashMap<String, ByteIterator>> result) {
     List<List<String>> response;
     try {
-      response = this.connection.select(this.spaceNo, 0, Arrays.asList(startkey), 0, recordcount, 6);
-    } catch (TarantoolException exc) {
-      LOGGER.log(Level.SEVERE, "Can't select range elements", exc);
-      return Status.ERROR;
-    } catch (NullPointerException exc) {
+      response = this.connection.select(this.spaceNo, 0, Collections.singletonList(startkey), 0, recordcount, 6);
+    } catch (Exception exc) {
```
Vogel612: see above

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -117,10 +113,9 @@ public Status scan(String table, String startkey,
   public Status delete(String table, String key) {
     try {
       this.connection.delete(this.spaceNo, Collections.singletonList(key));
-    } catch (TarantoolException exc) {
-      LOGGER.log(Level.SEVERE, "Can't delete element", exc);
-      return Status.ERROR;
-    } catch (NullPointerException e) {
+    } catch (Exception exc) {
```
Vogel612: see above

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
isopov: To reuse this method in two cases:
- read single row where resulting map is given (and it is Map - we cannot accept HashMap as parameter)
- read a row from scan where we need to create our own map and add it to the result (and it must be HashMap - we cannot return Map)

Alternatively this method can be void with usage 
```
      HashMap<String, ByteIterator> temp = new HashMap<String, ByteIterator>();
      tupleConvertFilter(i, fields, temp);
```
but it still need to have `<T extends Map<String, ByteIterator>>` parameter.

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
isopov: How about removing `catch NullPointerException` and leave any RuntimeExceptions to machinery higher on the stack?

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -19,28 +19,25 @@
...
+import java.io.IOException;
 import java.util.*;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
  * YCSB binding for <a href="http://tarantool.org/">Tarantool</a>.
  */
 public class TarantoolClient extends DB {
-  private static final Logger LOGGER = Logger.getLogger(TarantoolClient.class.getName());
```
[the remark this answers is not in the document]
isopov: There are 6 loggers in the entire YCSB (including this one, but excluding one from tests)
And there are 41 database clients.

I dropped logger here to make this client more consistent with others.

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
```diff
@@ -19,28 +19,25 @@
 import com.yahoo.ycsb.*;
 import org.tarantool.TarantoolConnection16;
 import org.tarantool.TarantoolConnection16Impl;
-import org.tarantool.TarantoolException;
```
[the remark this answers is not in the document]
isopov: This is done for the previous review iteration of review. I can revert back to previous state or state currently in master or whatever you like, just tell what is the preferred way (preferably it should be preferable to all reviews or at least till the end if this particular review)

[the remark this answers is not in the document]
busbey: exceptions that come out of the DB layer will eventually cause the Client to die. In general, we expect report errors from underlying storage rather than abort the test run. Thus, the preferred approach is for bindings to do error handling for the datastore they are talking with.

tarantool/src/main/java/com/yahoo/ycsb/db/TarantoolClient.java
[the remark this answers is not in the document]
busbey: Please leave the logging in place then. I'd much rather we move in the direction of more logging and not less.

PR #1110, opened 2018-03-17
replace custom ThreadLocalRandom with standard one

isopov:
Seems to be leftover from Java < 7


busbey:
Java 7 only just recently got dropped, so good timing on your part!

thanks for the fix.

isopov:
@busbey standard ThreadLocalRandom was introduced in Java7, so timing was actually late for several years.

PR #1112, opened 2018-03-19
[core] Optimize toArray of all ByteIterators except RandomByteIterator

isopov:
Method nextBuf is a clever hack that outperforms Random.nextBytes
but performs poorly for all other ByteIterator implementations

This commit moves it to RandomByteIterator and adds efficient
toArray implementations for other ByteIterator classes.

Also InputStreamByteIterator.reset method that unconditionally
throws UnsupportedOperationException is fixed

Benchmarks for this change is available at https://github.com/isopov/isopov-jmh/blob/master/src/main/java/com/sopovs/moradanen/jmh/ycsb/ByteIteratorBenchmark.java

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/ByteIterator.java
+++ b/core/src/main/java/com/yahoo/ycsb/ByteIterator.java
@@ -96,9 +96,8 @@ public abstract class ByteIterator implements Iterator<Byte> {
       throw new ArrayIndexOutOfBoundsException("Too much data to fit in one array!");
     }
     byte[] ret = new byte[(int) left];
-    int off = 0;
-    while (off < ret.length) {
-      off = nextBuf(ret, off);
+    for (int i = 0; i < ret.length; i++) {
+      ret[i] = nextByte();
     }
     return ret;
   }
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/yahoo/ycsb/ByteIterator.java
```diff
@@ -56,16 +56,6 @@ public Byte next() {
 
   public abstract byte nextByte();
 
-  /** @return byte offset immediately after the last valid byte */
-  public int nextBuf(byte[] buf, int bufOff) {
```
manolama: Keep these around since a private DB implementation may rely on it.

core/src/main/java/com/yahoo/ycsb/RandomByteIterator.java
```diff
@@ -73,8 +73,7 @@ public byte nextByte() {
     return buf[bufOff - 1];
   }
 
-  @Override
-  public int nextBuf(byte[] buffer, int bufOffset) {
+  private int nextBuf(byte[] buffer, int bufOffset) {
```
manolama: Keep it public please as mentioned above.

core/src/main/java/com/yahoo/ycsb/ByteIterator.java
isopov: Done.

core/src/main/java/com/yahoo/ycsb/RandomByteIterator.java
isopov: Done.

bimargulies-google:
It looks to me as if all the requested changes were addressed.

PR #1114, opened 2018-03-23
[accumulo1.6, scripts] deprecate Accumulo 1.6 client, since Accumulo 1.6 is EOM.

busbey:
According to http://accumulo.apache.org/release/accumulo-1.6.6/

### DIFF
--- a/bin/ycsb.sh
+++ b/bin/ycsb.sh
@@ -142,6 +142,14 @@ be removed in a future YCSB release."
   BINDING_DIR="accumulo1.6"
 fi
 
+# Accumulo 1.6 deprecation message
+if [ "${BINDING_DIR}" = "accumulo1.6" ] ; then
+  echo "[WARN] The 'accumulo' client has been deprecated because Accumulo 1.6 \
+is EOM. If you are using Accumulo 1.7+ try using the 'accumulo1.7' client \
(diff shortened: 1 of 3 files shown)


busbey:
@joshelser think you could take a quick look at this?

bin/ycsb.sh
```diff
@@ -142,6 +142,14 @@ be removed in a future YCSB release."
   BINDING_DIR="accumulo1.6"
```
ctubbsii: Should also change the default binding for `accumulo` to a non-1.6 version.

busbey:
the binding name `accumulo` has already been deprecated and is planned for removal at the same time as the `accumulo1.6` binding

busbey:
Went back to look and #1034 (which deprecated the bare `accumulo` name) hasn't made it into a release yet. So we could keep it if we want to. But I've been trying to push any module with version-specific clients to always specify the version so we don't have to do breaking changes when we change where a bare name points. thoughts?

ctubbsii:
I guess the only risk with `accumulo` pointing to `accumulo1.6` is that the user gets 2 warnings, which is probably fine.

busbey:
thanks for the review @ctubbsii!

PR #1111, opened 2018-03-17
[core] Added graph workload and classes for recording and recreating …

ChristianNavolskyi:
…benchmark runs.

The graph workload should be used with coming graph databases, but can also be used with traditional relational databases or anything else in YCSB.

Since there was no README.md in the core module I added information about the additional parameters to the CHANGES.md file.

The data (values for the databases, operations and the randomly generated numbers) are stored by the *Recorder classes and can be reproduced with the corresponding *Recretor classes.

I think the recording and recreating part of my implementation could be extended to be used with all workloads if that's a desire, maybe in a future issue/pull request.

Looking forward to your constructive critics and suggestions.

PS: I ditched the filestore binding, since I decided to "record" the data this way.

PR #814, opened 2016-07-28
Add Neo4j support

ysnglt:
Neo4j is a graph oriented database. Since YCSB already supports OrientDB, I thought it might be interesting to add another graph database.
Note : this is the latest version of Neo4j (3.0), which only works with Java >= 1.8

PR #1123, opened 2018-04-05
Neo4j binding

ChristianNavolskyi:
I used #814 as a base together with #1111 for the graph specific classes.

New changes are only in the neo4j module and in some configuration files such as in `/bin/binding.properties`.

Every operation uses the `table`parameter to distinguish between a node and a relationship.

I developed this binding with a `GraphWorkload`in mind, but it also works with the other workloads, since the else case in every operation handles all requests except the ones with `table`set to `Edge.EDGE_IDENTIFIER`.


Confuze:
I know this pr is ancient but I think it's a really good addition to the project and I started writing an up-to-date binding for neo4j based on this.

Out of curiosity, (if by some miracle you remember) why did you choose an embedded neo4j instance? Pretty much all bindings in ycsb connect to dbs on standalone servers so would it not be a bit unfair to give the neo4j binding the advantage of having an embedded db?

c15yi:
Hey @Confuze, I wouldn't have expected to hear from this again.

I needed this for my bachelors thesis and if I remember correctly I couldn't find a free Neo4j DB to use. Maybe I am mixing things up here, but it might be something along that line.
You are absolutely correct, the in memory advantage shouldn't be given to Neo4j. 
Hope this helps.

Issue #1105, opened 2018-03-06
Cannot execute YCSB


ChristianNavolskyi:
I have cloned the repository and packaged everything via

`mvn clean package`

but when I try to run it by

`bin/ycsb.sh load basic -P workloads/workloada`

it prints

> /usr/bin/java  -classpath /Users/navolskyi/Desktop/YCSB/conf:/Users/navolskyi/Desktop/YCSB/core/target/core-0.14.0-SNAPSHOT.jar:/Users/navolskyi/Desktop/YCSB/core/target/core-0.14.0-SNAPSHOT.jar com.yahoo.ycsb.Client -load -db com.yahoo.ycsb.BasicDB -P workloads/workloada
> Command line: -load -db com.yahoo.ycsb.BasicDB -P workloads/workloadaYCSB Client 0.14.0-SNAPSHOT
> 
> Loading workload...
> Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/htrace/core/Tracer$Builder
>         at com.yahoo.ycsb.Client.getTracer(Client.java:902)
>         at com.yahoo.ycsb.Client.main(Client.java:751)
> Caused by: java.lang.ClassNotFoundException: org.apache.htrace.core.Tracer$Builder
>         at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
>         at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
>         at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
>         at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
>         ... 2 more
>  (might take a few minutes for large data sets)

I am using macOS version 10.13.3 and Maven version 3.5.0.

As far as I can tell the shell script does not include the needed dependencies in the classpath.

busbey:
This sounds like #908. Could you try downloading and using our last release's binaries?

busbey:
In the event that you need some functionality that isn't in a release yet, could you try using the python script `bin/ycsb`?

ChristianNavolskyi:
>This sounds like #908. Could you try downloading and using our last release's binaries?

When I run the benchmark from the binary everything works as expected.

>In the event that you need some functionality that isn't in a release yet, could you try using the python script bin/ycsb?

Running it with the python script works within my IDE, strange that the other scripts don't work.

[...]

PR #1125, opened 2018-04-06
[core] Fixed absence of dependencies when running after building.

ChristianNavolskyi:
Solves the problem mentioned in #1105.

As I am on a Unix machine the changes to the `ycsb.bat` file are not tested. 
I just copied the few lines above and changed the directory to look for the dependencies.


busbey:
This seems like it's the same solution as #908. am I missing something?

busbey:
I believe this is now fixed for the 0.14.0 release due to the inclusion of #908. please reopen if that's not the case.

PR #1096, opened 2018-02-21
improvement: support up to 1 trillon items

dbjoa:
This PR increases the two size limits:
* increase the maxinum value of 'operationcount', 'insertcount', and 'recordcount' up to '9,223,372,036,854,775,807' from '2,147,483,647'
* increase the maxinum number of items up to '1 trillion' from '10 billion' for 'ScrambledZipfianGenerator'

PR #1137, opened 2018-04-18
update Int to Long to support more tha 2147483647 records/ops

stutiredboy:
1. mongodb can have billions of records/docs, 2 billions(Int Limit) may not enough for some special scenes.
2. update the mongodb-java-driver to latest 3.6.3


stutiredboy:
#1096 issued before, so I close this request.

PR #1140, opened 2018-04-19
[hbase10] Remove use of guava

### LINKED ISSUE #1136: [hbase12,hbase14,hbase20] Setting clientbuffering=true causes NoClassDefFoundError: com/google/common/base/Preconditions
I set clientbuffering=true (this enables clientSideBuffering) on the command line. The job fails to launch because no guava on CLASSPATH providing Precondition (clientSideBuffering goes a slightly different codepath which has a Precondition check).

It works in hbase10 and there is a guava jar in the bindings lib dir for hbase10. There is no such guava in the hbase12 bindings lib dir.

Workaround is to copy the guava from hbase10 bindings to hbase12 for now.

It looks like guava comes in transitively. In hbase12, the client is shaded so we do not (intentionally) have a transitive include. Could add explicit dependency mention in the hbase12 pom.

busbey:
- module previously had an undeclared dependency on guava
- removed VisibleForTesting in favor of a comment
- removed Preconditions.checkNotNull in favor of NPE through use

closes #1136

### DIFF
--- a/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
+++ b/hbase10/src/main/java/com/yahoo/ycsb/db/HBaseClient10.java
@@ -15,9 +15,6 @@
 
 package com.yahoo.ycsb.db;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
-
 import com.yahoo.ycsb.ByteArrayByteIterator;
 import com.yahoo.ycsb.ByteIterator;
 import com.yahoo.ycsb.DBException;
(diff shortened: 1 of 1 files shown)


saintstack:
+1 from me. Thanks.

busbey:
I tested this by running the hbase12 binding in particular both with and without client side buffering:

```
# time ./bin/ycsb load hbase12 -P workloads/workloada -cp /etc/hbase/conf -p table=test:a_table -p columnfamily=family -p clientbuffering=true
...
real	0m1.910s
user	0m3.832s
sys	0m0.281s
[ycsb-hbase12-binding-0.14.0-SNAPSHOT]# time ./bin/ycsb load hbase12 -P workloads/workloada -cp /etc/hbase/conf -p table=test:a_table -p columnfamily=family
...
real	0m3.347s
user	0m4.577s
sys	0m0.503s
```

PR #1150, opened 2018-05-21
[licensing] add missing headers called out by the RAT plugin.

busbey:
While getting ready to cut the staging branch to run the 0.14.0 release, I hit a bunch of "files don't have header" errors from the rat check


busbey:
copyright dates are all based on when the file had edits done, fyi.

PR #1157, opened 2018-05-28
[cassandra] support load balancing policies: roundrobin, dcawareroundrobin

dbjoa:
This PR supports a new load balancing policy "roundrobin" in addition to the default one, "dcawareroundrobin".


cassandra/pom.xml
```diff
@@ -76,6 +76,11 @@ LICENSE file.
 	    <type>zip</type>
 	    <scope>test</scope>
 	  </dependency>
+     <dependency>
+       <groupId>com.datastax.cassandra</groupId>
+       <artifactId>cassandra-driver-core</artifactId>
+       <version>3.3.0</version>
+     </dependency>
```
cricket007: This dependency is already here at the top of these dependencies. If you need `3.3.0`, then find and change `cassandra.cql.version`

robertpang:
@busbey Sure. Will get to this today.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -90,6 +91,8 @@
   public static final String KEYSPACE_PROPERTY_DEFAULT = "ycsb";
   public static final String USERNAME_PROPERTY = "cassandra.username";
   public static final String PASSWORD_PROPERTY = "cassandra.password";
+  public static final String LOAD_BALANCING_POLICY_PROPERTY = "cassandra.loadbalancingpolicy";
+  public static final String LOAD_BALANCING_POLICY_PROPERTY_DEFAULT = "dcawareroundrobin";
```
robertpang: Making the load-balancing policy as a configurable parameter is good. But I am unsure if DC-aware round-robin policy alone should be the default. Apache Cassandra's own default is datacenter-aware + token-aware policy together.
https://docs.datastax.com/en/developer/java-driver/3.5/manual/load_balancing/. Let's not have a default here.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -163,6 +166,8 @@ public void init() throws DBException {
 
         String keyspace = getProperties().getProperty(KEYSPACE_PROPERTY,
             KEYSPACE_PROPERTY_DEFAULT);
+        String policyname = getProperties().getProperty(LOAD_BALANCING_POLICY_PROPERTY,
+            LOAD_BALANCING_POLICY_PROPERTY_DEFAULT);
```
robertpang: Making the load-balancing policy as a configurable parameter is good. But I am unsure if DC-aware round-robin policy alone should be the default. Apache Cassandra's own default is datacenter-aware + token-aware policy together.
https://docs.datastax.com/en/developer/java-driver/3.5/manual/load_balancing/. Let's not have a default here.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -172,11 +177,21 @@ public void init() throws DBException {
                 WRITE_CONSISTENCY_LEVEL_PROPERTY_DEFAULT));
 
         if ((username != null) && !username.isEmpty()) {
-          cluster = Cluster.builder().withCredentials(username, password)
-              .withPort(Integer.valueOf(port)).addContactPoints(hosts).build();
+          if ((policyname != null) && policyname.equalsIgnoreCase("roundrobin")) {
```
robertpang: Let's deduplicate the code here like this. Also, let's take the load-balancing policy full class name as the property and instance it like this:

```
Cluster.Builder builder = Cluster.builder().addContactPoints(hosts).withPort(Integer.valueOf(port));
if ((username != null) && !username.isEmpty()) {
  builder.withCredentials(username, password);
}
if ((policyname != null) && !policyname.isEmpty()) {
  .withLoadBalancingPolicy((LoadBalancingPolicy)Class.forName(policyname).newInstance());
}
cluster = builder.build();
```

busbey:
@dbjoa   are you still working on this issue?

dbjoa:
@busbey, Yes, I am.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
dbjoa: Thank you for your comment. I've updated the PR.

cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
dbjoa: Thank you for your suggestions. I've updated the PR.

busbey:
ready for review again @dbjoa?

dbjoa:
@busbey I'm ready. Here is the updated PR(https://github.com/brianfrankcooper/YCSB/pull/1157/commits/065d916c6056e269ada125ca27ee50dde2a70100).

busbey:
@robertpang, @cricket007, or @dorlaor could one of y'all take a look at the update PR?

cassandra/pom.xml
```diff
@@ -76,6 +76,11 @@ LICENSE file.
 	    <type>zip</type>
 	    <scope>test</scope>
 	  </dependency>
+    <dependency>
+      <groupId>com.datastax.cassandra</groupId>
+      <artifactId>cassandra-driver-core</artifactId>
+      <version>${cassandra.cql.version}</version>
+    </dependency>
```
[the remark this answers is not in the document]
dbjoa: @cricket007 Thank you for the comment. I've updated the PR.

cricket007:
I think it looks fine, though, I've not worked with Cassandra for two years now, so I don't know this API

PR #1162, opened 2018-05-31
[foundationdb] add a new client for FoundationDB

dbjoa:
Here is a new client for FoundationDB, which is based on the API version 5.1.7.

### DIFF
--- a/foundationdb/pom.xml
+++ b/foundationdb/pom.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you
+may not use this file except in compliance with the License. You
+may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing
+permissions and limitations under the License. See accompanying
+LICENSE file.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.yahoo.ycsb</groupId>
+    <artifactId>binding-parent</artifactId>
+    <version>0.15.0-SNAPSHOT</version>
+    <relativePath>../binding-parent</relativePath>
+  </parent>
+
+  <artifactId>foundationdb-binding</artifactId>
+  <name>FoundationDB Binding</name>
+
+  <packaging>jar</packaging>
+  <properties>
+    <!-- Tests do not run on jdk9,10,11 -->
+    <skipJDK9Tests>true</skipJDK9Tests>
+    <skipJDK10Tests>true</skipJDK10Tests>
+    <skipJDK11Tests>true</skipJDK11Tests>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>com.yahoo.ycsb</groupId>
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.foundationdb</groupId>
+      <artifactId>fdb-java</artifactId>
+      <version>${foundationdb.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <version>1.7.10</version>
+    </dependency>
+  </dependencies>
+</project>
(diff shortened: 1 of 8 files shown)


dbjoa:
In order to test the binding, you need to download the official java client from the links below and install them manually to your local maven repo.

https://www.foundationdb.org/downloads/5.1.7/bindings/java/fdb-java-5.1.7.jar
https://www.foundationdb.org/downloads/5.1.7/bindings/java/fdb-java-5.1.7-javadoc.jar

``$ mvn install:install-file -Dfile=fdb-java-5.1.7.jar -DgroupId=com.apple.foundationdb -DartifactId=fdb-java -Dversion=5.1.7 -Dpackaging=jar``

busbey:
is there an issue filed with the foundationDB folks already about publishing to maven central or otherwise adding a repository?

dbjoa:
@busbey 
Yes, the issue has been filed already: https://github.com/apple/foundationdb/issues/219

busbey:
thanks for the pointer! we'll block on that issue closing out then.

busbey:
foundation db have published client jars to maven central now. FYI, @dbjoa. I'll see if restarting the travis build is enough.

dbjoa:
@busbey I've updated the PR.

foundationdb/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+    <artifactId>binding-parent</artifactId>
+    <version>0.15.0-SNAPSHOT</version>
+    <relativePath>../binding-parent</relativePath>
+  </parent>
+
+  <artifactId>foundationdb-binding</artifactId>
+  <name>FoundationDB Binding</name>
+  <packaging>jar</packaging>
+  <repositories>
+    <repository>
```
busbey: I don't think we need the snapshot repository. can we drop it?

foundationdb/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+/**
+ * The YCSB binding for <a href="https://www.foundationdb.org">FoundationDB</a>.
+ */
+package com.yahoo.ycsb.db;
```
busbey: please put this into a dedicated package e.g. `com.yahoo.ycsb.db.foundationdb`

foundationdb/src/main/java/com/yahoo/ycsb/db/FoundationDBClient.java
```diff
@@ -0,0 +1,318 @@
...
+      String apiVersion = props.getProperty(API_VERSION, API_VERSION_DEFAULT);
+      String clusterFile = props.getProperty(CLUSTER_FILE, CLUSTER_FILE_DEFAULT);
+      String dbBatchSize = props.getProperty(DB_BATCH_SIZE, DB_BATCH_SIZE_DEFAULT);
+      dbName = props.getProperty(DB_NAME, DB_NAME_DEFAULT);
+
+      logger.info("API Version: {}", apiVersion);
+      logger.info("Cluster File: {}\n", clusterFile);
+
+      try {
+        fdb = FDB.selectAPIVersion(Integer.parseInt(apiVersion.trim()));
```
busbey: is this the bit that needs to be protected by the synchronized block? is the db object itself threadsafe?

foundationdb/src/main/java/com/yahoo/ycsb/db/FoundationDBClient.java
```diff
@@ -0,0 +1,318 @@
...
+  public void cleanup() throws DBException {
+    synchronized (INIT_COUNT) {
+      final int curInitCount = INIT_COUNT.decrementAndGet();
+      if (batchCount > 0) {
+        batchInsert();
+        batchCount = 0;
+      }
+      if (curInitCount == 0) {
+        try {
+          db.close();
```
busbey: over in `init()` we did fdb.open once for each client thread, but here we're only doing a single close. is that correct?

foundationdb/pom.xml
dbjoa: Thank you for the comment. I've dropped the snapshot repository from the PR.

foundationdb/src/main/java/com/yahoo/ycsb/db/package-info.java
dbjoa: Thank you for your suggestion. I've made the dedicated package.

foundationdb/src/main/java/com/yahoo/ycsb/db/FoundationDBClient.java
dbjoa: Thank you for the comment. The synchronized block is not needed. I've removed it from the PR.

foundationdb/src/main/java/com/yahoo/ycsb/db/FoundationDBClient.java
dbjoa: Thank you for your detailed comment. db.close() should be called for each client thread. I've updated the PR.

dbjoa:
@busbey I've changed the PR.

foundationdb/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+    <relativePath>../binding-parent</relativePath>
+  </parent>
+
+  <artifactId>foundationdb-binding</artifactId>
+  <name>FoundationDB Binding</name>
+
+  <packaging>jar</packaging>
+  <properties>
+    <!-- Tests do not run on jdk9 -->
+    <skipJDK9Tests>true</skipJDK9Tests>
```
busbey: Sorry for the churn, but a PR just landed that adds JDK10 and JDK11 test runs. Could you add the additional skip properties for those?

dbjoa: @busbey , No problem. I've added the skip properties for JDK10 and JDK11 to the updated PR.

busbey:
Thanks for getting this done!

PR #1164, opened 2018-05-31
[kudu] Add Kerberos keytab support

rebrendler:
[kudu] Allow specifying kerberos keytab and principal in configuration

### DIFF
--- a/kudu/README.md
+++ b/kudu/README.md
@@ -42,6 +42,13 @@ Additional configurations:
 * `kudu_block_size`: The data block size used to configure columns. The default
   is 4096 bytes.
 
+To connect to Kudu on a cluster that requires Kerberos authentication, some additional 
+configuration is needed:
+* `kudu_krb5_conf_file`: The location of the kerberos configuration file.  If not 
+  explicitly specified, the default configuration `/etc/krb5.conf` will be used. 
+* `kudu_krb5_keytab`: The location of the keytab to use for authentication
+* `kudu_krb5_principal`: The kerberos principal name
+
 Then, you can run the workload:
 
 ```
--- a/kudu/pom.xml
+++ b/kudu/pom.xml
@@ -51,5 +51,10 @@ LICENSE file.
       <artifactId>slf4j-log4j12</artifactId>
       <version>1.7.21</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+      <version>RELEASE</version>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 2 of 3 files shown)


busbey:
any of the prior kudu contributors up for reviewing this? @toddlipcon @jdcryans @danburkert ?

kudu/README.md
```diff
@@ -42,6 +42,13 @@ Additional configurations:
 * `kudu_block_size`: The data block size used to configure columns. The default
   is 4096 bytes.
 
+To connect to Kudu on a cluster that requires Kerberos authentication, some additional 
+configuration is needed:
```
toddlipcon: Isn't this only necessary if you plan to run YCSB for longer than the expiration time of your Kerberos tickets (often 7 days)? Normally you should be able to just 'kinit' so that you have Kerberos credentails in your ticket cache, and then the client will pick them up with no other changes necessary.

kudu/pom.xml
```diff
@@ -51,5 +51,10 @@ LICENSE file.
       <artifactId>slf4j-log4j12</artifactId>
       <version>1.7.21</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+      <version>RELEASE</version>
```
toddlipcon: is 'RELEASE' some maven magic that I'm not aware of?

kudu/src/main/java/com/yahoo/ycsb/db/KuduYCSBClient.java
```diff
@@ -127,21 +136,82 @@ private void initClients() throws DBException {
 
       Properties prop = getProperties();
 
-      String masterAddresses = prop.getProperty(MASTER_ADDRESSES_OPT,
+      final String masterAddresses = prop.getProperty(MASTER_ADDRESSES_OPT,
                                                 "localhost:7051");
```
toddlipcon: fix indentation

kudu/src/main/java/com/yahoo/ycsb/db/KuduYCSBClient.java
```diff
@@ -127,21 +136,82 @@ private void initClients() throws DBException {
...
+                new PrivilegedExceptionAction<KuduClient>() {
+                  @Override
+                  public KuduClient run() {
+                    return new KuduClient.KuduClientBuilder(masterAddresses)
+                        .defaultSocketReadTimeoutMs(DEFAULT_SLEEP)
+                        .defaultOperationTimeoutMs(DEFAULT_SLEEP)
+                        .defaultAdminOperationTimeoutMs(DEFAULT_SLEEP)
+                        .build();
+                  }
+                }));
```
toddlipcon: Instead of duplicating the builder incantation twice, what if you assign the PrivilegedExceptionAction to a variable, and then just call 'action.run()' vs 'loginUser.doAs(action.run())'?

Also, I think if you just use UserGroupInformation.getLoggedInUser() or somesuch you are guaranteed to get _some_ user which can do a doAs, so you might not need this conditional at all

kudu/pom.xml
bimargulies-google: It's very evil magic, and should be replaced by the use of a specific version.

Issue #1117, opened 2018-03-28
Release 0.14.0


busbey:
follow up from #981. There's already a fair bit of churn on master in the ~9 months since the 0.13.0-staging branch started.

Additionally, we have a good deal of testing work left over that never happened in the 0.13.0 RC process.

busbey:
yep! we'll need to test a number of bindings and any specific fixes that are supposed to be present.

The first step is to get a list up of which bindings / features /etc. It's on my todo list, but I've been a bit swamped on a different project lately.

busbey:
generally speaking, testing a binding should be straight forward. It's essentially:

1. Download RC and build binary bits
1. stand up an instance of whatever backing datastore (cluster deploy optimal, but single-node is fine)
1. follow the README for the specific binding for configuration
1. run the [recommended sequence of workloads](https://github.com/brianfrankcooper/YCSB/wiki/Core-Workloads#running-the-workloads)
1. post results here, with details about versions / sizes / etc.

using either the binding specific tarball or the omnibus is fine, but which one got used should probably be noted in the results.

busbey:
waiting on these merges:

* #1051 
* #1098 
* #1147

busbey:
merged #908

busbey:
merged #1098 and #1147 .

looks like folks are in progress on #1051

busbey:
would like to get #1148 "Update HBase 2 version to 2.0.0 GA" in for polish on the hbase addition

busbey:
As a reminder, this release needs to cover essentially everything since 0.12.0. So the release notes will end up copying some things from the 0.13.0 release notes (like known incompatibilities)

This list of things to check presumes both #1148 and #1051 will land before I cut the feature branch.

# New Topics

Needs a review because they'll get pushed in the release notes:

* We have [a contributor guide for folks who want to get involved in YCSB](https://github.com/brianfrankcooper/YCSB/blob/master/CONTRIBUTING.md)! #1026
* Timeseries workload now added #1008, #1091, #1095

# Datastore bindings that changed and will need to be tested to stay in the "verified" category

* Kudu  #879, #1098
* Solr #887
* Solr6 #887
* Aerospike #937
* Accumulo 1.6 #947, #1034, #1056, #1114
* Memcached w/binary protocol #965
* Geode v1.2 and v1.3 #995, #1069
* s3 #1010, #1021
* Google Datastore w/default credentials #984
* JDBC #910
[...]

PR #1166, opened 2018-06-07
updated ArangoDB binding (Java driver 4.4.1)

mpv1989:
Updated arangodb binding with Java driver 4.4.1. This version is compatible with ArangoDB version 2.7 up to current 3.3. I also removed binding `arangodb3` from PR #903 because it is not needed anymore.


busbey:
Does this need to be included in the 0.14.0 release (#1117) for ArangoDB to work?

Should we include it to avoid a depreciation cycle for removing the arangodb3 binding?

mpv1989:
There is no need of including it in 0.14.0 release. Both bindings `arangodb` and `arangodb3` will still work in the current stage with current ArangoDB versions. I just wanted to clean things up and get rid of this ugly `arangodb3` binding.

busbey:
Hi @mpv1989! I'm going to start rolling the next release soon. Do you have the time to add in the deprecation warning and placeholder for the `arangodb3` binding?

mpv1989:
Hi @busbey,
I'll try to take time for it today

mpv1989:
@busbey done

busbey:
thanks!

PR #1167, opened 2018-06-07
[accumulo] Use latest versions of Accumulo

ctubbsii:


pom.xml
```diff
@@ -67,8 +67,8 @@ LICENSE file.
 
     <!-- datastore binding versions, lex sorted -->
     <accumulo.1.6.version>1.6.6</accumulo.1.6.version>
-    <accumulo.1.7.version>1.7.3</accumulo.1.7.version>
-    <accumulo.1.8.version>1.8.1</accumulo.1.8.version>
+    <accumulo.1.7.version>1.7.4</accumulo.1.7.version>
+    <accumulo.1.8.version>1.9.1</accumulo.1.8.version>
```
busbey: just to confirm, the 1.9.1 client works with 1.8.0 and 1.8.1 servers?

busbey:
quote from the 0.14.0 release thread:

>  (yes, 1.9.1, not 1.8.1; 1.8.0 and 1.8.1 had critical data loss issues; 1.9.x replaces the 1.8 series)

Since there won't be more Accumulo 1.8.z releases, should we call this version-specific binding "accumulo1.9" instead of "accumulo1.8"?

pom.xml
ctubbsii: Yes, it should, and vice-versa. But, users should definitely *NOT* be using 1.8.x servers. Versions 1.8.0, 1.8.1, and 1.9.0 are subject to a series of critical data loss bugs that were fixed in 1.9.1.

Version 1.9.x is basically a continuation of 1.8.x maintenance line, but with some new superficial client APIs (superficial, as in, does not change class path requirements or RPC). Semver required a version bump, but it's basically 1.8.

ctubbsii:
> Since there won't be more Accumulo 1.8.z releases, should we call this version-specific binding "accumulo1.9" instead of "accumulo1.8"?

It could be called 1.9, but only if it completely replaces 1.8. They should not be supported concurrently. 1.9 is a continuation of 1.8 maintenance, and having them side-by-side would imply that there is a bigger gap between them than there is, and that 1.8 might still be okay to use (which, again, it is not).

busbey:
> > Since there won't be more Accumulo 1.8.z releases, should we call this version-specific binding "accumulo1.9" instead of "accumulo1.8"?

> It could be called 1.9, but only if it completely replaces 1.8. They should not be supported concurrently. 1.9 is a continuation of 1.8 maintenance, and having them side-by-side would imply that there is a bigger gap between them than there is, and that 1.8 might still be okay to use (which, again, it is not).

Right. The "accumulo1.8" binding hasn't been in a release yet. So I could include this w/a rename to "accumulo1.9" in 0.14.0 and we'd skip having to worry about confusing folks.

PR #1168, opened 2018-06-07
[build] Add missing version to rat plugin

ctubbsii:


rest/pom.xml
```diff
@@ -120,6 +120,7 @@
       <plugin>
         <groupId>org.apache.rat</groupId>
         <artifactId>apache-rat-plugin</artifactId>
+        <version>0.12</version>
```
busbey: let's set this in a pluginManagement entry on the top level pom so that we're not doing whack-a-mole if/when other modules end up needing to configure exclusions.

ctubbsii: Normally, I'd agree... but YCSB seems to do pluginManagement very inconsistently. There's only one version specified (for checkstyle) in the project root `/pom.xml`, and a few others in the `/binding-parent/pom.xml`, but most plugins throughout the modules have versions specified directly. There's no consistency. (There's not even consistency about line-endings... some POMs have LF and some have CRLF.)

It might make sense to just satisfy the build warning first... and then later tackle the problem of inconsistency throughout the project's POMs, as a separate quality effort.

ctubbsii: In either case, I don't intend to spend a lot of time on YCSB any time soon. I just wanted to make sure the issues I found while testing 0.14 RC1 Accumulo bindings had issues/PR created. So, if an additional change is needed, hopefully somebody else will be willing to pick it up.

PR #1177, opened 2018-06-25
[core] Add options for user-specified min field length

helenchw:
Allow user to specify the following parameters in `CoreWorkload` 

- the minimum field length for request generation via property  `minfieldlength`
- the minimum number of records to scan via property `minscanlength`

The original value of both parameters ("1") is retained as their default values.


busbey:
thanks!

PR #1227, opened 2018-09-10
Adding support for MySQL JSON

yashadajadhav:
MySQL supports JSON as a data type 5.7 onwards. If one wishes to benchmark how MySQL fares as a JSON data store compared to other NoSQL like Cassandra, this binding enables user to run a benchmark with a table of the format (PK, JSON)


busbey:
How much of this can be reused by adding a dependency on the existing JDBC module and extending things? Right now we'll have a lot of duplicate code to maintain in two places.

The README needs to be updated to cover just what this module is aimed at. Right now it has a bunch of generic JDBC stuff. The example of using postgres, for example, will be confusing for folks looking to test MySQL JSON specifically.

yashadajadhav:
You are right, most of this can be done by adding a dependency on the existing JDBC module and overloading or extending. I will refactor the code and update the README.

busbey:
great! please ping me when the PR is updated, in case I miss the notification for the branch push.

PR #1231, opened 2018-09-12
[elasticsearch5-rest] add forgotten elasticsearch5-rest binding.

dfquaresma:
This PR completes the elasticsearch5-rest binding in ./bin/ycsb.


busbey:
Great catch!

Issue #1184, opened 2018-07-11
JDBC Binding does not support multiple host for failover


seybi87:
Hi guys,
I was trying to benchmark CockroachDB (using the JDBC binding and the PostgreSQL JDBC connector) with respect to node failures and I came across the following issue:

JDBC support multiple host in the connection String to enable failover, e.g.:
`jdbc:postgresql://host1:port1,host2:port2/database`

Yet, the current JDBC-binding epects multiple hosts only in the following form:  
`jdbc:postgresql://host1:port1/database,jdbc:postgresql://host2:port2/database`

And as the provided connection String is split by `","` to create multiple connections  (check the respsective line 198 [JDBC-Binding](https://github.com/brianfrankcooper/YCSB/blob/a9ad5b6f27c5b8cd9af7b97dc14d5fd4c4c98280/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java#L198) ) this will result in an error for the failover connection String with multiple hosts.

Hence, the splitting identifier at line 198 should be changed to sth. which not conflicts with JDBC connection string chars.

I did a simple test by replacing `","` with `";"` and it worked without a problem for a failover connection String.

Would be great if you can apply this fix for upcoming releases

Cheers,
Daniel

busbey:
Hi Daniel! do you have the time to put together a PR?

seybi87:
Yes, I will try to do it within the next days.

busbey:
Any progress?

seybi87:
Sorry for coming back to this quite late.

Just submitted a [pull request](https://github.com/brianfrankcooper/YCSB/pull/1233)  to fix this issue

busbey:
#1233 has been merged.

PR #1233, opened 2018-09-13
Changed JDBC string splitting delimiter

seybi87:
As discussed issue #1184 , currently the JDBC connections Strings are split by `,`

This hinders the usage of the JDBC auto-failover features as connection String in the form `jdbc:postgresql://host1:port1,host2:port2/database` are not possible.

By simply changing the splitting delimiter to `;` auto-failover Strings can be used as well as the existing multiple shards connections `jdbc:postgresql://host1:port1/database;jdbc:postgresql://host2:port2/database`

Updated the readme accordingly  

Tested with PostgresSQL driver 42.2.5 and CockroachDB 2.0.1


jdbc/README.md
```diff
@@ -73,6 +73,24 @@ db.passwd=admin
...
+For connection fail-over in a DBMS cluster  specify the connection string as follows (example based on Postgres):
+
+```sh
+db.driver=org.postgresql.Driver
+db.url=jdbc:postgresql://IP1:PORT1,IP2:PORT2,IP3:PORT3/ycsb
+db.user=admin
+db.passwd=admin
+```
+
+For using multiple shards  in a DBMS cluster specify the connection string as follows by using `;`as delimiter (example based on PostgreSQL):
```
bimargulies-google: extra spaces.

jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
```diff
@@ -195,7 +195,7 @@ public void init() throws DBException {
       }
       int shardCount = 0;
       conns = new ArrayList<Connection>(3);
-      final String[] urlArr = urls.split(",");
+      final String[] urlArr = urls.split(";");
```
bimargulies-google: Please add a comment explaining what's going on here, syntactically.

seybi87: Sorry for coming back to this so late. 
In case a connection String with multiple hosts is provided it will split the String based on the delimiter ";" instead of using ",". 
This is required to enable as well the useage of the failover mechanisms of MySQL and PostgesSQL as the syntax for failover connection Strings are `jdbc:mysql/postgres://host1:port1,host2:port2,..`. 
So in case a sharding connection (as it is now) needs to be specified as follows: 
`jdbc:mysql/postgres://host1:port1;host2:port2;..`
A failover connection String as follows:
`jdbc:mysql/postgres://host1:port1,host2:port2,..`
And a sharded failover connection as follows:
`jdbc:mysql/postgres://host1:port1,host2:port2;host3:port3,host4:port4...`

PR #1255, opened 2018-11-10
[asynchbase] update to 1.8.2

timoha:
Allows to run asynchbase against hbase2


busbey:
Does asynchbase drop support for any HBase versions as a part of the update from 1.7 to 1.8?

PR #1260, opened 2018-11-19
Enabling exponential distribution in Core Workload

2121159i:
I required to use the exponential distribution provided in YCSB and found that when running YCSB with `requestdistribution=exponential` was failing. After checking the source code, it seemed that the checks for the exponential was not in the correct place, so I just moved it a bit further down. 

Please share any feedback and explain if I'm making a mistake in my reasoning here.


busbey:
Thanks for the contribution!

PR #1263, opened 2018-11-26
support trillion operation, larger than 2^31-1(int)

stutiredboy:


mariuszbarczak:
Hi Guys.
Actually recently I faced with this limitation. It'd be wonderful to have it in the next release.

PR #1266, opened 2018-12-06
Added support for Apache Crail (Incubating)

patrickstuedi:


patrickstuedi:
@busbey Sean, this is an updated PR integrating Crail (crail.apache.org) we discussed last year. Do you think the PR will be merged soon, or are there an missing pieces or issues you would like me to address?

patrickstuedi:
Ok, thanks, I'll should have this ready next week!

patrickstuedi:
The PR is ready for review. 
- Moved the classes to a crail specific package
- Setting table based on properties
- Using logging framework instead of System.out
- Also added one new parameter to control whether or not the keys can be enumerated

busbey:
Thanks for pushing on this again @patrickstuedi

I'll ping you for testing when the next release gets going.

PR #1283, opened 2019-03-11
[hbase098] document the client side buffering config.

busbey:
for whatever reason, the hbase client property for turning on client side buffering never got properly documented.

to add to the confusion, we do document a config used to tune client side buffering, but that config is ignored unless client side buffering is turned on. Also, it's used in one of our examples. :/


allanbank:
Looks good to me.

PR #1294, opened 2019-04-19
Adding the ability to connect to Cassandra using SSL connections.

voellm:
Small change to add a new flag called useSSL.  It flag adds the ability to use secure connections. The default is "false" which is how the current driver connects.  Personally I think the default should be secure connections but I dont want to break compatibility.

### DIFF
--- a/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
+++ b/cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
@@ -113,7 +113,10 @@ public class CassandraCQLClient extends DB {
 
   public static final String TRACING_PROPERTY = "cassandra.tracing";
   public static final String TRACING_PROPERTY_DEFAULT = "false";
-  
+
+  public static final String USE_SSL_CONNECTION = "cassandra.useSSL";
+  private static final String DEFAULT_USE_SSL_CONNECTION = "false";
+
   /**
    * Count the number of times initialized to teardown on the last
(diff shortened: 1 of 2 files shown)


cassandra/src/main/java/com/yahoo/ycsb/db/CassandraCQLClient.java
```diff
@@ -171,9 +174,17 @@ public void init() throws DBException {
             getProperties().getProperty(WRITE_CONSISTENCY_LEVEL_PROPERTY,
                 WRITE_CONSISTENCY_LEVEL_PROPERTY_DEFAULT));
 
+        Boolean useSSL = Boolean.parseBoolean(getProperties().getProperty(USE_SSL_CONNECTION,
+            DEFAULT_USE_SSL_CONNECTION));
+
         if ((username != null) && !username.isEmpty()) {
-          cluster = Cluster.builder().withCredentials(username, password)
-              .withPort(Integer.valueOf(port)).addContactPoints(hosts).build();
+          if (useSSL) {
```
stfeng2: Can these lines of code be merged into just a couple of lines like below?

clusterBuilder = Cluster.builder().with.....addContactPoints(hosts)
if (useSSL) ClusterBuilder = clusterBulider.withSSL()
Cluster = clusterBuilder.build()

voellm: Sure let me try and optimize it.

PR #1299, opened 2019-04-23
Updating to the latest driver and features

voellm:
Update to use the latest 2.2.3 driver along with removing "myid" which is no longer needed since Azure Cosmos supports range scans on id.  Also remove other unneeded flags like "useSinglePartition" since that is a dead feature.  All collections are partitionable now.


voellm:
Fixing history  / etc.

PR #1237, opened 2018-09-25
Added Support For Google Firestore

88sanjay:
* Added client support for google firestore.
* Updated versions of conflicting google projects
* Updated Readme for configuring the firestore client.

PR #1301, opened 2019-04-26
[googlefirestore] add standard firestore client

flint-dominic:
uses standard google-cloud-firestore per recommended on firestore website.
https://firebase.google.com/docs/firestore/quickstart


busbey:
just to make sure I'm clear, is this meant to obviate #1237 ?

googlefirestore/src/main/java/com/yahoo/ycsb/db/GoogleFirestoreClient.java
```diff
@@ -0,0 +1,247 @@
...
+     * The Cloud Firestore project ID to use when running the YCSB benchmark.
+     */
+    static final String PROJECT = "googlefirestore.projectId";
+    static final String PRIVKEYFILE = "googlefirestore.serviceAccountKey";
+    static final String DEBUG = "googlefirestore.debug";
+  }
+
+  private static final Logger LOGGER = Logger.getLogger(GoogleFirestoreClient.class);
+
+  private static Firestore db;
```
busbey: `init` gets called once per client thread. this variable either needs to be per-instance or, more likely, you should update `init` to protect access to the variable.

googlefirestore/src/main/java/com/yahoo/ycsb/db/GoogleFirestoreClient.java
```diff
@@ -0,0 +1,247 @@
...
+  @Override
+  public Status read(String table, String key, Set<String> fields, Map<String, ByteIterator> result) {
+    DocumentReference docRef = toReference(table, key);
+
+    try {
+      DocumentSnapshot docSs = docRef.get().get();
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("read: Collection: " + table + " document: " + key);
+      }
+      // If document doesn't exist, will cause NPE.
```
busbey: Is this comment correct? If we do a read on a key that doesn't exist we should return `Status.ERROR` rather than throw a NPE.

googlefirestore/src/main/java/com/yahoo/ycsb/db/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+/**
+ * The YCSB binding for Google's Cloud Firestore.
```
busbey: classes should be in a binding-specific package. e.g. `com.yahoo.ycsb.db.firestore`

googlefirestore/src/main/java/com/yahoo/ycsb/db/GoogleFirestoreClient.java
flint-dominic: removed the comment, as the code below it resolves issue.

googlefirestore/src/main/java/com/yahoo/ycsb/db/package-info.java
flint-dominic: this has been moved.

googlefirestore/src/main/java/com/yahoo/ycsb/db/GoogleFirestoreClient.java
flint-dominic: as in this should be a class only variable as far as init is concerened? or it need to be accessible in main memory thread?

themaffeo:
Hey @busbey - I'm going to be helping finish off this PR - can you let me know which issues are still outstanding?  I saw you made some comments and then @flint-dominic  pushed some changes and asked some q's.   Thanks!

bin/bindings.properties
```diff
@@ -50,6 +50,7 @@ foundationdb:com.yahoo.ycsb.db.foundationdb.FoundationDBClient
 geode:com.yahoo.ycsb.db.GeodeClient
 googlebigtable:com.yahoo.ycsb.db.GoogleBigtableClient
 googledatastore:com.yahoo.ycsb.db.GoogleDatastoreClient
+googlefirestore:com.yahoo.ycsb.db.GoogleFirestoreClient
```
busbey: This needs to be updated for the change in package name

bin/ycsb
```diff
@@ -76,6 +76,7 @@ DATABASES = {
     "geode"        : "com.yahoo.ycsb.db.GeodeClient",
     "googlebigtable"  : "com.yahoo.ycsb.db.GoogleBigtableClient",
     "googledatastore" : "com.yahoo.ycsb.db.GoogleDatastoreClient",
+    "googlefirestore": "com.yahoo.ycsb.db.GoogleFirestoreClient",
```
busbey: Needs to be updated for the change in package name

googlefirestore/src/main/java/com/yahoo/ycsb/db/firestore/GoogleFirestoreClient.java
```diff
@@ -0,0 +1,246 @@
...
+    static final String PRIVKEYFILE = "googlefirestore.serviceAccountKey";
+    static final String DEBUG = "googlefirestore.debug";
+  }
+
+  private static final Logger LOGGER = Logger.getLogger(GoogleFirestoreClient.class);
+
+  private static Firestore fsDb;
+
+  @Override
+  public void init() throws DBException {
```
busbey: YCSB will create a DB instance per client thread and call init on each of them.

The current implementation here has unguarded access to the static member fsDb. Instead either fsDb needs to be per-instance or you need to guard the concurrent access.

Most of the other bindings do some combination of synchronization and ref counting on a static member.

charlie-lobo:
Hello @busbey I'm taking over this pull request, due to not having write access to the original branch, I am starting a new PR #1320 which should continue this.

Apologies for all the issues.

PR #1309, opened 2019-05-11
[rest] set default url.prefix to specified at readme file

dfquaresma:
As specified at Rest's [default-configuration-parameters](https://github.com/brianfrankcooper/YCSB/blob/master/rest/README.md#default-configuration-parameters), url.prefix default value is `http://127.0.0.1:80/`. Happens that the default port defined at implementation was 8080, so I set it to 80.


busbey:
Could you correct the docs instead?

Port 80 is a privileged port (one <=1024), which will make it harder to run out ofnthe box.

dfquaresma:
> Could you correct the docs instead?
> 
> Port 80 is a privileged port (one <=1024), which will make it harder to run out ofnthe box.

For sure!

PR #1310, opened 2019-05-12
Add random distribution for requests

alexwong:
See if sending requests in a random distribution has better performance by not overloading the network.


alexwong:
Gah sorry. Really not a fan of how git defaults PRs back to the main repo rather than on my fork.

PR #1286, opened 2019-03-28
[tablestore] Add binding for Alibaba Cloud's TableStore

HoboChen:
1. TableStore is a NoSQL service provided by Alibaba Cloud. [Link](https://www.alibabacloud.com/product/table-store)
2. Code, Guide of Howto is provided.

PR #1242, opened 2018-10-08
[PostgreNoSQL] Added NoSQL support for PostgreSQL.

leschekhomann:
@busbey: I hope to have addressed all suggested adaptions.
- Removed package information from testing.
- Added caching for the SQL Statements by using the Statement Class.
- Added additional information about code changes in the license part.
- Simplified test code (removed creation and deletion of test database, only use travis).

PR #1313, opened 2019-06-04
[core] add missing license headers.

busbey:
* #1249 missed headers when moving things out of Client.java
* #1242 missed header on a properties file is added
* #1286 missed header on the new pom file


busbey:
thanks!

PR #1315, opened 2019-06-14
[hbase10] Use Admin.tableExists instead of Table.getTableDescriptor t…

Apache9:
…o test whether the table exists, as the latter has been removed on master branch


Apache9:
Hit this problem when testing the HBASE-21512 branch for HBase...

https://issues.apache.org/jira/browse/HBASE-22577

busbey:
Okay if this waits for next release cycle, or do I need to pull it into the 0.16.0 RC?

Apache9:
Can be next cycle, no hurry :)

Issue #1212, opened 2018-08-13
Disable hbase‘s durability 


adally:
The YCSB readme is as the following, which is a little misleading.Because we can't use '-p durability=false' to disable durability. Instead it worked when I used '-p durability=SKIP_WAL'. The readme should be written more clearly .

durability: Whether or not writes should be appended to the WAL. Bypassing the WAL can improve throughput but data cannot be recovered in the event of a crash. The default is true.

risdenk:
Can you open a PR to address the README?

PR #1322, opened 2019-07-12
Update README.md

adally:
https://github.com/brianfrankcooper/YCSB/issues/1212

The YCSB readme is as the following, which is a little misleading.Because we can't use '-p durability=false' to disable durability. Instead it worked when I used '-p durability=SKIP_WAL'. The readme should be written more clearly.

durability: Whether or not writes should be appended to the WAL. Bypassing the WAL can improve throughput but data cannot be recovered in the event of a crash. The default is true.

### DIFF
--- a/core/src/main/java/com/yahoo/ycsb/Client.java
+++ b/core/src/main/java/com/yahoo/ycsb/Client.java
@@ -859,7 +859,7 @@ public final class Client {
       }
       if (threadcount > opcount){
         threadcount = opcount;
-        System.println("Warning: the threadcount is bigger than recordcount, the threadcount will be recordcount!");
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/yahoo/ycsb/Client.java
```diff
@@ -857,7 +857,10 @@ public static void main(String[] args) {
           opcount = Integer.parseInt(props.getProperty(RECORD_COUNT_PROPERTY, DEFAULT_RECORD_COUNT));
         }
       }
-
+      if (threadcount > opcount){
```
busbey: this looks unrelated to the README change. can you break it into a different PR please?

adally: Ok

PR #1327, opened 2019-07-25
[zookeeper]:  add the zk binding

maoling:


maoling:
ping @busbey Sir, could you please review this PR when you're free? :)

busbey:
I'm back from vacation now. I have some time booked friday to pick up YCSB stuff. Thanks for working on this!

zookeeper/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+### 1. Start ZooKeeper Server(s)
+
+### 2. Install Java and Maven
+
+### 3. Set Up YCSB
+
+Git clone YCSB and compile:
+
+    git clone http://github.com/brianfrankcooper/YCSB.git
```
busbey: please just direct folks to look at the landing page for instructions on downloading YCSB.

zookeeper/pom.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2012 - 2016 YCSB contributors. All rights reserved.
```
busbey: copyright needs update.

zookeeper/src/main/java/com/yahoo/ycsb/db/ZKClient.java
```diff
@@ -0,0 +1,188 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ * <p>
+ * ZooKeeper client binding for YCSB.
+ * <p>
+ */
+
+package com.yahoo.ycsb.db;
```
busbey: please put stuff into a binding specific package. e.g. `com.yahoo.ycsb.db.zookeeper`

zookeeper/src/main/java/com/yahoo/ycsb/db/ZKClient.java
```diff
@@ -0,0 +1,188 @@
...
+    if (sessionTimeoutString != null) {
+      sessionTimeout = Integer.parseInt(sessionTimeoutString);
+    } else {
+      sessionTimeout = DEFAULT_SESSION_TIMEOUT;
+    }
+
+    try {
+      zk = new ZooKeeper(connectString, (int) sessionTimeout, null);
+    } catch (IOException e) {
+      throw new DBException("Closing connection failed.");
```
busbey: "creating connection failed" ?

zookeeper/src/main/java/com/yahoo/ycsb/db/ZKClient.java
```diff
@@ -0,0 +1,188 @@
...
+
+  @Override
+  public Status read(String table, String key, Set<String> fields,
+                     Map<String, ByteIterator> result) {
+    String path = getPath(key);
+    try {
+      zk.getData(path, watcher, null);
+      return Status.OK;
+    } catch (KeeperException | InterruptedException e) {
+      e.printStackTrace();
```
busbey: use a logging system and log a message w/this exception instead of calling `printStackTrace`. e.g. slf4j.

zookeeper/src/main/java/com/yahoo/ycsb/db/ZKClient.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  @Override
+  public Status insert(String table, String key,
+                       Map<String, ByteIterator> values) {
+    String path = getPath(key);
+    String data = getJsonStrFromByteMap(values);
+    try {
+      zk.create(path, data.getBytes(), ZooDefs.Ids.OPEN_ACL_UNSAFE,
```
busbey: specify a charset for `getBytes` so we don't get the platform default.

zookeeper/src/main/java/com/yahoo/ycsb/db/ZKClient.java
```diff
@@ -0,0 +1,188 @@
...
+      throw new DBException("Closing connection failed.");
+    }
+  }
+
+  @Override
+  public Status read(String table, String key, Set<String> fields,
+                     Map<String, ByteIterator> result) {
+    String path = getPath(key);
+    try {
+      zk.getData(path, watcher, null);
```
busbey: need to actually parse out the data into the result set

maoling:
@busbey 
Sir, I have addressed all your comments.

zookeeper/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
```
busbey: You need an slf4j API dependency

zookeeper/src/main/java/com/yahoo/ycsb/db/zookeeper/ZKClient.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+  }
+
+  @Override
+  public Status update(String table, String key,
+                       Map<String, ByteIterator> values) {
+    String path = getPath(key);
+    String data = getJsonStrFromByteMap(values);
+    try {
+      zk.setData(path, data.getBytes(UTF_8), -1);
```
busbey: Unless I'm missing something, I think this is overwriting any fields that aren't in the update to be empty?

maoling: - zookeeper does support the simple k-v data struct which the value is plaintext, unlike the other no-sql database(e.g redis,hbase).
- For the update, the value can be updated from multi field-values to only one field-value.For the benchmark,It doesn't matter?

busbey: it matters because you aren't doing the action the benchmark is trying to measure. In the current implementation you have to pay the cost of reading and deserializing the current record in order to do an update of a subset of fields.

maoling: It's tricky.

> you have to pay the cost of reading and deserializing the current record in order to do an update of a subset of fields.

Yep.Since the read operation only reads from the memory, so the cost may be cheap.

maoling:
@busbey Sir, PTAL :)

zookeeper/README.md
busbey: Please remove the stuff about cloning and building. Just the line about the landing page will be enough.

zookeeper/src/main/java/site/ycsb/db/zookeeper/ZKClient.java
```diff
@@ -0,0 +1,213 @@
...
+      LOG.error("Error when deleting a path:{},tableName:{}", path, table, e);
+      return Status.ERROR;
+    }
+  }
+
+  @Override
+  public Status update(String table, String key,
+                       Map<String, ByteIterator> values) {
+    String path = getPath(key);
+    String data = getJsonStrFromByteMap(values);
```
busbey: This will effectively delete any fields not updated, right?

maoling: - It will pick up one field from the multi-fields map, zookeeper supports the simple k-v data struct which the value is plaintext without any complex data struct, so only pick up one field is enough for benchmark test, to avoid the cost of reading and deserializing the current record.For example:

```
update **/user4177720500626811450** from
**{"field8":">A-8W1#5z&)n2O5$6h)0.(<|' <=\"x3Dy>Qq,N5)\/~>\"`*3z 3*0Q%:!j6Dg2^? #>.92$:\"-W}:T}5>`?-j3N1%!b0?,(I%7$,4"}**
to
**{"field5":".Z9)Ew:\\q:&x3@i+I;%V}.<0#,p61 94 0Cc'Zs?&,=>$9Ri:=:%M14I+59b:=6$Tq4;4$A132(1Lk94`63v\/\\u'C=5Jy#\\c37x'"}**
```

- the UT was added.

maoling:
@busbey PTAL:)

maoling:
@symat  @scriptnull  @asya999  @granthenke  @busbey @risdenk
Any of you have time to review this patch? :)

maoling:
The CI build failure is from the unavailable of `mongodb-async-driver` from `allanbank`. Not related with my patch?

[`ERROR] Failed to execute goal on project mongodb-binding: Could not resolve dependencies for project site.ycsb:mongodb-binding:jar:0.18.0-SNAPSHOT: Failed to collect dependencies at com.allanbank:mongodb-async-driver:jar:2.0.1: Failed to read artifact descriptor for com.allanbank:mongodb-async-driver:jar:2.0.1: Could not transfer artifact com.allanbank:mongodb-async-driver:pom:2.0.1 from/to allanbank (https://www.allanbank.com/repo/): Transfer failed for https://www.allanbank.com/repo/com/allanbank/mongodb-async-driver/2.0.1/mongodb-async-driver-2.0.1.pom: Connect to www.allanbank.com:443 [www.allanbank.com/206.210.70.161] failed: Connection refused (Connection refused)`

busbey:
Not related and should be fixed now. Please rebase

maoling:
@busbey  
I'm very eager for your review. Could you please resume to review this PR when you're free?
What I can promise is that I can address your review comments in a timely manner. Hope it doesn't put you too much burden:)

zookeeper/src/main/java/site/ycsb/db/zookeeper/ZKClient.java
```diff
@@ -0,0 +1,213 @@
...
+
+  @Override
+  public Status read(String table, String key, Set<String> fields,
+                     Map<String, ByteIterator> result) {
+    String path = getPath(key);
+    try {
+      byte[] data = zk.getData(path, watcher, null);
+      if (data == null || data.length == 0) {
+        return Status.OK;
+      }
```
busbey, line 118: if there's no data for the given key we should return `Status.NOT_FOUND`

zookeeper/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>1.7.21</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <version>1.7.21</version>
+    </dependency>
```
busbey, line 65: should also add in a direct dependency on log4j to go with this slf4-log4 wrapper and the log4j config file.

zookeeper/src/main/java/site/ycsb/db/zookeeper/ZKClient.java
```diff
@@ -0,0 +1,213 @@
...
+      JSONObject jsonObject = (JSONObject)JSONValue.parse(new String(data));
```
busbey, line 119: Specify the charset here when creating a String from the data byte array.

busbey (changes requested):

zookeeper/src/main/java/site/ycsb/db/zookeeper/ZKClient.java
maoling: Done

zookeeper/src/main/java/site/ycsb/db/zookeeper/ZKClient.java
maoling: Done

zookeeper/pom.xml
maoling: I'm sorry for not familar with the log stuff. I append a log4j dependency and we already have a  log4j config file under `resource ` directory

zookeeper/src/main/java/site/ycsb/db/zookeeper/ZKClient.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    String watchFlagString = props.getProperty(WATCH_FLAG);
+    if (watchFlagString == null || "false".equals(watchFlagString)) {
+      watcher = null;
+    } else if ("true".equals(watchFlagString)) {
+      watcher = new SimpleWatcher();
+    } else {
+      throw new IllegalArgumentException("zookeeper.watchFlag only has two options:true or false");
+    }
```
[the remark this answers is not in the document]
maoling: done

zookeeper/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+    # cluster test
+    ./bin/ycsb run zookeeper -P workloads/workloadb -p zookeeper.connectString=192.168.10.43:2181,192.168.10.45:2181,192.168.10.27:2181/benchmark
+
+    # test leader's read/write performance by setting zookeeper.connectString to leader's(192.168.10.43:2181)
+    ./bin/ycsb run zookeeper -P workloads/workloadb -p zookeeper.connectString=192.168.10.43:2181/benchmark
+
+    # test for big znode. Notice:jute.maxbuffer should also be set the same value in the server side.
+    ./bin/ycsb run zookeeper -jvm-args="-Djute.maxbuffer=4194304" -s -P workloads/workloadc -p zookeeper.connectString=127.0.0.1:2181/benchmark
+
```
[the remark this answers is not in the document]
maoling: Add more detail about the default limit

pom.xml
```diff
@@ -151,6 +151,7 @@ LICENSE file.
     <voldemort.version>0.81</voldemort.version>
     <tablestore.version>4.8.0</tablestore.version>
     <voltdb.version>9.1.1</voltdb.version>
+    <zookeeper.version>3.5.5</zookeeper.version>
```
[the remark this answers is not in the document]
maoling: Upgrade to lastest version:`3.6.2`. AFAIK, ZK has great compatibility between all the client and server version.

maoling:
Eeeee, the unit case has failed, I'll take a look tomorrow morning

zookeeper/README.md
```diff
@@ -0,0 +1,94 @@
...
+    # YCSB workloadb is the most suitable workload for read-heavy workload for the ZooKeeper in the real world.
+
+    # -p fieldlength, test the length of value/data-content took effect on performance
+    ./bin/ycsb run zookeeper -s -P workloads/workloadb -p zookeeper.connectString=127.0.0.1:2181/benchmark -p fieldlength=1000
+
+    # A Notice for the property:-p fieldcount
+    zookeeper does support the simple k-v data struct which the value is plaintext without any complex data struct,
+    so only pick up one field is enough for benchmark test, to avoid the cost of reading and deserializing the current record
+    in order to do an update of a subset of fields.So fieldcount is strictly constrained to 1,other value(like the below example) is invalid.
+    ./bin/ycsb run zookeeper -s -P workloads/workloadb -p zookeeper.connectString=127.0.0.1:2181/benchmark -p fieldcount=20
```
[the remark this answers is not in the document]
maoling: - zookeeper does support the simple k-v data format and its value is simple without any complex data structure. The `setData ` api doesn't suport to update a subset of the znode data content. I cannot to read the znode with multi-fields, generate the new view of  multi-fields, then update the record in the method `update`  that will cause the measure of write opertion inaccurate(mixed with a read opertion).
- I constrained `fieldcount `to 1 because I want to insert/read/update a record with one field(In fact, for the zk benchmark ,we don't care about the `fieldcount `, we care about how the data length takes effect on the performance). If allowing for a configurable number of fields, I'll insert/read with 10 fields and update that znode to 1 field. The data volume of read and write is unequal and a record will shrink. Is it OK in the context of YCSB framework?
- I'm glad to listen to your insight :)

maoling:
I revert zookeeper client version from `3.6.2 `to `3.5.8`(the latest version of branch-3.5.X), because when using the version 3.6.X and test with multi-thread([1]), the output of console always prints a little weird and uncomfortable logs([2]) although they may don't fail anything(I dig into the source codes and don't know what's the matter)

```
[1] 
./bin/ycsb run zookeeper -threads 500 -P workloads/workloadc -p zookeeper.connectString="xxxxxx"
```

```
[2]
2020-12-01 13:34:50 WARN  ClientCnxn:1279 - An exception was thrown while closing send thread for session 0x200088a8e080057.
EndOfStreamException: Unable to read additional data from server sessionid 0x200088a8e080057, likely server has closed socket
  at org.apache.zookeeper.ClientCnxnSocketNIO.doIO(ClientCnxnSocketNIO.java:77)
  at org.apache.zookeeper.ClientCnxnSocketNIO.doTransport(ClientCnxnSocketNIO.java:350)
  at org.apache.zookeeper.ClientCnxn$SendThread.run(ClientCnxn.java:1275)
```

zookeeper/README.md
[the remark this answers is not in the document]
busbey: > I cannot to read the znode with multi-fields, generate the new view of multi-fields, then update the record in the method update that will cause the measure of write opertion inaccurate(mixed with a read opertion).

Having to do the read in order to do the correct write is exactly what the benchmark in this case should reflect for an update. The update benchmark is specifically meant to show the difference between when an underlying datastore can do partial updates to a record and when it needs to read the complete record to update a field.

In order to avoid the read, either ZK would have to have some notion of a complex data type or you would have to structure record storage for this application so that a given znode was only the content of 1 field.

Given the current binding implementation here, where a znode corresponds to a single record, you will need to read the underlying znode before updating the set of fields passed for the update call.

busbey (changes requested):

maoling:
@busbey  Thanks for your patient explanation. I do a read operation before `setData` to meet the YCSB's update semantics(I found the RocksDB's binding for update also used the same approach). PTAL:)

busbey (approved):

maoling:
@busbey Sir, thanks for your review work:)

PR #1328, opened 2019-07-25
Fix the markdown syntax

wurikiji:


busbey:
Looks fine, could you rebase to current master?

Also the commit message should be "[jdbc] fix some markdown syntax in README"

busbey:
I see. this has a conflict because it got obviated by #1335.

Thanks for the effort all the same! sorry for the lag.

PR #1336, opened 2019-08-02
Update ycsb python wrapper

Sraw:
This wrapper has been updated to be Python2-Python3 compatible.
And the Python2.6 supported has been removed as it is almost impossible to be used.

### DIFF
--- a/bin/ycsb
+++ b/bin/ycsb
@@ -15,21 +15,16 @@
 # permissions and limitations under the License. See accompanying
 # LICENSE file.
 #
+from __future__ import print_function, unicode_literals
 
+import argparse
 import errno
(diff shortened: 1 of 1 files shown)


Sraw:
I've never gone into that run from source branch honestly. It should work for both two branches now. Tested on my laptop.

busbey:
Thanks for the update!

busbey:
ah geez. it looks like this closed automatically when I made the 0.16.0 release and I totally didn't notice. my apologies!

I need to figure out which of the three attempts at python 3 support we adopt and then move forward with one. I'll keep this one in the running.

busbey:
favoring #1421 instead

PR #1342, opened 2019-08-15
add fdb record layer

Sraw:
Add support for FoundationDB Record Layer.

### DIFF
--- a/fdbrecord/src/main/java/com/yahoo/ycsb/db/fdbrecord/FDBRecordClient.java
+++ b/fdbrecord/src/main/java/com/yahoo/ycsb/db/fdbrecord/FDBRecordClient.java
@@ -1,0 +1,309 @@
+package com.yahoo.ycsb.db.fdbrecord;
+
+import com.apple.foundationdb.record.*;
+import com.apple.foundationdb.record.metadata.Key;
+import com.apple.foundationdb.record.provider.foundationdb.*;
+import com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpace;
+import com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpaceDirectory;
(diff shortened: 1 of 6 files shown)


fdbrecord/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+See the next section for the list of configuration parameters for FoundationDB.
+
+## FoundationDB Configuration Parameters
+
+* ```fdb.clusterfile``` - The cluster file path.
+  * Default: ```./fdb.cluster```
+* ```fdb.keyspacepath``` - The key space path.
+  * Default: ```YCSB```
+* ```fdb.batchsize``` - The number of rows to be batched before commit.
```
busbey: Specify that this defualt size means inserts won't be batched.

fdbrecord/src/main/java/com/yahoo/ycsb/db/fdbrecord/FDBRecordClient.java
```diff
@@ -0,0 +1,309 @@
...
+    String rowKey = (String) record.getField(msgDesc.findFieldByName("YCSB"));
+    logger.debug("insert key = {}", rowKey);
+    try {
+      batchRecords.add(record);
+      batchCount++;
+      if (batchSize == 0 || batchSize == batchCount) {
+        batchInsert();
+        batchCount = 0;
+      }
+      return Status.OK;
```
busbey: When batched but not executed this should be BATCHED_OK.

Sraw: Actually I copy this logic from `foundationdb-binding` driver. But I think it doesn't fully follow the interface. Which means this is not the batch behavior defined in the interface, but a kind of new batch behavior.
In this case, I cannot change it to `BATCHED_OK`, it will ruin the normal inserting behavior.
Or maybe I wrongly understand the interface?

PR #1350, opened 2019-09-05
[jdbc] Add mssql support for scan operation

KowsarAtz:
Fixing the syntax issue SQL Server faced during scan operation.
( Microsoft SQL Server doesn't support LIMIT statement, so I added a condition to use TOP instead when we are using jdbc-binding to benchmark SQL Server scan operation. )

### DIFF
--- a/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
+++ b/jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010 - 2016 Yahoo! Inc., 2016 YCSB contributors. All rights reserved.
+ * Copyright (c) 2010 - 2019 Yahoo! Inc., 2019 YCSB contributors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/com/yahoo/ycsb/db/flavors/DBFlavor.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2016 YCSB contributors. All rights reserved.
+ * Copyright (c) 2019 YCSB contributors. All rights reserved.
```
busbey: This should be 2016, 2019 to reflect work from both periods.

jdbc/src/main/java/com/yahoo/ycsb/db/JdbcDBClient.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010 - 2016 Yahoo! Inc., 2016 YCSB contributors. All rights reserved.
+ * Copyright (c) 2010 - 2019 Yahoo! Inc., 2019 YCSB contributors. All rights reserved.
```
busbey: This should remain 2010-2016 Yahoo and then 2016, 2019 ycsb contributors

jdbc/src/main/java/com/yahoo/ycsb/db/flavors/DefaultDBFlavor.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2016 YCSB contributors. All rights reserved.
+ * Copyright (c) 2019 YCSB contributors. All rights reserved.
```
busbey: This should be 2016, 2019 to reflect work from both periods.

KowsarAtz: Thank you for your review! I've made the changes as requested.

Issue #1265, opened 2018-11-29
Release 0.16.0


busbey:
let's try to get a release out before the US winter holidays. I'll update here with changes after I make the release branch.

busbey:
Fyi, I'll be picking this up again this weekend.

busbey:
#1313 is a blocker

busbey:
I created the staging branch for 0.16.0. Here's the change summary, by my reading, since 0.15.0.

# core changes
* Allow configuration of field names via fieldnameprefix property (#117)
* JDK 8 or 11 should be used, we no longer test JDK9 or JDK10 (#1253)
* Improvements to some internal byte iterators (notably Strings) (#1112)
* Correct configuration of exponential distribution in the Core workloads (#1260)
# added bindings
* ElasticSearch 5 REST client (#1231)
* Azure Cosmos DB (#1264 #1298 #1300)
* Postgres NoSQL (#1242)
* Apache Crail (incubating) (#1266)
* Table Store from Alibaba Cloud (#1286)
* Toshiba GridDB (#1258)
# updated bindings
* Compat for MongoDB 4 (#1217)
* JDBC adds multi-row inserts (#1220)
* OrientDB updated to version 2.2.37 (#1256)
* memcached fixes a logic error where individual fields would not get checked (#1153)
* Apache Cassandra can optionally use SSL to connect to the cluster (#1294)
* Apache HBase add documentation for using client side buffering (#1283)
* Apache Ignite version upgraded (#1270)
* Google Cloud Bigtable client upgraded (#1293)
* Cloud Spanner client upgraded (#1293)
* AsyncHBase client upgrade (#1255)
# removed bindings
* Azure Document DB has been removed (#1295)

busbey:
Release candidate 1 is up:
https://github.com/brianfrankcooper/YCSB/releases/tag/0.16.0-RC1

As a reminder, testing a given binding should at a minimum run through a `load` and a `run` phase. Ideally, one should go through the recommended sequence of workload steps from the ["Running the workloads" section of the Core Workloads documentation](https://github.com/brianfrankcooper/YCSB/wiki/Core-Workloads#running-the-workloads).

busbey:
known issue for mongodb-async #1314

busbey:
checked field name prefix configuration on OS X using the basic binding. worked fine for `bin/ycsb` but looks like #908 is still an issue.

```
[...]

PR #1352, opened 2019-09-08
[build] update group id to site.ycsb

busbey:
This moves all group ids in the project to be "site.ycsb" from "com.yahoo.ycsb." This is necessary if we want to publish artifacts to Maven Central since as a project we're no longer associated with Yahoo! and they control who publishes artifacts within "com.yahoo".

Tested locally with a clean build and running the basicdb to load and run workload a both in the source checkout and the distribution tarball.

Note that this doesn't change java packages (will be a different PR) and it doesn't address the regression on #908 that came up in the 0.16.0 testing on #1265 .

### DIFF
--- a/accumulo1.6/README.md
+++ b/accumulo1.6/README.md
@@ -34,7 +34,7 @@ Git clone YCSB and compile:
 
     git clone http://github.com/brianfrankcooper/YCSB.git
     cd YCSB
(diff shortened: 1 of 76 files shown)


busbey:
rebased the branch to current master and updated it for the Volt DB binding landing.

lewismc:
@busbey `mvn clean install` is successful. +1

PR #1353, opened 2019-09-10
[rocksdb] add rocksdb.optionsfile property

twblamer:
This adds support to run RocksDB with a custom [options file](https://github.com/facebook/rocksdb/wiki/RocksDB-Options-File).

This also updates the RocksDB Maven artifact version to the latest (6.2.2).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,7 +102,7 @@ LICENSE file.
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <redis.version>2.9.0</redis.version>
     <riak.version>2.0.5</riak.version>
-    <rocksdb.version>5.11.3</rocksdb.version>
+    <rocksdb.version>6.2.2</rocksdb.version>
     <s3.version>1.10.20</s3.version>
     <solr.version>5.5.3</solr.version>
     <solr6.version>6.4.1</solr6.version>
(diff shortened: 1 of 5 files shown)


busbey:
If I have a data file for rocksdb written by version 5, will it still work when I run with the version 6 library?

twblamer:
From looking at the Rocks [compatibility strategy](https://github.com/facebook/rocksdb/wiki/RocksDB-Compatibility-Between-Different-Releases) and the [release log](https://github.com/facebook/rocksdb/releases), I don't anticipate any breaking issues moving from 5.11.3 to 6.2.2. Especially since YCSB is currently initializing the RocksDB options using the `optimizeForLevelCompaction()` helper function and will continue doing so by default, so any DB initialized by YCSB shouldn't depend on any exotic options that got deprecated.

I just tested it myself to confirm (loaded a YCSB DB with 100K records using Rocks 5.11.2, rebuilt YCSB with Rocks 6.2.2 and ran a run workload with no issues).

However, my knowledge of RocksDB internals is minimal, so It would be could if @adamretter or another Rocks maintainer could chime in.

adamretter:
@twblamer I think the version upgrade should be fine.

twblamer:
Fixed a typo and squashed commits.

twblamer:
@adamretter Does the PR look OK to you?

adamretter:
@twblamer It would be good to see a test case.

twblamer:
@adamretter I've added a test, please have a look.

PR #1354, opened 2019-09-19
Move all packages from com.yahoo.ycsb to site.ycsb.

busbey:
in addition to #1352, this moves all the packages for our classes to be in `site.ycsb` instead of `com.yahoo.ycsb`.

Tested locally with a clean build and running the basicdb to load and run workload a both in the source checkout and the distribution tarball.


lewismc:
@busbey `mvn clean install` is successful +1

PR #1360, opened 2019-10-02
[MongoDB] latest driver, allow SRV URI format, readme fix

asya999:
upgraded MongoDB to latest driver
allow SRV style mongodb URI
fix readme example for building just mongodb-binding

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,4 +77,4 @@ To build the full distribution, with all database bindings:
 
 To build a single database binding:
 
-    mvn -pl mongodb-binding -am clean package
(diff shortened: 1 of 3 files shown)


busbey:
Thanks for this!  one quick question, does the SRV URI format work for both the normal and the async client? It's okay if it doesn't work for the Async client, but we should warn folks in the README if it's a known problem.

asya999:
Should work with any client as long as connecting to 3.6 server or later.

busbey:
thanks for this update! sorry for the lag.

PR #1366, opened 2019-10-21
[accumulo] remove EOM Accumulo versions (1.6, 1.7) and rename 1.8 binding to 1.9.

busbey:
Accumulo 1.9 is a continuation of the 1.8 line with an API addition, so we've been using Accumulo 1.9 client libraries in the "accumulo-1.8" binding. However, since Accumulo's website positions Accumulo 1.9 as a current release and not Accumulo 1.8 we should update the binding to reflect that version.


busbey:
thanks for the review!

PR #1372, opened 2019-11-12
Npe-fix

JulyChen728:
I updated the JdbcDBClient.java .


jdbc/src/main/java/site/ycsb/db/JdbcDBClient.java
```diff
@@ -195,6 +193,9 @@ public void init() throws DBException {
     this.batchUpdates = getBoolProperty(props, JDBC_BATCH_UPDATES, false);
 
     try {
+      if (driver.contains("sqlserver")) {
+        sqlserver = true;
+      }
       if (driver != null) {
```
busbey: I think we need to move into this null check

busbey:
Thanks for the fix!

PR #1373, opened 2019-11-12
[jdbc] Add support for SQL:2008 FETCH FIRST n ROWS ONLY for Scans

FranckPachot:
Currently, the JDBC interface implements the Scan with:

- SELECT TOP for SQL Server
- ORDER BY ... LIMIT for others

Today, most databases use the ORDER BY ... FETCH FIRST ... ROWS which is the SQL standard since 2008. 
See [Markus Winand matrix](https://www.slideshare.net/MarkusWinand/modern-sql/120)

In this PR I kept the "sqlserver" boolean but renamed it "sqlserverScans" as it is used only for scans. And I added a "sqlansiScans" for the FETCH FIRST syntax.

For compatibility, I kept it at false, and enable it only when the driver contains "oracle" or "postgres" as these are the common databases supporting the SQL ANSI syntax. But actually only SQLite, MySQL and MariaDB need to use this LIMIT syntax in current versions.

In the future, it is probably better to check the product and version once connected. Or maybe try the 3 syntaxes at connection by parsing a simple query and chose the first one with no error.

### DIFF
--- a/jdbc/src/main/java/site/ycsb/db/JdbcDBClient.java
+++ b/jdbc/src/main/java/site/ycsb/db/JdbcDBClient.java
@@ -82,10 +82,11 @@ public class JdbcDBClient extends DB {
   /** The field name prefix in the table. */
   public static final String COLUMN_PREFIX = "FIELD";
 
-  /** Use SQL:2008 FETCH FIRST for Scan. */
-  private boolean ansiFetchFirst = false;
-  
-  private boolean sqlserver = false;
+  /** SQL:2008 standard: FETCH FIRST n ROWS after the ORDER BY. */
+  private boolean sqlansiScans = false;
+  /** SQL Server before 2012: TOP n after the SELECT. */
+  private boolean sqlserverScans = false;
+
   private List<Connection> conns;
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/site/ycsb/db/JdbcDBClient.java
```diff
@@ -184,8 +188,23 @@ public void init() throws DBException {
     String passwd = props.getProperty(CONNECTION_PASSWD, DEFAULT_PROP);
     String driver = props.getProperty(DRIVER_CLASS);
 
+  /* The SQL Syntax for Scan depends on the DB engine
```
busbey: The if block for sqlserver recently moved down within our null check for the driver string. These should all be down there.


This is probably the merge conflict too.

FranckPachot: Ok I rebased and committed the merge.  I'm not sure if have something else to do on the PR or if that's ok.

busbey:
Thanks! I'll ping you to test this out once it's in a release candidate.

PR #1376, opened 2019-12-01
Read ResultSet was not iterated correctly.

rogerioacp:
The first ResultSet.next() call to test if the ResultSet was empty moved the cursor without adding the resulting row to the YCSB result callback.


busbey (approved):
would be nice to add a test that catches this since our existing unit tests didn't flag this up.

busbey:
thanks for fixing this!

Issue #1382, opened 2020-01-06
not possible to use zero operation count


symat:
Previously, if you defined `maxexecutiontime` and not defined `operationcount` parameter (or set `operationcount=0`), then you were able to run YCSB for a given period of time and measure the number of operations you get during the load test.

Since https://github.com/brianfrankcooper/YCSB/pull/1323 was merged, it it not possible to use `operationcount=0`, as the thread count will be set to 0 in this case and no client thread will be created:

https://github.com/brianfrankcooper/YCSB/blob/4b19340e3bab5e4c88eda75ad56e83dc4d5cc503/core/src/main/java/site/ycsb/Client.java#L421-L424

The solution would be to change the is statement to: `if (threadcount > opcount && opcount > 0)`

What do you think?
(should I create a PR?)

busbey:
Yes please, that solution sounds great.

PR #1384, opened 2020-01-08
[core] make it possible to use zero operation count again. (#1382)

symat:
Solving https://github.com/brianfrankcooper/YCSB/issues/1382

Previously, if you defined maxexecutiontime and not defined operationcount
parameter (or set operationcount=0), then you were able to run YCSB for a
given period of time and measure the number of operations you get during
the load test.

Since #1323 was merged, it it not possible to use operationcount=0, as
the thread count will be set to 0 in this case and no client thread will
be created. This PR makes it possible to use operationcount=0 again.

### DIFF
--- a/core/src/main/java/site/ycsb/Client.java
+++ b/core/src/main/java/site/ycsb/Client.java
@@ -418,7 +418,7 @@ public final class Client {
           opcount = Integer.parseInt(props.getProperty(RECORD_COUNT_PROPERTY, DEFAULT_RECORD_COUNT));
         }
       }
-      if (threadcount > opcount){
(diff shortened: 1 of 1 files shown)


symat:
I don't think this build error is related to anything I did... is it possible to re-trigger the CI build?

busbey:
Yes, but I need to make sure the failure was transient. It showed up in a bunch of places.

busbey:
okay it's not transient, but just happens on Travis. they must be futzing about with maven mirrors again. :(

In any case I ran through the tests locally and they look fine.

PR #1389, opened 2020-01-23
[ignite] Updates ignite to latest version

scriptnull:
Reference: https://ignite.apache.org/download.cgi 

Without this patch, benchmarking fails for the latest version of Apache Ignite.


busbey:
travis failure is #1390 . ran tests locally and they're fine.

PR #1395, opened 2020-02-03
[readme] Document the use of `-Psource-run` flag

granthenke:
Adds a quick note to the main readme about the `-Psource-run` flag.

I also added the flag to the Kudu readme command.


busbey:
That flag is supposed to be set automatically when using `bin/ycsb.sh` or `bin/ycsb.bat` and shouldn't have to be set when building yourself. (though I am not sure why they're not just calling the dependency plugin to build a classpath as the python loader does)

can you describe the failure mode you saw?

PR #1396, opened 2020-02-03
[hbase] update HBase bindings for eom

### LINKED ISSUE #1172: [hbase098] README instructions should refer to 'hbase098' binding and not 'hbase'
the current README gives example commands for hte 'hbase' binding, which doesn't exist.

### LINKED ISSUE #1173: [hbase20] no slf4j implementation; hbase client doesn't log
running through testing the `hbase20` binding, there's no logs from the client library (as there are in the earlier bindings).

```
Loading workload...
Starting test.
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Looks like the hbase 1.2 and hbase 1.4 shaded jars incorrectly included a dependency on the `slf4j-log4j12` bridge. so to have the same logging on the hbase 2.0 client we'll need to add it.

busbey:
* remove 0.98, 1.0, 1.2, and 2.0 bindings
* add 2.2 binding
* incorporate README from 0.98 binding into current bindings
* incorporate README on bigtable testing from 1.0 binding into 1.4 binding
* incorporate implementation from 1.0 client into current bindings
* updated asynchbase binding to include parts of removed bindings it referenced
* update 1.4 and 2.2 to current releases
* use shaded client test for all hbase bindings.
* make hbase bindings consistently use log4j
* fixes #1173
* fixes #1172


busbey:
ugh. the maprdb stuff _also_ had a reference to the hbase 1.0 binding implementation? I'll chase that down.

hbase22/README.md
```diff
@@ -15,25 +15,17 @@ permissions and limitations under the License. See accompanying
 LICENSE file.
 -->
 
-# HBase (0.98.x) Driver for YCSB
-This driver is a binding for the YCSB facilities to operate against a HBase 0.98.x Server cluster.
-To run against an HBase >= 1.0 cluster, use the `hbase10` binding.
+# HBase (2.1+) Driver for YCSB
```
joshelser: If this works with HBase 2.1.x, why is the module called hbase22? Prior convention?

busbey: I only wanted to include the versions with stable pointers pointing at them. When I started the PR back in October though there was no stable branch-2, so it was just the latest which was 2.1.

I can update this text to say 2.2+, maybe? In theory `hbase14` is really the "HBase 1.y API" and `hbase22` is really "HBase 2.y API". AFAIK the only minor release specific thing either module makes use of is the shaded testing util artifacts.

joshelser:
Gotcha. I think your phrasing of "really 1.x and 2.x artifacts" is the most meaningful for users. The numbers in the path can just be things we use :)

busbey:
Okay I switched both remaining HBase bindings to refer to "hbase 1.y" and "hbase 2.y" instead of specific minor releases. how's that?

joshelser:
:+1: sounds great!
Ship it.

asynchbase/README.md
```diff
@@ -21,9 +21,24 @@ This driver provides a YCSB workload binding for Apache HBase using an alternati
...
+### 2. Set up YCSB
+
+Download the [latest YCSB](https://github.com/brianfrankcooper/YCSB/releases/latest) file. Follow the instructions.
+
+### 3. Create a HBase table for testing
+
+For best results, use the pre-splitting strategy recommended in [HBASE-4163](https://issues.apache.org/jira/browse/HBASE-4163):
+
+```
+hbase(main):001:0> n_splits = 200 # HBase recommends (10 * number of regionservers)
```
ndimiduk: this is gross. we have the split algorithm classes... assuming they're now available on all release branches, we should promote those instead.

hbase1/README.md
```diff
@@ -15,17 +15,66 @@ permissions and limitations under the License. See accompanying
...
+Then, you can run the workload:
+
+```
+bin/ycsb run hbase1 -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family
+```
+
+Please see the general instructions in the `doc` folder if you are not sure how it all works. You can apply additional properties (as seen in the next section) like this:
+
+```
+bin/ycsb run hbase1 -P workloads/workloada -cp /HBASE-HOME-DIR/conf -p table=usertable -p columnfamily=family -p clientbuffering=true
```
ndimiduk: is there a way for the user to specify a specific version of the hbase-1.y client? From the poms, seems it can be specified at build time.

hbase2/README.md
```diff
@@ -15,25 +15,17 @@ permissions and limitations under the License. See accompanying
 LICENSE file.
 -->
 
-# HBase (0.98.x) Driver for YCSB
-This driver is a binding for the YCSB facilities to operate against a HBase 0.98.x Server cluster.
-To run against an HBase >= 1.0 cluster, use the `hbase10` binding.
+# HBase (2.y) Driver for YCSB
```
ndimiduk: Is there a reason not to consolidate the 3 different readmes into a single file?

asynchbase/README.md
busbey: I would love to have a simpler way to presplit. Can you make a PR with specifics?

hbase1/README.md
busbey: you can build against a specific different version if you like yes. after this PR it's by specifying `hbase1.version` or `hbase2.version` respectively

hbase2/README.md
busbey: the way packaging works, we create a tarball of binaries for each binding and that packaging gets the README associated with that binding. We could restructure the hbase1, hbase2, and asynchbase modules to try to change this, but that's going to be a bunch of work.

asynchbase/README.md
ndimiduk: The class I was thinking of is `o.a.h.h.util.RegionSplitter$SplitAlgorithm`. `RegionSplitter` is marked as IA.Private, but `SplitAlgorithm` is consumed from public interfaces of the likes of `TableSnapshotInputFormat` and others. It does not seem to have an existing algorithm that is sympathetic to YCSB, but I suppose one could drop such an implementation into the class path of the target cluster...

PR #1403, opened 2020-02-04
[arangodb] update to latest arangodb java driver.

busbey:
AFAICT there's no easy way to add unit tests, so this still needs to be checked against an ArangoDB instance.

According to the [ArangoDB Drivers Website](https://www.arangodb.com/arangodb-drivers/) this version of the java driver should work with server versions  3.6, 3.5, 3.4, 3.3, 3.2, 3.1, 3.0. AFAICT that's all versions they current support running.


busbey:
@smallsmallc or @mvollmary could you give this a quick review?

PR #1404, opened 2020-02-04
[couchbase] move all of couchbase to the latest 2.x SDK.

busbey:
According to Couchbase docs, this should cover all versions of Couchbase that are not EOM.

https://docs.couchbase.com/java-sdk/3.0/project-docs/compatibility.html#couchbase-versionsdk-version-matrix
https://www.couchbase.com/support-policy

the 2.7 SDK is EOM in Dec 2020 so we'll need to update to 3.0 before then, but that looks like it requires code changes.


busbey:
It looks like EOM for the 2.7 SDK got pushed out to March 2021. I'd like to get this merged before doing the work to get us to the 3.0 SDK.

@daschl, @ingenthr could either of you do a quick review? suggestions for someone else if not?

PR #1407, opened 2020-02-19
[WIP] TimeSeriesWorkload enhancements + New TS Workloads + Cassandra Adapater for TimeSeriesWorkload

smartygus:
This is still a work in progress with still stuff to be done to get it to a mergeable standard, but I am opening a pull request in its current state to get some feedback from the maintainers on the direction and whether anything major would need to be changed to get it merged.

The three main things this Pull Request includes are as follows:

- Updated version of `TimeSeriesWorkload` to broaden the range of workloads that are possible
- A new set of time series workloads based on the capabilities of the updated `TimeSeriesWorkload` class
- A Cassandra client adapter specifically for running `TimeSeriesWorkload` workloads.

Here's the more detailed explanation of these things:

This pull request makes quite a few changes/updates to the `TimeSeriesWorkload` class, to provide support for richer and more flexible time series workloads. The full scope of the changes is probably best seen in the comments of the updated `tsworkload_template`, where all the new configuration properties are documented. But in essence it adds/changes the following:

- Per-transaction type configuration of most parameters like starting/ending timestamp, group by parameters and so-on
- The possibility to specify multiple, indexed, SCAN queries in one workload, each with their own set of parameters, while still retaining the ability to specify a single SCAN query without any indexing
- Additional option for the dividing up of the write workload over the available threads: this can now by done in groups of individual time series (a time series = metric name + set of tag key-value pairs), so that even workloads designed around a single metric can be parallelised for maximum loading potential from a single YCSB client

I also made some tweaks to `TimeseriesDB` ( introduced in #1095 ) along the way, to fix some bugs I found and support the updated `TimeSeriesWorkload` class.

My hope would be that this could take the `TimeSeriesWorkload` from experimental status to actually being worthy of being used to benchmark time series data stores.

In addition to these changes, I've developed a set of time series workloads based on these richer capabilities and included them in this pull request. These were developed as part of a deep-dive into time series data, its analysis, and workloads that I did over the past few months as part of my bachelor thesis. The workloads are divided up into two groups based on scenario: Smart Metering, and DevOps Monitoring. Perhaps these could be included as the new standard workloads for `TimeSeriesWorkload`?

Lastly, I've created a new client adapter for Cassandra (and by extension ScyllaDB) specifically designed to run the workloads from `TimeSeriesWorkload`. I've used this to carry out numerous experiments and compare the performance of Cassandra and ScyllaDB, and it works reasonably well, but I would characterise it as an "initial implementation", not least because it has been designed with a particular schema, and Cassandra doesn't have a standardised schema for time series data. Depending on one's requirements, one type of schema might perform better than another. I chose a flexible schema that would support all of the various options available in `TimeSeriesWorkload`, but at the moment the tradeoff is that all the aggregation is done client-side and some queries require the use of `ALLOW FILTERING`.

I created this primarily because Cassandra (and ScyllaDB) are quite often promoted as being a good fit for time series data, but the only benchmarks out there that I could find were from vendors of other time series databases looking to compare their own product to Cassandra (among others). I'm automatically somewhat skeptical of vendor-published benchmarks, and thought it would be great to have a way to benchmark Cassandra/ScyllaDB versus other time series databases that's independent of any vendor. The initial work done on `TimeSeriesWorkload` and `TimeseriesDB` struck me as a good foundation for building out support for rich and flexible time series workloads in YCSB, and building a client adapter for time series, so I took these and ran with them.

That said, at the moment, there are still no other adapters in YCSB for time series databases - many were planned to be brought over from YCSB-TS in #1068, but work there has stalled a bit. Nevertheless, there was no adapter for Cassandra in YCSB-TS and I think its a very interesting use case for a lot of people, because Cassandra (and by extension ScyllaDB) is not purpose-built for time series data, but as I said is often used for this purpose. Maybe 

It'd be great to have some feedback!

There's still some stuff to do. Here's a list of the things I can think of at the moment:

- [ ] Fix tests for `TimeSeriesWorkload`
- [ ] Write some tests for `CassandraCQLClientTS`
- [ ] Return query results from `CassandraCQLClientTS`


busbey:
this is an excellent development on the TS benchmarking stuff. I'll need to block out some time to dig in, but thanks for taking this first step!

smartygus:
Yeah no worries, there’s quite a bit to digest there, so take your time :)
Thanks for reviewing!

Issue #1426, opened 2020-05-28
Enable reading all fields of a table by column name


justinuang:
I have observed that for certain databases (google bigtable), it can be faster to read all the columns by passing in all the field names, vs forcing the client to do a wildcard filter .* on the columns. This is especially true when there is only 1 column, where I have observed an also 50% increase in performance when reading by column name. One way to force the client to pass in all the names is to enable dataintegrity:

https://github.com/brianfrankcooper/YCSB/blob/cd1589ce6f5abf96e17aa8ab80c78a4348fdf29a/core/src/main/java/site/ycsb/workloads/CoreWorkload.java#L718

However, dataintegrity incurs the CPU cost of verifyRow, which I would rather not have. Can we create a new flag called readallfieldsbyname? I'm happy to create the PR for it.

busbey:
Sure that sounds reasonable. Default to doing things by the explicit list of all expected fields I'd say.

justinuang:
I linked a PR with what I was thinking. Regarding

> Default to doing things by the explicit list of all expected fields I'd say.

Do you mean that readallfieldsbyname should default to true? I'm a bit wary of that because I don't want to change the benchmark behavior for existing ones. For example, it is slower to readallfieldsbyname if there are many columns in googlebigtable, but faster if there are only a few columns.

justinuang:
Friendly ping =)

busbey:
i'm in the middle of a household move, gonna be a bit for me. If any other maintainers have a chance I don't need to personally be the one to sign off.

justinuang:
Ah I understand, yea I'm moving too. Good luck!

busbey:
fixed by #1427

PR #1427, opened 2020-05-28
[core] Support readallfieldsbyname

justinuang:
>    Some clients may be more efficient when you pass in the column name,
>    instead of forcing the client to do a wildcard scan. This is the case
>    with googlebigtable, when there is only 1-3 columns.

Fixes https://github.com/brianfrankcooper/YCSB/issues/1426


justinuang:
Looks like flake. Force pushing to retrigger.

busbey:
merged above. thanks!

PR #1440, opened 2020-07-05
[dynamodb] Improve endpoint and region config

ahachete:
* Upgrade AWS SDK from version 1.10.48 to 1.11.812.

* Introduce a new configuration parameter, dynamodb.region, which
  represents a valid AWS region code (see
  https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions
  for example). If specifying the region, it is not necessary to specify
  the dynamodb.endpoint parameter.

* Introduce support for non-standard endpoints. Before this commit,
  setting the endpoint to a non-standard one, like a proxy, but
  ultimately connecting to a region different from 'us-east-1' would
  lead to a credential errors due to bad signature:

  ERROR site.ycsb.db.DynamoDBClient
  -com.amazonaws.AmazonServiceException: Credential should be scoped to
  a valid region, not 'us-east-1'.  (Service: AmazonDynamoDBv2; Status
  Code: 400; Error Code: InvalidSignatureException;

  With this commit, if using a proxy as an endpoint, by setting also the
  region via dynamodb.region, it will work with no error.

* Set TCP Keep-Alive to true. Even it makes sense to be used by YCSB, it
  apparently doesn't improve performance notably given the connection
  reuse that the DynamoDB client is doing.

* Update the example config file with the endpoint and region parameters
  and behavior.

### DIFF
--- a/dynamodb/src/main/java/site/ycsb/db/DynamoDBClient.java
+++ b/dynamodb/src/main/java/site/ycsb/db/DynamoDBClient.java
@@ -19,13 +19,15 @@ package site.ycsb.db;
 import com.amazonaws.AmazonClientException;
 import com.amazonaws.AmazonServiceException;
 import com.amazonaws.ClientConfiguration;
-import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.auth.PropertiesCredentials;
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient;
+import com.amazonaws.client.builder.AwsClientBuilder;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder;
 import com.amazonaws.services.dynamodbv2.model.*;
-import site.ycsb.*;
 import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
+import site.ycsb.*;
 
 import java.io.File;
 import java.util.HashMap;
@@ -35,7 +37,7 @@ import java.util.Set;
 import java.util.Vector;
 
 /**
- * DynamoDB v1.10.48 client for YCSB.
+ * DynamoDB client for YCSB.
  */
 
 public class DynamoDBClient extends DB {
@@ -52,7 +54,7 @@ public class DynamoDBClient extends DB {
     HASH_AND_RANGE
   }
 
-  private AmazonDynamoDBClient dynamoDB;
+  private AmazonDynamoDB dynamoDB;
   private String primaryKeyName;
   private PrimaryKeyType primaryKeyType = PrimaryKeyType.HASH;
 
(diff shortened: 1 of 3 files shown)


dynamodb/src/main/java/site/ycsb/db/DynamoDBClient.java
```diff
@@ -35,7 +37,7 @@
 import java.util.Vector;
 
 /**
- * DynamoDB v1.10.48 client for YCSB.
+ * DynamoDB v1.10.48 client for YCSB. Updated to use v 1.11.812.
```
busbey, line 40: just say it's using the v 1.11.812 or that folks should look at the pom file for the specific client version

busbey (commented):
overall looks good. I'll need to set aside time to use it against dynamodb unless another maintainer has time first.

ahachete:
OK. I have sent a rebased patch removing the version of the Dynamo client in the class Javadoc. I agree it's much better to see that in the pom.xml, I was just originally following what was there.

Please @busbey  let me know if I can help otherwise, hope it can be merged soon.

aka-chris:
Hello! Can the YCSB test the DynamoDB Local？I try it, but there is an error  `ERROR site.ycsb.db.DynamoDBClient  -com.amazonaws.services.dynamodbv2.model.ResourceNotFoundException: Cannot do operations on a non-existent table (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ResourceNotFoundException; `

ahachete:
> Hello! Can the YCSB test the DynamoDB Local？I try it, but there is an error `ERROR site.ycsb.db.DynamoDBClient -com.amazonaws.services.dynamodbv2.model.ResourceNotFoundException: Cannot do operations on a non-existent table (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ResourceNotFoundException; `
> OK, Let me check this branch `ahachete:dynamodb_improve_endpoint_region_config`

Yes you can, you just need to set appropriately the `dynamodb.endpoint = http://localhost:$port`, using this branch.

aka-chris:
I've checked it. The same error.

aka-chris:
```
dynamodb.endpoint = http://localhost:8000
dynamodb.awsCredentialsFile = /usr/data1/ycsb-dynamodb-binding-0.18.0-SNAPSHOT/AWSCredentials.properties
dynamodb.primaryKey = id
```
This what I set in the  `dynamodb.properties`

ahachete:
You error is not very well documented. You should attached the full config files used, command used, how the local version is running, etc.

I haven't tested myself with dynamodb local, but with a proxy service running locally, proxying to DynamoDB, and works well.

aka-chris:
OK.I'll show you later.Do I need to create `usertable` manually? And I don't know what I should assign `dynamodb.primaryKey`.

ahachete:
Yes, you need to create `usertable` manually. PK can be simple or composite, up to you. For the simplest case, a command similar to the following may help you:

```sh
aws --region $region --endpoint http://localhost:8080 dynamodb create-table --attribute-definitions '[{"AttributeName": "pk", "AttributeType": "S"}]' --key-schema '[{"AttributeName": "pk", "KeyType": "HASH"}]' --billing-mode PAY_PER_REQUEST --table-name usertable
```

aka-chris:
I did it! Thank you so so much

ahachete:
Hi @busbey Any changes we can get the PR reviewed? I'd love to see it merged, if that would be the case ;)

busbey (approved):
looks good! I'll ask for some help checking things again for a release.

PR #1447, opened 2020-07-28
[rocksdb] change mvn build target

xinglin:
change mvn build target to just build rocksdb-binding. This also makes the build command to be consistent with these for other bindings.


xinglin:
Done. thanks,

PR #1462, opened 2020-09-29
[hbase] HBase server side value filtering for long SCAN operations

bea0113:
The Apache HBase users usually implement primary use cases that fit well with HBase's intended short-scan and random get/put focus which already can be tested with YCSB.

However we can see in the industry that more and more users are also having secondary use cases that are better suited for traditional data warehouses, like longer scan operations with filtering and also aggregation. We wish to contribute to YCSB to enable the performance measurement of these secondary workloads.

First we would like to submit a patch about enabling optional server side value filtering for HBase. In the Core Workload, we have values randomly generated as 100 long byte arrays. We would introduce some HBase2 driver specific parameters that would setup value filtering on the scan operation, comparing a randomly generated 100 long byte array with the values in the scan results. The operator used for comparison should be configurable as well (like "lessOrEqual", "greater", etc).

We are not very sure if this should be a generic feature of the Core Workload or not. We would implement this first as an HBase2 specific option (as the operators supported by HBase might be limited / too HBase specific), but we are open for suggestions.


busbey:
Given the current implementation when this is active we'll pass a filter for each field that works at a different random set of bytes.  I'm trying to see the advantage of using random values for the filter by default compared to e.g. having another property that sets the value. Also why all fields rather than a configurable set of them?

Can you walk through an example comparison using YCSB with these changes in place? maybe that would make it easier for me to work through my questions.

symat:
Thank you for the comments!

> Given the current implementation when this is active we'll pass a filter for each field that works at a different random set of bytes. I'm trying to see the advantage of using random values for the filter by default compared to e.g. having another property that sets the value. 

I couldn't really come up with any benefit of having a random filter value. Having another property for setting a filtering value is great idea I think. It would allow us to have more fine control over the ratio of the returned cells. 
So maybe defining a configuration option like `hbase.scanfiltervalue`, a hexa string representing the byte array we use for filter value (we could default it to "7FFFFFFF...", so that using the default "lessOrEqual" operator, approx 50% of the cells would be filtered in)

> Also why all fields rather than a configurable set of them?

The Core workload already let us to use the `readallfields` property to decide if we want to scan only a single column, or all the 10 columns. I think this would be enough for the value filters too. However, in the PR we actually don't handle the `readallfields=true` case, when the `scan()` methods is called by the CoreWorkload with the `fields=null` parameter. I think we will need to add that, if it is easy.

> Can you walk through an example comparison using YCSB with these changes in place? 

So assuming a worlkoad with long scan operations for analytic queries, you can either 1) fetch the whole column (or a large part of it) and do the filtering on the client side, or 2) pushing down some filtering criteria to HBase itself. Option 1) would stress both the file system + block cache in HBase, and also the networking + RPC layers. While Option 2) would stress the filesystem/block-cache more but the networking + RPC (and the client) less. 

E.g. if we use the core workload (10 column, 100 bytes each) and set `readallfields=false` and enable value filtering (with filter value "7FFFFFFF...") and setting the average scan length to 4000 (`minscanlength=1000, maxscanlength=7000, scanlengthdistribution=uniform)`, then for each scan:
- HBase will need to read 4000 (rows) * 10 (columns) * 100 (bytes) ~= 4 MB (plus row and column keys)
- HBase will send to the client only 4000 (rows) * 1 (column) * 100 (bytes) * 0.5 (filtered) ~= 200 KB (plus row and column keys)

So by playing with the `readallfields` and the proposed `hbase.scanfiltervalue` parameters, we can finetune how much stress we want to put to the filesystem / cache compared to the network / RPC layers.

busbey:
Okay, I think adding a parameter for a value to filter on is a good starting point for this.

Later on we could make something more complicated to handle filtering with different values for different fields but we should try to have that expressed at the workload level before we do the extra work for that.

symat:
after re-checking the code, I see my numbers above were incorrect. When we are setting the average scan length to 4000 (minscanlength=1000, maxscanlength=7000, scanlengthdistribution=uniform), then HBase will send back ~4000 rows always, regardless of filtering. But if the value filtering is enabled, then we can expect that HBase will need to read more, in order to find the ~4000 rows. Correcting the previous example:

- HBase will need to read 4000 (expected rows) * 2 (as half of the rows will be filtered) * 10 (columns) * 100 (bytes)  ~= 8 MB (plus row and column keys)
- HBase will send to the client only 4000 (expected rows) * 1 (column) * 100 (bytes)  ~= 400 KB (plus row and column keys)

busbey:
Sure, but it's the same principal where the workload assumption is that those extra server side read records would need to be sent to the client and filtered if there was no push down.

symat:
yes, exactly

bea0113:
Hi Sean, thank you for the comments, we tried to address the issues you mentioned in the last commit. Please have a look at them and get back to us, if there's anything else we should change!

busbey (changes requested):

symat (approved):
Looks good to me! Thank you both for the comments and the fixes!

In a recent study, Bea was measuring HBase performance comparing Ozone and HDFS persistence layers using this patch. We had some interesting results. Hopefully we can publish the results soon in a blog post. :)

hbase2/src/main/java/site/ycsb/db/hbase2/HBaseClient2.java
```diff
@@ -165,9 +181,18 @@ public void init() throws DBException {
       debug = true;
     }
 
-    if ("false"
-        .equals(getProperties().getProperty("hbase.usepagefilter", "true"))) {
-      usePageFilter = false;
+    if (isBooleanParamSet("hbase.usepagefilter", true)) {
+      usePageFilter = true;
+    }
```
busbey, line 186: Currently `usePageFilter` defaults to true, so with this implementation there's no way to opt out. replace the if block, e.g.

```
usePageFilter = isBooleanParamSet("hbase.usepagefilter", usePageFilter);
```

hbase2/src/main/java/site/ycsb/db/hbase2/HBaseClient2.java
```diff
@@ -523,6 +557,15 @@ public Status delete(String table, String key) {
...
     this.config = newConfig;
   }
+
+  private boolean isBooleanParamSet(String param, boolean defaultValue){
+    return Boolean.parseBoolean(getProperties().getProperty(param, Boolean.toString(defaultValue)));
+  }
+
+  private CompareOperator getCompareOperator(String operator) {
+    return CompareOperator.valueOf(operator.toUpperCase());
+  }
```
busbey, line 567: this is only called in one place. replace the call with this implementation directly.

hbase2/src/test/java/site/ycsb/db/hbase2/HBaseClient2Test.java
```diff
@@ -15,6 +15,8 @@
 
 package site.ycsb.db.hbase2;
 
+import static org.apache.hadoop.util.StringUtils.hexStringToByte;
```
busbey, line 18: use `org.apache.hadoop.hbase.util.Bytes.fromHex` instead since it is part of HBase's public API.

hbase2/src/test/java/site/ycsb/db/hbase2/HBaseClient2Test.java
```diff
@@ -209,5 +256,68 @@ public void testUpdate() throws Exception{
...
+      final HashMap<String, ByteIterator> row = result.get(i);
+      assertEquals(2, row.size());
+      assertTrue(row.containsKey("col_1") && row.containsKey("col_2"));
+      assertArrayEquals(expectedValuesReturned[i], row.get("col_1").toArray());
+      assertArrayEquals(expectedValuesReturned[i], row.get("col_2").toArray());
+    }
+
+    // now scan only a single column (the filter should work here too)
+    result = new Vector<>();
+    client.scan(tableName, "00000", scanRowLimit, ImmutableSet.of("col_1"), result);
```
busbey, line 292: Instead of this `ImmutableSet` import use the jdk provided `Collections.singleton`

busbey (changes requested):
couple of small things left. Sorry I missed the stuff in the test code on the first pass.

bea0113:
I fixed the code based on your comments, thanks for taking the time and reviewing my PR!

busbey (approved):
Great! Thanks for going through all these revisions.

PR #1477, opened 2020-10-29
Add transaction workload

martijndeheus:
Hello! 

Thanks for the work you have done on this. 

I was looking for a benchmark to test a system which does have transaction features, but simpler than TPCC. I stumbled upon YCSB+T (#169). I wanted something like this, unfortunately this was never merged.

I implemented something similar myself. It is a very simple closed economy model. It adds an integer field to the documents (`balance`) and adds a `transfer` operation to the DB. This `transfer` operation represents a transaction where a certain amount of balance is subtracted from one document and added to another.

For this to work I also added an `insert` operation where the workload can set an initial balance for the documents. I also added a `deleteAndTransfer` operation. This deletes a document and transfers all of its balance to another document.

The implementation of these "transactional" operations (`transfer` and `deleteAndTransfer`) is left to the DB binding.

I created a workload which includes these new operations. 

I also added a REST client binding which implements the new operations.

I think I will add a validation step (as described in #169) as well. 

I am interested to hear what you think of this addition and whether you think it could be a good addition to YCSB.

### DIFF
--- a/bin/bindings.properties
+++ b/bin/bindings.properties
@@ -75,3 +75,4 @@ solr7:site.ycsb.db.solr7.SolrClient
 tarantool:site.ycsb.db.TarantoolClient
 tablestore:site.ycsb.db.tablestore.TableStoreClient
 voltdb:site.ycsb.db.voltdb.VoltClient4
+transactionrest:site.ycsb.transactionwebservice.transactionrest.TransactionRestClient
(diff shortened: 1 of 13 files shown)


morcelicaio:
Hello @martijndeheus , is this functionality already working in the latest version of YCSB?
I'm also looking for a benchmark to test a system with transaction capabilities.

PR #1481, opened 2020-11-25
[core] don't print dbwrapper err log msg many times

sitano:


busbey (commented):
Oh thank you.

Please leave out unrelated formatting improvements and update the commit message to start with "[core]".

Issue #1391, opened 2020-01-31
[hypertable] time to remove hypertable?


busbey:
in 2016 the company backing hypertable announced they were shuttering and stopping active dev:
https://hypertable.com/blog/hypertable_inc._is_closing_its_doors

user mailing list has atleast 1 active person from last year though:
https://groups.google.com/forum/#!topic/hypertable-user/p4V4HxWHxdI

PR #1483, opened 2020-11-30
[hypertable] remove the hypertable binding.

busbey:
this will close out #1391 . there's been no real additional activity I can find in the project.


symat (approved):
Looks good to me.
I also checked the hypertable website and the mailing list, I think it is safe to say that there is no activity for a long time and people are not using Hypertable in production.

busbey:
Thanks for the review!

PR #1484, opened 2020-11-30
[solr, solr6] remove EOL solr versions.

busbey:
According to the solr community solr 5 and solr 6 are both EOM and shouldn't be used.

https://lucene.apache.org/solr/downloads.html#about-versions-and-support


risdenk (approved):

busbey:
Thanks for the review!

PR #1487, opened 2020-12-04
[cassandra] replace consistency ONE with QUORUM

sitano:
replace consistency ONE with QUORUM

### DIFF
--- a/cassandra/README.md
+++ b/cassandra/README.md
@@ -68,9 +68,8 @@ For keyspace `ycsb`, table `usertable`:
 * `cassandra.readconsistencylevel`
 * `cassandra.writeconsistencylevel`
 
-  * Default value is `ONE`
+  * Default value is `QUORUM`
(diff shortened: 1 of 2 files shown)


cassandra/README.md
```diff
@@ -68,7 +68,7 @@ For keyspace `ycsb`, table `usertable`:
 * `cassandra.readconsistencylevel`
 * `cassandra.writeconsistencylevel`
 
-  * Default value is `ONE`
+  * Default value is `QUORUM`
```
busbey, line 71: The note about durability in the face of node failure should get updated?

busbey (changes requested):

cassandra/README.md
sitano: removed line. Sorry, I did not see it before.

busbey (approved):

Issue #1491, opened 2020-12-07
Scylla Support: enable token aware driver by default


tzach:
https://github.com/brianfrankcooper/YCSB/commit/ff4ecbd37f1316c3bce773a8989920fbb5d5d82a adds Scylla support to YCSB.
For best result with Scylla, its better to enable token aware driver by default.

tzach:
@sitano

PR #1507, opened 2021-01-26
[scylla] enable token aware LB by default, improve the docs

sitano:
[scylla] enable token aware LB by default
[scylla] doc: add latency correction section
[scylla] doc: dump and merge histograms
[scylla] doc: don't configure connections manually

closes https://github.com/brianfrankcooper/YCSB/issues/1491

### DIFF
--- a/scylla/README.md
+++ b/scylla/README.md
@@ -53,7 +53,6 @@ Create a keyspace, and a table as mentioned above. Load data with:
         -p readproportion=0 -p updateproportion=0 \
         -p fieldcount=10 -p fieldlength=128 \
         -p insertstart=0 -p insertcount=1000000000 \
-        -p cassandra.coreconnections=14 -p cassandra.maxconnections=14 \
         -p cassandra.username=cassandra -p cassandra.password=cassandra \
         -p scylla.hosts=ip1,ip2,ip3,...
 
@@ -63,7 +62,6 @@ Use as following:
         -target 120000 -threads 840 -p recordcount=1000000000 \
         -p fieldcount=10 -p fieldlength=128 \
         -p operationcount=50000000 \
-        -p scylla.coreconnections=280 -p scylla.maxconnections=280 \
         -p scylla.username=cassandra -p scylla.password=cassandra \
         -p scylla.hosts=ip1,ip2,ip3,...
 
@@ -171,6 +169,16 @@ want to keep shards incoming queues non-empty. For that your parallelism factor
 must be at least 2.
 
 ### 5. Number of connections
+
+If you use original Cassandra drivers you need to pick the proper number
+of connections per host. Scylla drivers do not require this to be configured
+and by default create a connection per shard. For example if your node has
+16 vCPU and thus 14 shards Scylla drivers will pick to create 14 connections
+per host. An excess of connections may result in degraded latency.
+
+Database client protocol is asynchronous and allows queueing requests in
+a single connection. The default queue limit for local keys is 1024 and 256
+for remote ones. Current binding implementation do not require this.
 
 Both `scylla.coreconnections` and `scylla.maxconnections` define limits
 per node. When you see `-p scylla.coreconnections=280 -p scylla.maxconnections=280`
--- a/scylla/README.md
+++ b/scylla/README.md
@@ -71,18 +71,22 @@ Use as following:
 
 ### 1. Load target
 
-You want to test how a database handles load. To get the performance picture
-you would want to look at the latency distribution and utilization under the
-constant load. To select load use `-target` to state desired throughput level.
+Suppose, you want to test how a database handles an open-class system load.
+
+In this case, to get the performance picture you want to look at the latency
+distribution and utilization under the constant throughput. Use the `-target` flag
+to state desired throughput level.
 
 For example `-target 120000` means that we expect YCSB workers to generate
 120,000 requests per second (RPS, QPS or TPS) to the database.
 
 Why is this important? Because without setting target throughput you will be
-looking only on the system equilibrium point that in the face of constantly
-varying latency will not allow you to see either throughput, nor latency.
+looking at the system equilibrium point which in the face of constantly
+varying latency would show you neither throughput, nor latency. Your system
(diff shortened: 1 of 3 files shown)


sitano:
@tarzanek take a look at this fix

scylla/README.md
```diff
@@ -245,4 +290,4 @@ of 3 nodes of i3.4xlarge (16 vCPU per node) and target of 120000 is:
 
 - `scylla.lwt`
   - Use LWT for operations
-  - Default is false.
+  - Default is true.
```
tarzanek, line 293: revert
default is false:
https://github.com/brianfrankcooper/YCSB/blob/master/scylla/src/main/java/site/ycsb/db/scylla/ScyllaCQLClient.java#L65

tarzanek:
scylla.tokenaware
- default driver setting is DCAware(TokenAware) - you don't need to specify local DC, if there is just one DC
so this is true if there is just single DC

scylla/README.md
tarzanek: (and I'd keep it false to be in sync with cassandra)

scylla/src/main/java/site/ycsb/db/scylla/ScyllaCQLClient.java
```diff
@@ -87,7 +87,7 @@
   public static final String SCYLLA_LWT = "scylla.lwt";
 
   public static final String TOKEN_AWARE = "scylla.tokenaware";
-  public static final String TOKEN_AWARE_DEFAULT = "false";
+  public static final String TOKEN_AWARE_DEFAULT = "true";
```
tarzanek, line 90: this shouldn't have any impact, if there is just single DC
look at
https://docs.datastax.com/en/developer/java-driver/3.10/manual/load_balancing/
```
If you don’t explicitly configure the policy, you get the default, which is a datacenter-aware, token-aware policy:

new TokenAwarePolicy(DCAwareRoundRobinPolicy.builder().build());
```

tarzanek: so
TOKEN_AWARE_LOCAL_DC 
is the only param that is needed, in case there are multi DCs

sitano: oh. I didn't know that it works by default.

sitano: @tarzanek I have removed this flag then "scylla.tokenaware" as it is does not make sense having it.

scylla/README.md
sitano: Thx! did this by mistake.

sitano:
@busbey @tarzanek ping

tarzanek:
it makes sense to keep TOKEN_AWARE_LOCAL_DC (or use some other name, your choice)
since for multi DC testing cluster specifying this local DC sets proper distance and makes queries enforce to go for local DC
so under the if you'd have something ala
```
if TOKEN_AWARE_LOCAL_DC is set { 
String localDc = getProperties().getProperty(TOKEN_AWARE_LOCAL_DC);
policy = new TokenAwarePolicy(DCAwareRoundRobinPolicy.builder().withLocalDc(localDc).build());
builder = builder.withLoadBalancingPolicy(policy);
}
```

- if you specify local DC queries will go there, then based on retry policy they might try other DCs, but you can later enforce and specify the behaviour using CL
LOCAL_ONE or LOCAL_QUORUM only go to local DC

so you should also change the default to LOCAL_QUORUM and claim the default supported is local DC
and for multi DC one would need to set TOKEN_AWARE_LOCAL_DC and proper consistency level based on what you want to test (QUORUM or LOCAL_QUORUM)

sitano:
@tarzanek ok, I returned configuration for `scylla.local_dc` back. Do you know if use of TokenAwarePolicy here is necessary?

```
policy = new TokenAwarePolicy(DCAwareRoundRobinPolicy.builder().withLocalDc(localDc).build());
```

Because as far as I saw Shard-Awareness is implemented in a different place / level.

tarzanek:
> 
> 
> @tarzanek ok, I returned configuration for `scylla.local_dc` back. Do you know if use of TokenAwarePolicy here is necessary?
> 
> ```
> policy = new TokenAwarePolicy(DCAwareRoundRobinPolicy.builder().withLocalDc(localDc).build());
> ```
> 
> Because as far as I saw Shard-Awareness is implemented in a different place / level.

I'd keep it, since above is default, we can ask @haaawk to be 100% sure
I'll review now

tarzanek:
lgtm actually, so from my side feel free to merge!
and thank you for addressing my comments

haaawk:
> > @tarzanek ok, I returned configuration for `scylla.local_dc` back. Do you know if use of TokenAwarePolicy here is necessary?
> > ```
> > policy = new TokenAwarePolicy(DCAwareRoundRobinPolicy.builder().withLocalDc(localDc).build());
> > ```
> > 
> > 
> > Because as far as I saw Shard-Awareness is implemented in a different place / level.
> 
> I'd keep it, since above is default, we can ask @haaawk to be 100% sure
> I'll review now

Shard awareness is implemented in different place - that's correct. It won't work without TokenAwarePolicy though.

sitano:
thank you guys! @busbey would you be interested in taking a look at this one?

scylla/README.md
```diff
@@ -92,7 +96,51 @@ See
...
+or 
+
+    -p hdrhistogram.fileoutput=true
+    -p hdrhistogram.output.path=file.hdr
+
+merge them manually and extract required percentiles out of the
+joined result.
+
+Remember that running multiple workloads may distort original
+workloads distributions they were intended to produce.
```
busbey, line 141: oh lord, thank you. I can't tell you how many times this comes up. I'd like to get some version of this into the main README.

busbey (approved):

scylla/README.md
sitano: https://github.com/brianfrankcooper/YCSB/pull/1512/

PR #1508, opened 2021-01-27
core: handle sigint (ctrl+c)

sitano:
This patch makes YCSB to print measurement
results when the process receives a SIGINT
(CTRL+C) signal to output results before exit.

It is useful for all who don't want to wait
for the process to finish requested load amount
and don't want to lose measured results.


sitano:
@busbey ^^

sitano:
@busbey ping

core/src/main/java/site/ycsb/Client.java
```diff
@@ -325,6 +324,7 @@ public static void main(String[] args) {
     }
 
     Thread terminator = null;
+    final CountDownLatch l = new CountDownLatch(1);
```
allanbank, line 327: The use of the CountDownLatch I think obscures what is it doing. I'd suggest moving to a Semaphore. 

Here would be:

```java
final Semaphore done = new Semaphore(1);
done.acquire();
```  

Below it would be:
```java
done.release();
```

Then in the hook:
```java
if( ! done.tryAquire() ) {
  // ...stopWorkload()
  try {
    if( ! done.tryAquire(5, TimeUnit.SECONDS) ) {
     System.err.println("Shutdown await timeout. Exit.");
    }
  } catch (InterruptedException timeout) {
  }
}
```

Extra credit to move this to a Phaser where things can then track the progress of the run (init, workload, cleanup, export measurement). In this case, as an example, you probably don't want to trigger the hook if we are already in the export phase.

core/src/main/java/site/ycsb/TerminatorThread.java
```diff
@@ -28,26 +28,29 @@
...
       Thread.sleep(maxExecutionTime * 1000);
     } catch (InterruptedException e) {
       System.err.println("Could not wait until max specified time, TerminatorThread interrupted.");
       return;
     }
+    stopWorkload();
+  }
+
+  public void stopWorkload() {
     System.err.println("Maximum time elapsed. Requesting stop for the workload.");
```
allanbank, line 54: This message should be in the `run()` method.

core/src/main/java/site/ycsb/Client.java
```diff
@@ -336,6 +336,20 @@ public static void main(String[] args) {
         threads.put(new Thread(tracer.wrap(client, "ClientThread")), client);
       }
 
+      // Handle SIGINT signal (CTRL+C) to export measurements before shutdown
+      Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+          if (l.getCount() == 1) {
+            new TerminatorThread(maxExecutionTime, threads.keySet(), workload).stopWorkload();
```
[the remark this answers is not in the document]
sitano: I was lazy last time to make `stopWorkload` static, so I had to create an instance. This time a made it static.

core/src/main/java/site/ycsb/Client.java
```diff
@@ -394,9 +409,11 @@ public static void main(String[] args) {
     } catch (IOException e) {
       System.err.println("Could not export measurements, error: " + e.getMessage());
       e.printStackTrace();
+      l.countDown();
       System.exit(-1);
     }
 
+    l.countDown();
```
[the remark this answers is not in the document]
sitano: Done

core/src/main/java/site/ycsb/Client.java
sitano: Done. I am sorry I didnt have time this time to port it to Phaser.

core/src/main/java/site/ycsb/TerminatorThread.java
sitano: Fixed

sitano:
@allanbank ping

PR #1512, opened 2021-02-23
doc: percentiles and hdr merge

sitano:
@busbey you asked for it (https://github.com/brianfrankcooper/YCSB/pull/1507#pullrequestreview-591345620), you get it ;)

This PR updates a README.md file with the information about how to work with latency percentiles and that a user shall not look up for averages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,57 @@ To build the full distribution, with all database bindings:
 To build a single database binding:
 
     mvn -pl site.ycsb:mongodb-binding -am clean package
+
+Running multiple instances and latency percentiles
+--------------------------------------------------
+
+In general, you shall be interested in 99% percentile (P99) of the latency
+distribution, and the rest of the tail - 99.9%, 99.99%, 99.999%. The difference
+between the amount of requests that will be observed by a user that fall
+into 95% (P95) percentile and 99% percentile may be sufficiently large.
+
+For example, see "How Many Nines?" at https://bravenewgeek.com/everything-you-know-about-latency-is-wrong/.
+The formula to calculate probability of how many clients will observe
+a specific percentile is: 
+
+    Probability_to_observe = 1 - Percentile ^ Requests
+
+That is why almost 30% of the users will observe latency worse than P99
+just by loading the default _google.com_ web page:
+
+    1 - 0.99 ^ 30 = 0.27
+
+Remember, that
+
+- _latencies_ percentiles can't be averaged. Don't fall into this
+  [trap](http://latencytipoftheday.blogspot.com/2014/06/latencytipoftheday-you-cant-average.html).
+  Neither latency averages, nor P99 averages do not make any sense.
+
+If you run multiple loaders dump result histograms with:
+
+    -p hdrhistogram.fileoutput=true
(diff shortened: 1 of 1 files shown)


sitano:
@busbey ping

README.md
```diff
@@ -78,3 +78,79 @@ To build the full distribution, with all database bindings:
 To build a single database binding:
 
     mvn -pl site.ycsb:mongodb-binding -am clean package
+
+Running multiple instances and latency percentiles
+--------------------------------------------------
+
+In general, you shall be interested in 99% percentile (P99) of the latency
+distribution and the rest - 99.9%, 99.99%, 99.999% - the latency tail.
+95% percentile has a very low probability to be observed by the user
```
allanbank, line 87: >  95% percentile has a very low probability to be observed by the user...

This was a little confusing to me at first.

Maybe reword to something like:

> A user is very unlikely to have _all_ of their requests happen within the 95% percentile, so don't even bother.

README.md
```diff
@@ -78,3 +78,79 @@ To build the full distribution, with all database bindings:
...
+so don't even bother (see "How Many Nines?" at
+https://bravenewgeek.com/everything-you-know-about-latency-is-wrong/
+). The formula to calculate probability of how many clients will
+observe a specific percentile is: 
+
+    Probability_to_observe = 1 - Percentile ^ Requests
```
allanbank, line 93: Would this be clearer as:

> Probability_to_observe = 1 - Percentile<sup>Requests</sup>

README.md
```diff
@@ -78,3 +78,79 @@ To build the full distribution, with all database bindings:
...
+    -p hdrhistogram.fileoutput=true
+    -p hdrhistogram.output.path=file.hdr
+
+merge them manually and extract required percentiles out of the
+joined result.
+
+Remember that running multiple workloads may distort original
+workloads distributions they were intended to produce.
+
+Merging HDR histogram percentiles
```
allanbank, line 117: I like that you pointed the user to tool to help them but I think it would be better to link them to the documentation for that tool rather than providing a synopsis here. As the HDRLogProcessing tool changes the likelihood of this documentation being updated is fairly low which will lead to questions and confusion for users.

sitano: I removed all examples and left the link only to the tool github which has a README. I left an example for CSV extraction. If you want ping me to remove it.

README.md
sitano: I don't know how to do that and I did find a way to do that.

README.md
sitano: I changed the wording to

> In general, you shall be interested in 99% percentile (P99) of the latency
distribution, and the rest of the tail - 99.9%, 99.99%, 99.999%. The difference
between the amount of requests that will be observed by a user that fall
into 95% (P95) percentile and 99% percentile may be sufficiently large.

> For example, see "How Many Nines?" at https://bravenewgeek.com/everything-you-know-about-latency-is-wrong/.
The formula to calculate probability of how many clients will observe
a specific percentile is:

sitano:
@allanbank ping

busbey (approved):
thanks for the contribution, someone can redo the formulas with latex formatting later

PR #1515, opened 2021-03-08
scylla: make formula clear to pick num of threads

sitano:


sitano:
@allanbank I am curious if the example that I have added here makes sense from your PoV.

sitano:
ping

PR #1535, opened 2021-04-28
Redis: Using JedisPool instead of Jedis

yangbodong22011:
Hi, I am one of the [Jedis](https://github.com/redis/jedis) Reviewers. When I used YCSB to write large amounts of data (over 100 million) to Redis, I encountered a failure when the connection is abnormal (even if the insertionRetryLimit is configured). After reading the YCSB code, I found that YCSB uses `Jedis` instead of `JedisPool`. When Jedis encounters a connection timeout or disconnection, it cannot be restored. In order to run the `long-term stable` test, I introduced JedisPool. This PR mainly includes:

- Replace Jedis with JedisPool
- Exception is processed and `Status.ERROR` is returned
- Upgrade the Jedis version to 3.6.0 (the latest stable version)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,7 +138,7 @@ LICENSE file.
     <openjpa.jdbc.version>2.1.1</openjpa.jdbc.version>
     <orientdb.version>2.2.37</orientdb.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <redis.version>2.9.0</redis.version>
+    <redis.version>3.6.0</redis.version>
     <riak.version>2.0.5</riak.version>
     <rocksdb.version>6.2.2</rocksdb.version>
     <s3.version>1.10.20</s3.version>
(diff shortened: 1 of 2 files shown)


yangbodong22011:
@busbey please have a look when you have time, thanks.

pom.xml
```diff
@@ -138,7 +138,7 @@ LICENSE file.
     <openjpa.jdbc.version>2.1.1</openjpa.jdbc.version>
     <orientdb.version>2.2.37</orientdb.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <redis.version>2.9.0</redis.version>
+    <redis.version>3.6.0</redis.version>
```
sazzad16, line 141:
```suggestion
    <redis.version>3.7.0</redis.version>
```

sazzad16 (approved):

Issue #370, opened 2015-07-24
Repository used in the mongodb pom.xml not directly accessible


gkamat:
The pom.xml file for the MongoDB binding contains a reference to the following repository:

  http://www.allanbank.com/repo

However trying to access it directly does not work, for example:

$ wget http://www.allanbank.com/repo/
--2015-07-24 14:50:24--  http://www.allanbank.com/repo/
Resolving www.allanbank.com (www.allanbank.com)... 206.210.70.161
Connecting to www.allanbank.com (www.allanbank.com)|206.210.70.161|:80... connected.
HTTP request sent, awaiting response... 403 Forbidden
2015-07-24 14:50:24 ERROR 403: Forbidden.

Also, it would also be good to use the https transport here, if possible.

allanbank:
Yes, file listing is turned off on the server since it is not needed.  If you try to download the artifacts directly (see links below) it works. More details here:http://www.allanbank.com/mongodb-async-driver/userguide/maven.html

https is on the todo list.

http://www.allanbank.com/repo/com/allanbank/mongodb-async-driver/2.0.1/mongodb-async-driver-2.0.1.jar
http://www.allanbank.com/repo/com/allanbank/mongodb-async-driver/2.0.1/mongodb-async-driver-2.0.1-sources.jar
http://www.allanbank.com/repo/com/allanbank/mongodb-async-driver/2.0.1/mongodb-async-driver-2.0.1.pom

mthak:
The problem we have is "if this artifact changes to something else tomorrow, or if we add more artifacts like this in future" . Without a mirror It would need someone to download the artifacts manually and upload it on our internal maven repository.

busbey:
@mthak isn't that why folks set up nexus (or whatever repository) to mirror things? How is this particular repo different from the other ones we use in e.g. for tarantool or gemfire?

gkamat:
It would be helpful to have maven mirror/snapshot all dependent repositories at each YCSB release, and use those for the build.  Perhaps there is a way to enhance the build specification to do that.

busbey:
I was checking and I believe the maven-dependency-plugin should make this easy enough to do on releases.

I'll make a go at putting one in place for the 0.2.0 Release and we can see what would make it sustainable.

risdenk:
[...]

PR #1540, opened 2021-05-27
Add jaxb XML bindings to pom.xml.

sears:
These are no longer included in Java by default.  They are needed when building with OpenJDK 17.

I haven't tested with older versions of Java (hopefully CI will take care of that for me).


sears:
Hmm:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.5.5:single (default) on project memcached-binding: Failed to create assembly: Error creating assembly archive jar-with-dependencies: Problem creating TAR: entry 'javax/xml/bind/ModuleUtil.class' closed at '0' before the '1107' bytes specified in the header were written -> [Help 1]
```
This looks like: https://issues.apache.org/jira/browse/MASSEMBLY-34 I tried upgrading maven-assembly-plugin, and it got further on my machine.  However, it then hit https://github.com/brianfrankcooper/YCSB/issues/370

I pushed the updated assembly plugin, and CI is running again now.

PR #1549, opened 2021-07-15
Test runner

Maithem:
Add Test Runner and Reporting


core/src/main/java/site/ycsb/measurements/exporter/JSONMeasurementsExporter.java
```diff
@@ -38,6 +38,7 @@ public JSONMeasurementsExporter(OutputStream os) throws IOException {
     BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(os));
     g = factory.createJsonGenerator(bw);
     g.setPrettyPrinter(new DefaultPrettyPrinter());
+    g.writeStartArray();
```
Maithem, line 41: port this fix upstream

PR #1550, opened 2021-07-15
Fix JSON Exporter

Maithem:
The exporter was producing an invalid json string, this patch
adds the array structure.


Maithem:
@busbey  can you please review this ?

PR #1560, opened 2021-09-07
[bindings] add ipv6 support for memcached binding.

peterzheng98:
IPv6 address should be written in the form of "[address]:port".
The port may be omitted but the brackets outside the address should be
kept.
Device name such as "ens0" should be written in address like
"[fe80::2%ens0]"

### DIFF
--- a/memcached/src/main/java/site/ycsb/db/MemcachedClient.java
+++ b/memcached/src/main/java/site/ycsb/db/MemcachedClient.java
@@ -161,15 +161,52 @@ public class MemcachedClient extends DB {
     // when dealing with IPv6 addresses.
     //
     // TODO(mbrukman): fix this.
+    // Since the following address cannot be identified:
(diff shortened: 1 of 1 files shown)


Asim-21:
Hi I am not able to run this with ipv6. with ipv4 it works fine. please the commands I am trying:

`./bin/ycsb load memcached -s -P workloads/workloada -p "memcached.hosts=[fe80::222:48ff:fe25:2f10%eth0]:11211"`

adding `memcached.hosts = "[fe80::222:48ff:fe25:2f10%eth0]:11211"` in the property file as well and running the following:
` ./bin/ycsb load memcached -s -P workloads/workloada -P memcached/conf/memcached.properties`

for both, it gives me the following error:
`site.ycsb.DBException: java.lang.NumberFormatException: For input string: ":222:48ff:fe25:2f10]:11211""`

Can you please help me out on this?

PR #1513, opened 2021-03-05
long parse fix

Chester-Ong:
small bug fix that fixes record generation bug for **insert_count** > INT_MAX or **record_count** > INT_MAX

PR #1579, opened 2021-11-29
properly convert insertcount from string

tarzanek:


tarzanek:
support bigger insertcount than 2147483647

tarzanek:
duplicate of https://github.com/brianfrankcooper/YCSB/pull/1513 ?

tarzanek:
actually I guess https://github.com/brianfrankcooper/YCSB/pull/1513 is duplicate of this, I verified the patch works 

I understand there might be overflow problems once we reach 9,223,372,036,854,775,807 but we should worry about that once we get there ;-)

PR #1621, opened 2022-06-27
[core] Changed the datatype for keynum from integer to long to allow …

ragarkar:
Changed the datatype for keynum from integer to long to allow ycsb to load a larger dataset. Currently, ycsb restricts the number of rows that can be loaded to 2^32 rows.

### DIFF
--- a/core/src/main/java/site/ycsb/workloads/CoreWorkload.java
+++ b/core/src/main/java/site/ycsb/workloads/CoreWorkload.java
@@ -611,7 +611,7 @@ public class CoreWorkload extends Workload {
    */
   @Override
   public boolean doInsert(DB db, Object threadstate) {
-    int keynum = keysequence.nextValue().intValue();
+    long keynum = keysequence.nextValue().intValue();
     String dbkey = CoreWorkload.buildKeyName(keynum, zeropadding, orderedinserts);
     HashMap<String, ByteIterator> values = buildValues(dbkey);
 


core/src/main/java/site/ycsb/workloads/CoreWorkload.java
```diff
@@ -611,7 +611,7 @@ private String buildDeterministicValue(String key, String fieldkey) {
    */
   @Override
   public boolean doInsert(DB db, Object threadstate) {
-    int keynum = keysequence.nextValue().intValue();
+    long keynum = keysequence.nextValue().intValue();
```
joshelser, line 614:
```suggestion
    long keynum = keysequence.nextValue();
```

Drop the `intValue()`?  `nextValue()` is returning back a `Long` and it will be unboxed.

joshelser (changes requested):
It looks like the all of the rest of the `doTransaction*` operations have already been updated to use the generated `long`.

That `intValue()` in your patch isn't causing problems (since we know this patch works), but it's a little confusion to see :)

core/src/main/java/site/ycsb/workloads/CoreWorkload.java
ragarkar: @joshelser My earlier patch was missing the change to return the long value. Fixed the patch to get the long type from the Number returned by nextValue().

busbey:
there are already several open PRs with attempts to switch from a max size of MAX_INT to MAX_LONG. please review them to ensure you cover all the changes they do and expressly note that.

PR #1624, opened 2022-07-01
[DynamoDB] Implemented optional batchWriteItem support for inserts and upgraded to AWSK SDK for Java v2

arjanschaaf:
Adding support for using DynamoDB Batch write API and upgraded to AWS SDK for Java v2.
Using batch writing during the load phase of YCSB provides a significant performance benefit, especially when creating larger test data sets.


fruch:
@arjanschaaf looks good, especially the part of switching to newer aws client, and setting the credentials not only from a configuration file.

PR #1661, opened 2023-01-31
[hbase] add a property to allow for reverse scans

igorbernstein2:
This add the property `hbase.reversescans` for both the hbase1 & hbase2 drivers. This will allow endusers to compare performance between forward and reverse scans


igorbernstein2:
closing in favor of #1727 1727

PR #1677, opened 2023-05-10
feat: make JDBC connection URL delimiter configurable

### LINKED ISSUE #1458: Recent change of JDBC URL splitting delimiter from , to ; breaks connectivity to MSSQL
#1233 discusses a change of URL splitting delimiter from , to ; to allow connections to multiple shards. A comment from @busbey mentions that semicolons aren't present in JDBC URLs. That is not correct for MSSQL JDBC driver, which uses semicolons to delimit connection string properties. Doc reference: https://docs.microsoft.com/en-us/sql/connect/jdbc/building-the-connection-url. 

Thus, it is no longer possible to specify any connection properties in the URL, including a SQL Server database name to connect to. Only connections to the _master_ database are possible, but _master_ is a system database that should not be used for any user workload. This also makes it impossible to use YCSB against Azure SQL where specifying database name in the connection URL is mandatory.

Proposal: add support for quoting the db.url property. If quoted, any semicolon within the URL string will be treated as part of the string, not as a shard delimiter. Example of proposed syntax for two shards:

db.url="jdbc:sqlserver://server-name.database.windows.net:1433;databaseName=ycsb1";"jdbc:sqlserver://server-name.database.windows.net:1433;databaseName=ycsb2"

Currently, when URL is
db.url=jdbc:sqlserver://server-name.database.windows.net:1433;databaseName=ycsb

connection fails as follows:

YCSB Client 0.17.0

Loading workload...
Starting test.
Adding shard node URL: jdbc:sqlserver://server-name.database.windows.net:1433
Adding shard node URL: databaseName=ycsb
Error in database operation: java.sql.SQLException: No suitable driver found for databaseName=ycsb
site.ycsb.DBException: java.sql.SQLException: No suitable driver found for databaseName=ycsb

olavloite:
Make the JDBC connection URL delimiter configurable. The current delimiter is semicolon ';', which breaks JDBC drivers that use semicolons in the actual connection URL. The default delimiter remains semicolon for all databases, except a list of databases that are known to use semicolons in the connection URL. Users can always specifiy a custom delimiter to work around any issues with the default delimiter.

Fixes #1458

### DIFF
--- a/jdbc/README.md
+++ b/jdbc/README.md
@@ -91,6 +91,18 @@ db.user=admin
 db.passwd=admin
 ```
 
+You can specify a custom delimiter for multiple connection URLs by setting `db.urldelimiter` if your connection URL might contain a `;`.
+The default delimiter is `,` for Google Cloud Spanner, SQLServer, H2, and Apache Phoenix. It is `;` for all
+other databases.
+
+```sh
+db.driver=com.google.cloud.spanner.jdbc.JdbcDriver
+db.url=jdbc:cloudspanner:/projects/my-project/instances/my-instance/databass/my-database;minSessions=1000;maxSessions=1000;numChannels=20
+db.urldelimiter=,
+db.user=
+db.passwd=
+```
+
 You can add these to your workload configuration or a separate properties file and specify it with ```-P``` or you can add the properties individually to your ycsb command with ```-p```.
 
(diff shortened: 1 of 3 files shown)


olavloite:
@busbey @allanbank @kruthar  Friendly ping. Any chance that we could get a review for this?

PR #1773, opened 2025-11-03
[core] Fix issue where recordcount in the workload is limited to max …

kumaran-coder:
recordcount is limited to max integer value because the CoreWorkload.init routine is using Integer.parseInt.  Change the code to use Long.parseLong instead.


kumaran-coder:
Hi Community,

can someone review this fix. Thank you.

cooperb94040 (approved):

